PR #55, opened 2014-08-15
PATCH: boost serial speed 3x or more

Cougar:
pymodbus over serial port is slow. The reason is that once serial port is opened, the initial timeout values is an indication how long it takes to get Modbus response. If one knows exactly how long it takes for controller to respond to the request, it is possible to manually tune timeout with quite good precision. But when different requests are used where responses have different latency and different length, then timeout should be set based on the slowest response. Otherwise some data just remains in buffer and both this and next response will be incorrect.

To overcome this, we need to get rid of timer and know how big is the response of any request. When ModbusSerialClient reads the response, it should know how many bytes to read and not wait for any timeout. For that I implemented getResponseSize() method in ModbusRequest class (and its derived classes). If response size is known, then ModbusTransactionManager.execute() reads only this amount of bytes. If size is not known, then the old timeout based behavior is used. Latter happens also with exception responses which are usually shorter than full response frames. Because most operations do not depend on timeout any more, much longer timeout can be used.

This is implemented only for ModbusRtuFramer and serial port right now.

Using this technique alone could cause a new kind of problem: the next request might happen too quickly after previous one. Modbus standard requires that there is silent interval of at least 3.5 character times between message serial RTU frames. For that _silentinterval is calculated based on serial port baud rate in ModbusSerialClient.connect(), and serial port write times are always saved to the _lastsend variable. Before every send it is checked that there has been at least _silentinterval delay between consecutive requests. However, I'm not sure if serial port writes are always synchronous. This may need some more work if not.

I also added serial port input buffer flushing before sending request. This makes sure that any old or incorrect data can't mess up with the next response.

Some words about performance. I'm testing with Modbus RTU slave at 19200 bps. This controller responds to holding register reads usually within 0.4 … 14 ms. However sometimes the response takes around 20 ms. 7 byte response for one holding register request takes 4.5 ms. So, the minimum timeout that can be used should be at least 25 ms if only such requests are used. And results are:

pymodbus with timeout: 24…26 req/sec
pymodbus with getResponseSize(): 76…78 req/sec

Enjoy,
## 

Cougar

### DIFF
--- a/pymodbus/client/sync.py
+++ b/pymodbus/client/sync.py
@@ -304,6 +304,8 @@ class ModbusSerialClient(BaseModbusClient):
         self.parity   = kwargs.get('parity',   Defaults.Parity)
         self.baudrate = kwargs.get('baudrate', Defaults.Baudrate)
         self.timeout  = kwargs.get('timeout',  Defaults.Timeout)
+        self._last_frame_end = 0.0
+        self._silent_interval = 3.5 * (1 + 8 + 2) / self.baudrate
 
     @staticmethod
     def __implementation(method):
@@ -333,7 +335,6 @@ class ModbusSerialClient(BaseModbusClient):
             _logger.error(msg)
             self.close()
         self._last_frame_end = time.time()
-        self._silent_interval = 3.5 * (1 + 8 + 2) / self.baudrate
         return self.socket != None
 
     def close(self):
--- a/pymodbus/client/sync.py
+++ b/pymodbus/client/sync.py
@@ -343,12 +343,23 @@ class ModbusSerialClient(BaseModbusClient):
     def _send(self, request):
         ''' Sends data on the underlying socket
 
+        If receive buffer still holds some data then flush it.
+
         :param request: The encoded request to send
         :return: The number of bytes written
         '''
         if not self.socket:
             raise ConnectionException(self.__str__())
         if request:
+            try:
+                waitingbytes = self.socket.inWaiting()
+                if waitingbytes:
+                    result = self.socket.read(waitingbytes)
+                    if _logger.isEnabledFor(logging.WARNING):
+                        _logger.warning("cleanup recv buffer before send: " + " ".join([hex(ord(x)) for x in result]))
+            except NotImplementedError:
+                pass
+
             return self.socket.write(request)
         return 0
 
--- a/pymodbus/client/sync.py
+++ b/pymodbus/client/sync.py
@@ -1,5 +1,6 @@
 import socket
 import serial
+import time
 
 from pymodbus.constants import Defaults
 from pymodbus.factory import ClientDecoder
@@ -331,6 +332,8 @@ class ModbusSerialClient(BaseModbusClient):
         except serial.SerialException, msg:
             _logger.error(msg)
             self.close()
+        self._last_frame_end = time.time()
+        self._silent_interval = 3.5 * (1 + 8 + 2) / self.baudrate
         return self.socket != None
 
     def close(self):
@@ -345,12 +348,20 @@ class ModbusSerialClient(BaseModbusClient):
 
         If receive buffer still holds some data then flush it.
 
+        Sleep if last send finished less than 3.5 character
+        times ago.
+
         :param request: The encoded request to send
         :return: The number of bytes written
         '''
         if not self.socket:
             raise ConnectionException(self.__str__())
         if request:
+            ts = time.time()
+            if ts < self._last_frame_end + self._silent_interval:
+                _logger.debug("will sleep to wait for 3.5 char")
+                time.sleep(self._last_frame_end + self._silent_interval - ts)
+
             try:
                 waitingbytes = self.socket.inWaiting()
                 if waitingbytes:
@@ -360,7 +371,9 @@ class ModbusSerialClient(BaseModbusClient):
             except NotImplementedError:
                 pass
 
-            return self.socket.write(request)
+            size = self.socket.write(request)
+            self._last_frame_end = time.time()
+            return size
         return 0
 
     def _recv(self, size):
@@ -371,7 +384,9 @@ class ModbusSerialClient(BaseModbusClient):
         '''
         if not self.socket:
             raise ConnectionException(self.__str__())
-        return self.socket.read(size)
+        result = self.socket.read(size)
+        self._last_frame_end = time.time()
+        return result
 
     def __str__(self):
         ''' Builds a string representation of the connection
--- a/pymodbus/interfaces.py
+++ b/pymodbus/interfaces.py
@@ -151,6 +151,14 @@ class IModbusFramer(object):
         raise NotImplementedException(
             "Method not implemented by derived class")
 
+    def getResponseSize(self, message):
+        ''' Returns expected packet size of response for this request
+
+        :returns: The expected packet size
+        '''
+        raise NotImplementedException(
+            "Method not implemented by derived class")
+
 
 class IModbusSlaveContext(object):
     '''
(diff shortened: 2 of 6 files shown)


Cougar:
I added default getResponseSize() interface to the IModbusFramer class because only serial interface implements it. For example RTU over TCP works without it.

sjlongland:
This would definitely be worth pushing into the master... we just got bitten by this bug:

```
2015-10-21 09:15:56,765 _mm_io.modbus.modbus_pymodbus.conn_dev_ttySP2.worker[worker.py:  98] 15584/Thread-150 INFO Job BR9_A301_HHW_RT:W_RETURN_TEMP:a61af8ba-f7d5-4941-881c-7292a3226474 (read W_RETURN_TEMP from {ModbusPLCHoldingRegisterBlock: {ModbusPLCDevice: modbus.BR9_A301_HHW conn rtu baud[19200] unit 1} ModbusRegisterType.Holding/197..210}) expires in 66.362 sec
2015-10-21 09:15:56,779 vrt.da.source.modbus.BR9_A301_HHW.block-ModbusRegisterType.Holding/197-210[register.py: 241] 15584/Thread-150 INFO Polling block {ModbusPLCHoldingRegisterBlock: {ModbusPLCDevice: modbus.BR9_A301_HHW conn rtu baud[19200] unit 1} ModbusRegisterType.Holding/197..210} (197..210)
2015-10-21 09:15:56,847 pymodbus.transaction[transaction.py:  59] 15584/Thread-150 DEBUG Running transaction 91
2015-10-21 09:15:56,878 pymodbus.transaction[transaction.py:  66] 15584/Thread-150 DEBUG send: 0x1 0x3 0x0 0xc5 0x0 0xe 0xd4 0x33
2015-10-21 09:15:56,996 pymodbus.transaction[transaction.py: 222] 15584/Thread-150 DEBUG getting transaction 91
2015-10-21 09:15:57,007 pymodbus.transaction[transaction.py: 225] 15584/Thread-150 DEBUG got None(#-1), 0 transactions in queue
2015-10-21 09:15:57,027 pymodbus.transaction[transaction.py:  59] 15584/Thread-150 DEBUG Running transaction 92
2015-10-21 09:15:57,045 pymodbus.transaction[transaction.py:  66] 15584/Thread-150 DEBUG send: 0x1 0x3 0x0 0xc5 0x0 0xe 0xd4 0x33
2015-10-21 09:15:57,256 pymodbus.transaction[transaction.py:  76] 15584/Thread-150 DEBUG recv: 0x1 0x3 0x1c 0x68 0x15 0x91 0x7 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x1 0x0 0x0 0x0 0x18 0x0 0x0 0x0 0x18 0x15 0x9 0x1 0x3 0x1c 0x68 0x15 0x91 0x7 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x1 0x0 0x0 0x0 0x18 0x0 0x0 0x0 0x18 0x15 0x9
2015-10-21 09:15:57,284 pymodbus.factory[factory.py: 240] 15584/Thread-150 DEBUG Factory Response[3]: 0x1c 0x68 0x15 0x91 0x7 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x1 0x0 0x0 0x0 0x18 0x0 0x0 0x0 0x18
2015-10-21 09:15:57,306 pymodbus.transaction[transaction.py: 211] 15584/Thread-150 DEBUG adding transaction 0: ReadRegisterResponse (14)
2015-10-21 09:15:57,332 pymodbus.transaction[transaction.py: 213] 15584/Thread-150 DEBUG 1 transactions in queue
2015-10-21 09:15:57,355 pymodbus.factory[factory.py: 240] 15584/Thread-150 DEBUG Factory Response[3]: 0x1c 0x68 0x15 0x91 0x7 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x1 0x0 0x0 0x0 0x18 0x0 0x0 0x0 0x18
2015-10-21 09:15:57,403 pymodbus.transaction[transaction.py: 211] 15584/Thread-150 DEBUG adding transaction 0: ReadRegisterResponse (14)
2015-10-21 09:15:57,413 pymodbus.transaction[transaction.py: 213] 15584/Thread-150 DEBUG 2 transactions in queue
2015-10-21 09:15:57,420 pymodbus.transaction[transaction.py: 222] 15584/Thread-150 DEBUG getting transaction 92
2015-10-21 09:15:57,456 pymodbus.transaction[transaction.py: 225] 15584/Thread-150 DEBUG got ReadRegisterResponse (14)(#0), 1 transactions in queue
2015-10-21 09:15:57,484 vrt.da.source.modbus.BR9_A301_HHW[modbus_pymodbus.py: 446] 15584/Thread-150 DEBUG Read 4/197..210 Result: ReadRegisterResponse (14)
2015-10-21 09:15:57,493 vrt.da.source.modbus.BR9_A301_HHW[modbus_pymodbus.py: 466] 15584/Thread-150 DEBUG Raw data: 68159107000000000000000000000000000000010000001800000018
```

I've tweaked PyModbus code to give me some additional log messages, code is here: https://github.com/vrtsystems/pymodbus/tree/investigate-serial-bugs

We set the timeout small so that we improve performance, but then it trips up when the Modbus slave (in this case a Modbus→MBUS gateway) is a little slow with the response.

The other thing: I question the need for a full FIFO buffer for non-socket connections, which are inherently synchronous.  Seems to me, it would be perfectly reasonable to overwrite the old response with the latest one, as by rights, only one request should be processed at a time, and if we time out before the old reply arrives, we should discard it and only take the newest reply.

This would be consistent with the behaviour described here: http://www.modbus.org/docs/Modbus_over_serial_line_V1.pdf

sjlongland:
I found some test cases fail with this code branch.  Notably, there's no initialisation of members in __init__, so things go pear shaped if connect doesn't get called before write (shouldn't happen ordinarily, but does in test cases it seems), and the assumption that the socket implements an `inWaiting` method, which isn't the case for the `mockSocket` used.

https://github.com/vrtsystems/pymodbus/commits/cougar-dev is rebased off master and has these fixes added.

Cougar:
Thanks @sjlongland, I added your patch on top of my pull request.

itavero:
@bashwork We are still experiencing this issue. Can you please look at this PR?

mesias:
I made an fork from pymodebus to attend this issue. 
https://github.com/mesias/pymodbus

Not the best solution, but reads modbus using serial much faster than original pymodbus:
https://github.com/mesias/pymodbus/blob/master/pymodbus/transaction.py#L70

Pymodbus reads 1024 bytes every time, don't threat according response size:

`result = self.client._recv(1024)`
https://github.com/bashwork/pymodbus/blob/master/pymodbus/transaction.py#L68

Never undestood why @bashwork gave attetion for this issue.

sjlongland:
We're using `mbusd` at one site as a work-around for this issue.  Cougar's done some great work, but it still has issues when a Modbus device misbehaves.

On another project I wrote a `pyserial`-based interface to EDMI energy meters using Tornado as an asynchronous framework, and it has been considered whether we clean the `pyserial` bits up a bit and write our own Modbus interface as we're quickly running up against numerous performance issues with it.  (Hey, on 454MHz ARM CPUs you don't have many CPU cycles to spare!)

dhoomakethu:
Closed as part of PR #164

PR #56, opened 2014-08-16
FIX make commit c0ae343 compatible with Python 2.6

Cougar:
If Python version is older than 2.7, do not use source_address
parameter in socket.connect() (added with commit c0ae343)

https://docs.python.org/2/library/socket.html#socket.create_connection

### DIFF
--- a/pymodbus/client/sync.py
+++ b/pymodbus/client/sync.py
@@ -1,3 +1,4 @@
+import sys
 import socket
 import serial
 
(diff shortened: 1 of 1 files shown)


bashwork:
I am a little torn on this. I would like to work towards only supporting 2.7 and 3.x series python versions. I could include this, but I would like it to be included as a compatibility shim::

if version == "2.7":
  compat.create_connection = socket.create_connection
elif version == "2.6":
  compat.create_connection = ...

Cougar:
This sounds good. I see that there is an attempt to add such layer in python3 branch but this branch is not updated since 2013. All new commits still go directly to master which is Python2 only. Any suggestion how to proceed? Add similar file to current master branch or do you plan to sync master and python3 branches any time soon?

dhoomakethu:
Very OLD PR, officially pymodbus is supported for pyhton 2.7 and beyond only.

PR #61, opened 2014-09-06
Logging of packets' content

jkedra:
Adds logging of incoming and outgoing packets' content to
ModbusRtuFramer.processIncomingPacket and builtPacket.

Also added logging of incoming packets to:
ModbusAsciiFrammer.processIncomingPacket
ModbusBinaryFrammer.processIncomingPacket.

Above makes them consistent with ModbusSocketFramer.

### DIFF
--- a/pymodbus/transaction.py
+++ b/pymodbus/transaction.py
@@ -530,6 +530,8 @@ class ModbusRtuFramer(IModbusFramer):
         :param data: The new packet data
         :param callback: The function to send results to
         '''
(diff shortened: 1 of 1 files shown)


jkedra:
Hello,
Is there anything more I could to for you to merge this into master?
Jurek

bashwork:
It looks like this is already being done in the transaction manager. Can we move all this code to there once instead of recreating it in each framer?

jkedra:
OK you're right. That's obvious redundancy. I will check it out and
correct.

PR #64, opened 2014-09-30
ReportSlaveId: use correct slaveid/additional data

karlp:
The Modbus application spec isn't very clear, but the "slave id" field
in ReportSlaveId response is a single byte, then the single byte of
0x00/0xff for status, _then_ the trailing arbitrary data.  Likewise,
the byte count doesn't include either slave id field.

Further, the standard code was passing an extra argument to the execute
method, so it wasn't even getting as far as doing the wrong thing
normally.

This behaviour verified against a Frer and Schneider real world power
meters, and also the behaviour of libmodbus for the length.

Signed-off-by: Karl Palsson karlp@remake.is

(Notes: I don't know the appropriate way of validating that the identifier as passed is only a single byte)

### DIFF
--- a/pymodbus/other_message.py
+++ b/pymodbus/other_message.py
@@ -358,13 +358,15 @@ class ReportSlaveIdRequest(ModbusRequest):
         '''
         pass
 
-    def execute(self):
+    def execute(self, slave_object=None):
         ''' Run a read exeception status request against the store
 
         :returns: The populated response
         '''
-        identifier = '\x70\x79\x6d\x6f\x64\x62\x75\x73'
-        return ReportSlaveIdResponse(identifier)
+        # TODO - get this information from the slave_object, if provided
+        identifier = '\x55'
+        extra_data = '\x70\x79\x6d\x6f\x64\x62\x75\x73'
+        return ReportSlaveIdResponse(identifier, True, extra_data)
 
     def __str__(self):
         ''' Builds a representation of the request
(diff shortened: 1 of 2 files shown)


karlp:
Is there something wrong with this?  Do you have changes you'd like me to make?  I don't see anything in the recent commits that indicates this issue has been resolved in an alternative way.

dhoomakethu:
@karlp, I closed this as part of cleaning up of old issues, Please refer commit 336c46c for a recent change on `ReportSlaveIdResponse`. I am not sure if this fixes the issue mentioned here, if not please create a new issue with logs/expected response etc. 

Thanks

karlp:
Please, actually read my comments explaining this.  Please then compare to the code you've still got.  Clearly it's still not correct.  You've fixed the _existing_ broken behaviour to be _self consistent_ which is good, but it's not right.

dhoomakethu:
I am sorry but I am not able to understand the issue here,  As per the specs the behaviour of this function is device specific and there is no generic way we can handle this.
screen shot 2017-12-01 at 10 44 03
We are currently kind of ignoring the additional data field and consider the last byte in the response to be the `device status`.  Is this the issue you are talking here ?
 
```
self.byte_count = byte2int(data[0])
self.identifier = data[1:self.byte_count + 1]
status = byte2int(data[-1])
self.status = status == ModbusStatus.SlaveOn
```

PR #125, opened 2016-06-20
Port to Python 3: exception handlers, izip -> zip

ocschwar:


dhoomakethu:
python3 porting is done and is supported with pymodbus 1.3.0 and above

PR #130, opened 2016-07-26
fix ModiconPayloadDecoder __init__ fail

semyont:
Decoder fails with a missing endian attribute, needed for struct string decode.


semyont:
Tested in production.

PR #137, opened 2016-10-21
Minor tweaks.

bje-:


bje-:
Gaylen, there is a useful patch in 3d565eb. I would appreciate if you could accept it. Cheers.

PR #140, opened 2016-11-16
Enhancement ModbusTcpClient Timeout

semyont:
Today Scenario: when creating a TCP Client the Defaults timeout is Always used 
Enhancement Scenario: pool of clients with different timeouts (more like a UDP client that accepts different timeout on context)


semyont:
Will updating python 3 branch
`except socket.error, msg:`
to
`  except socket.error as msg:`
will pass the tests ?

PR #143, opened 2016-12-07
Recall socket recv until get a complete response.

ccatterina:
Socket recv waits until it gets some data from the host but
not necessarily the entire response that can be fragmented in
many packets. In this case it receives an incomplete frame
that is rejected by the frame handler.
This commit solves this issue by recalling socket recv until
get a complete modbus response.

### DIFF
--- a/pymodbus/transaction.py
+++ b/pymodbus/transaction.py
@@ -68,7 +68,7 @@ class ModbusTransactionManager(object):
 
     def _calculate_response_length(self, expected_pdu_size):
         if self.base_adu_size == -1:
-            return 1024
+            return None
         else:
             return self.base_adu_size + expected_pdu_size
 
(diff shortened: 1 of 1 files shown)


bje-:
+1

bje-:
This seems like a rather important patch (which seems to work well, for me at least). Without it, pymodbus for TCP slaves hardly ever works. My read register calls return None a lot of the time.  I am amazed no one else sees this issue.  I am happy to do the work to bring it up to line with master.

ccatterina:
Hi @bje- , i have already updated the commit in a branch of my repository but i have not tested it yet. I will update the PR as soon as possible.

ccatterina:
Updated.

dhoomakethu:
Thank you, I will add this to 1.3.1 milestone.

dhoomakethu:
would also partially fix #175

dhoomakethu:
@ccatterina  Just to highlight a bug here with this fix, when there is an exception received from the modbus device, client will be struck in an eternal while loop or ends with socket error ,resulting in a None being returned back. I will have to revert this change.

ccatterina:
Oh, rigth, i'm sorry that i didn't notice the bug before. I will try to fix this as soon as possible, meanwhile, revert it.

PR #74, opened 2015-03-01
PATCH: boost serial speed 3x or more #python3

Cougar:
Python3 support for fast serial RTU. Long description can be found in pull request #55.

PR #150, opened 2017-01-31
Patch: Extended timeout handling in ModbusTransactionManager

jackjweinstein:
Hi bashwork,

Long time no chat, hope you're well. I'm finally getting around to the PR you and I discussed last June.

Problem statement is, we're using ModbusSerialClient to talk to a solar inverter that responds to write holding registers requests with about a 1s delay. Using current pymodbus code, I either need to use a timeout >>1 s, or the FIFOTransactionManager will catch the inverter's response _after_ returning from read_holding_registers. If I use a long timeout, the call to `self.client._recv(1024)` will take the entire timeout to return; if not, I end up with a log file like [this one](http://www.jackweinste.in/s/2017-01-30_pymodbus_response_delay.txt). Note that at 2017-01-30 14:50:28,723 I submit a WriteHoldingRegistersRequest and its response is returned for the next call to read_holding_registers.

The truly correct way to handle this, as you mention in your comment in `ModbusTransactionManager.execute()`, would be to calculate the number of bytes we expect the response to contain, and call `self.client._recv(num_bytes)` instead of `self.client._recv(1024)`. That way we could both use a long socket timeout and avoid blocking for the whole timeout when a valid response is received.

I've been waiting to PR until I found time to implement that response length prediction. But I've been busy and it'd be some actual work... and I just haven't gotten to it.

Until then, would you accept this patch? I've reduced the serial socket's timeout and added a loop to poll it multiple times, handling the slow response while not affecting normal behavior.

### DIFF
--- a/pymodbus/client/sync.py
+++ b/pymodbus/client/sync.py
@@ -303,6 +303,8 @@ class ModbusSerialClient(BaseModbusClient):
         self.parity   = kwargs.get('parity',   Defaults.Parity)
         self.baudrate = kwargs.get('baudrate', Defaults.Baudrate)
         self.timeout  = kwargs.get('timeout',  Defaults.Timeout)
+        
+        self.socket_poll_rate = 0.1  # seconds -- magic number for how often to check the bus for data
 
     @staticmethod
     def __implementation(method):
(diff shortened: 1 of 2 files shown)


dhoomakethu:
@jackjweinstein , there are some similar PR's addressing the issue you have solved here #74 #55 #161 . I would be keeping your PR on hold . Once a new pymodbus is release 1.3.0, let me know if the issue still persists.

jackjweinstein:
@dhoomakethu sounds great. I'm glad to see how active you've been in development. To be honest I haven't been using Modbus much since submitting the PR, but I'll give 1.3.0 a try and see how it goes!

Thanks -

Jack

dhoomakethu:
Thanks. JFYR, the latest available on pip is 1.3.0rc2

dhoomakethu:
@jackjweinstein I will be closing this PR , the latest pymodbus has already the fix you are trying to submit. If you think that could be improved or find any bug, feel free to raise an issue or submit a new PR. Thanks for the PR and appreciate your efforts.

PR #158, opened 2017-04-24
Prevalidate addresses

CraigMarkwardt:
Optimization:
      prevalidate requested addresses;
    Background
      the validate() method doubles the transaction times on the ModBus bus;
      on slow ModBus buses, this can take many seconds for one request transaction;
      this increases bus loading and potential to miss data from target;
    Remedy
      the prevalidate() method will register an address as a known valid address;
      pre-validated addresses will be queried immediately for their value;
      addresses that successfully validate once will be marked prevalid;
      if a request does fail, the prevalidation notation is removed;
      this allows devices to appear and disappear from the bus without too much
      disturbance;

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -123,7 +123,7 @@ License Information
 Pymodbus is built on top of code developed from/by:
   * Copyright (c) 2001-2005 S.W.A.C. GmbH, Germany.
   * Copyright (c) 2001-2005 S.W.A.C. Bohemia s.r.o., Czech Republic.
-  * Hynek Petrak <hynek@swac.cz>
+  * Hynek Petrak, https://github.com/HynekPetrak
   * Twisted Matrix
 
 Released under the BSD License
--- a/doc/sphinx/examples/custom-datablock.rst
+++ b/doc/sphinx/examples/custom-datablock.rst
@@ -1,0 +1,6 @@
+==================================================
+Custom Datablock Example
+==================================================
+
+.. literalinclude:: ../../../examples/common/custom-datablock.py
+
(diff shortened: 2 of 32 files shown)


dhoomakethu:
Could you please provide some real time data regarding the timing optimisations achieved because of this fix .

CBGoodBuddy:
We have two devices on one RS-485 bus.  The first device responds to a single query in about 1400ms the other in about 250ms.  Yes, the first device is very slow.
The baseline pymodbus code performs an operation as a "verify" followed by actual query, so two query transactions.  The revised code will perform one verify at the very first transaction, then only the actual query.  So, excluding the first verify, the revised code will require about 50% less time than the baseline code.

Device      |  Baseline  | Revised
----------|-----------|---------
Device 1   |  2850 ms | 1400 ms
Device 2  |  550 ms    |  250 ms

This is important because for device 1 we perform about 12 transactions per minute, and 25-30 transactions per minute on device 2.    Without removing the "verify" portion of the bus traffic, we run into situations where we lose data or can't get a command to the device in a timely manner.

dhoomakethu:
Could you please give a try with latest pymodbus available (1.3.0rc1) and see if that version has addressed the issue you are facing ?

karlp:
It sounds like validate() itself should simply be dropped.  it's rather absurd to make duplicate requests, and can even result in data changing

CBGoodBuddy:
The version 1.3.0rc1 does not solve the problem.  Since validate() is being called by *_read_message.py there are still two Modbus transactions that ~double the total time.  I agree with @karlp that removing the validate() is a reasonable option.  I've seen one or two cases in our setup (before I made my changes), where the validate() succeeded but the getValues() did not, so validate() does not guarantee success.

dhoomakethu:
@CBGoodBuddy please change the destination branch to `dev` instead of `master`

PR #173, opened 2017-05-27
* register_write_message.py: Observe skip_encode option when encoding a single register request.

bje-:


pymodbus/register_write_message.py
```diff
@@ -35,9 +35,12 @@ def encode(self):
 
         :returns: The encoded packet
         '''
+        packet = struct.pack('>H', self.address)
         if self.skip_encode:
-            return self.value
-        return struct.pack('>HH', self.address, self.value)
+            packet += self.value
+        else:
+            packet += struct.pack('>H', self.value)
```
dhoomakethu: This line would throw `TypeError: cannot concatenate 'str' and 'int' objects` on python2 and `TypeError: can't concat bytes to int` on python 3.

The sample code to test this feature is below
`
from pymodbus.client.sync import ModbusTcpClient as ModbusClient
client = ModbusClient('127.0.0.1', port=5020)
client.connect()
address = 0x00
client.write_register(address, 25, unit=1, skip_encode=True)
rs = client.read_holding_registers(address, 1,  unit=1)
assert rs.registers = [25]
`

dhoomakethu:
please add a relevant test for your change for encode method .

dhoomakethu:
Just to highlight a bug here with this fix, when there is an exception received from the modbus device, client will be struck in an eternal while loop or ends with socket error ,resulting in a None being returned back. I will have to revert this change.

PR #180, opened 2017-06-06
WIP: Recall socket recv until get a complete response.

ccatterina:
Socket recv waits until it gets some data from the host but not necessarily the entire response that can be fragmented in many packets. In this case it receives an incomplete frame that is rejected by the frame handler.
This commit solves this issue by recalling socket recv until get a complete modbus response.

### DIFF
--- a/pymodbus/transaction.py
+++ b/pymodbus/transaction.py
@@ -68,9 +68,37 @@ class ModbusTransactionManager(object):
 
     def _calculate_response_length(self, expected_pdu_size):
         if self.base_adu_size == -1:
-            return 1024
+            return None
         else:
             return self.base_adu_size + expected_pdu_size
+
+    def _calculate_exception_length(self):
+        ''' Returns the length of the Modbus Exception Response according to
+        the type of Framer.
+        '''
+        if isinstance(self.client.framer, ModbusSocketFramer):
+            return self.base_adu_size + 2  # Fcode(1), ExcecptionCode(1)
+        elif isinstance(self.client.framer, ModbusAsciiFramer):
+            return self.base_adu_size + 4  # Fcode(2), ExcecptionCode(2)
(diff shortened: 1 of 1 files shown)


pymodbus/transaction.py
```diff
@@ -68,32 +68,70 @@ def _set_adu_size(self):
...
+        elif isinstance(self.client.framer, (ModbusRtuFramer, ModbusBinaryFramer)):
+            return self.base_adu_size + 2  # Fcode(1), ExcecptionCode(1)
+
+        return None
+
+    def _check_response(self, response):
+        ''' Checks if the response is a Modbus Exception.
+        '''
+        if isinstance(self.client.framer, ModbusSocketFramer):
+            if len(response) >= 8 and ord(response[7]) > 128:
```
dhoomakethu: Just an early heads up, do not use `ord` directly , instead use `byte2int` from `pymodbus.compat` module, this way the code works fine in both python version.

ccatterina: Ok, thanks for the advice. 
In these days i'm a bit busy, probably i will submit the final PR by the next week because i still have to test it and review the chages.

dhoomakethu: Sure, Thanks.

ccatterina:
Hi, i tested the patch using RTU and Socket framers and it seems to work.
I have not devices that use Binary or ASCII framers and i don't know much about Modbus ASCII and Modbus Binary, so if someone could test them, that'd be great.

dhoomakethu:
Thanks @ccatterina , I will give a test and merge if everything seems good.

dhoomakethu:
@ccatterina I could only test on real device with RTU and TCP modes , I will take the liberty that most of the people out there aren't using the binary/ascii modes and hope it will work for them as well. Thanks for your efforts

PR #195, opened 2017-07-17
Update callback-server.py

bondarevky:
Using Thread instead of Process in line 131. Otherwise it leads to the ImportError of multiprocessing library


dhoomakethu:
what particular error are you referring to ?

dhoomakethu:
Closing , please refer comment above

PR #203, opened 2017-08-08
Fix for hang when using serial comms

w4kpm:
This would work for me about 99% of the time, but then it would hang in this section of code. 

If I understand right, it would get part of the message in result in the first call to self.client._recv, but not the whole thing and then loop in this section would keep calling _recv, getting nothing - so there would still be something in result but not the expected length. self.client._recv call would honor the timeout, but because of the logic it would just spin here.

I added a secondary timeout value to break out of this local loop after the number of retries was reached and this solved the immediate problem for me - I have no idea what was actually being sent over the wire to get in this state in the first place.


w4kpm:
Looks like this section has been reworked in the release candidate in a way that would fix this - I'll try to give it a spin.

dhoomakethu:
Ref#186

dhoomakethu:
@w4kpm do you still think this PR needs to be taken care or the fix with 1.3.2rc1 do solve the issue you are addressing with this PR ?

w4kpm:
Sorry I haven't had a chance to try out 1.3.2rc1 just yet - I'm still using
my patch.

I'm kinda busy the next couple of days, but I'll try to get a test in.

Kelly

dhoomakethu:
@w4kpm I haven't heard anything on this from a long time, I assume it is fixed with 1.3.2 and above. Please give a try on the changes on the `dev` branch if you still think 1.3.2 does not fix this issue.

I will be closing this PR as of now.

PR #205, opened 2017-08-17
Add new skip_bytes method to BinaryPayloadDecoder to skip over bytes

bje-:
It is clearer in client code to use a method like this than reading a string of n bytes and discarding the return value.


dhoomakethu:
Please change the destination branch for this PR to `dev` instead of `master`

dhoomakethu:
Thanks , Could you also please modify the example [here](https://github.com/riptideio/pymodbus/blob/dev/examples/common/modbus-payload.py) so that it would be easier for others to use this feature ?

PR #206, opened 2017-08-17
Add new skip_bytes method to BinaryPayloadDecoder to skip over bytes

bje-:


bje-:
No need for this PR now.

Issue #212, opened 2017-09-08
InvalidMessageRecievedException when try to communicate with a modbus slave device (RTU)


thaxy:
Hello,

When I try to read modbus registers from a slave through a serial port via pymodbus:
I get the following error:

```
modbus_master_1  | [11:53:53.504] DEBUG [__main__.<module>:61] Initializing the client
modbus_master_1  | [11:53:53.520] DEBUG [master.config.validator._validate_with_yaml_schema:54] Config file valid according to yml schema.
modbus_master_1  | [11:53:53.520] DEBUG [master.config.core._load:63] Finished loading yaml file. Config is valid
modbus_master_1  | [11:53:53.521] DEBUG [__main__.__init__:31] Initializing serial port: '/dev/ttyS0', {'baudrate': 9600, 'parity': 'N', 'bytesize': 8, 'timeout': 1.0, 'stopbits': 1}
modbus_master_1  | [11:53:53.522] DEBUG [pymodbus.client.async.connectionMade:74] Client connected to modbus server
modbus_master_1  | [11:53:53.522] INFO [master.protocol.connectionMade:33] [/dev/ttyS0] Connection made. Reading settings are [{'start_address': 0, 'offset': 30001, 'device_id': 1, 'count': 10}]
modbus_master_1  | [11:53:53.522] DEBUG [__main__.<module>:92] Starting the reactor
modbus_master_1  | [11:53:54.524] DEBUG [master.protocol.fetch_modbus_table:42] [/dev/ttyS0] Start to fetch modbus table(s)
modbus_master_1  | [11:53:54.524] DEBUG [pymodbus.transaction.addTransaction:309] adding transaction 1
modbus_master_1  | [11:53:54.585] DEBUG [pymodbus.factory._helper:238] Factory Response[1]
modbus_master_1  | Unhandled Error
modbus_master_1  | Traceback (most recent call last):
modbus_master_1  |   File "/usr/local/lib/python3.5/site-packages/twisted/python/log.py", line 103, in callWithLogger
modbus_master_1  |     return callWithContext({"system": lp}, func, *args, **kw)
modbus_master_1  |   File "/usr/local/lib/python3.5/site-packages/twisted/python/log.py", line 86, in callWithContext
modbus_master_1  |     return context.call({ILogContext: newCtx}, func, *args, **kw)
modbus_master_1  |   File "/usr/local/lib/python3.5/site-packages/twisted/python/context.py", line 122, in callWithContext
[...]

PR #213, opened 2017-09-11
Do not trigger InvalidMessageRecievedException for incomplete frame 

thaxy:
**The following changes retrieve a valid response object but need to be reviewed.**

It is not clear why a ``while`` loop was used. Nothing seems to change the buffer within this block.
Function ``dataReceived`` of ``twisted.internet.protocol.Protocol`` triggers ``processIncomingPacket`` everytime data was read and appends this to the
buffer.

Please review the following changes. Especially the ``self._process`` calls with the ``kwarg error=True``.

Further questions:
* Do we need to reset the buffer?
* What happens if the client reads only a chunk of the frame and the cable gets unplugged?
  The next time we request new registers we would append the data from the second call to the present buffer which leads to an invalid frame.

#212

### DIFF
--- a/pymodbus/transaction.py
+++ b/pymodbus/transaction.py
@@ -460,7 +460,7 @@ class ModbusSocketFramer(IModbusFramer):
 
     def _process(self, callback, error=False):
         """
-        Process incoming packets irrespective error condition 
+        Process incoming packets irrespective error condition
         """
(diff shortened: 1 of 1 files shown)


dhoomakethu:
Please raise the PR against the `dev` branch instead of `master`

Issue #210, opened 2017-08-27
Release tar ball without generated files?


ghost:
Hi,

while I think, that for some users it is convenient to have a release .tar.gz file with all documentation pre-build, it makes life slightly harder for redistributors (like me, working on the Debian pymodbus package). For Debian packages (and probably most other Linux distributions), it is required, that all non-source files are build from source. One can, of course, just remove the files and rebuild them, but having a pristine tar file from you would be preferable. (E.g. users can easily check, that Debian distributes the original sources by comparing the checksum of the tar.gz).

Many thanks in advance!

PS: Also, I wonder wether maintaining generated files in version control is a good idea? In my experience, it is not, but YMMV, of course.

dhoomakethu:
The release is entirely controlled by GITHUB as of now, All I am doing is just click the button `Draft a new release`, create a tag and publish. I will have to go through the release documentation to fine tune it (to what you are asking). The majority of the users use the package via PIP and I know only a few people who depends on the release here on the github repo (cc. @markmevans )

MartinBorg:
I think, that both the pypi and the github archive have the pre-compiled documentation included, which is probably, because it is already in the version control.

markmevans:
I also require a source tarball to build from.  I ignore the generated documentation so this specific issue does not directly affect me but agree that it really does not need to be under source control.

I'd vote to remove the generated documentation files under source control and fixing the hooks to  [readthedocs.org]https://readthedocs.org/projects/pymodbus/) but I don't know what that entails.

See: http://dont-be-afraid-to-commit.readthedocs.io/en/latest/documentation.html

dhoomakethu:
@markmevans , JFYI  the latest documentation is upto date here [pymodbus-n](http://pymodbus-n.readthedocs.io/en/latest/), I do not have access to the original pymodbus documentation

dhoomakethu:
@MartinBorg @markmevans I have added a gitattributes files which should exclude the documentation/tests and other un releated stuff the generate release tarball. Refer PR #223 .

Please let me if this addresses the issue you raised in this issue.
[.gitattributes](https://github.com/riptideio/pymodbus/pull/223/files#diff-fc723d30b02a4cca7a534518111c1a66)

MartinBorg:
[...]

PR #223, opened 2017-10-06
#210 gitattributes to exclude files from release

dhoomakethu:
Exclude examples, docs, script, test folders from the release.


dhoomakethu:
Will be tracked with a new branch tar-ball-without-docs

PR #226, opened 2017-10-13
Async serial  using schedulers

dhoomakethu:
Async clients with Twisted/Tornado/Asyncio


dhoomakethu:
Will be tracked with branch `async`

Cougar:
I run small program that collects data from different sources (RS-485 Modbus slaves via serial ports, Modbus slaves via TCP/IP, Mbus devices etc), does different calculations, sends data to other systems and also has its own REST API with websocket support for live streaming. In my case, the main loop is Tornado based.

Sleeps here don't help much. When slave doesn't answer in any reason (restart, bug, communication loss etc) then this loop becomes infinite anyway and no code can break it. Modbus slave doesn't send anything itself and without sending new commands everything stalls in this.

This loop should read data only when it is available and return back to main ioloop otherwise. After some timeout there is probably not necessary to check these bytes at all. I'm not sure if it should exit without any action or with some exception.

I try to familiarize myself with the async logic in this change but it is way too different from the one that I once made and it takes some time to understand enough to provide PR. I'm sure, You have some ideas how to avoid this loop :)

dhoomakethu:
Got it. Will see what could be done.

dcneeme:
Hi dhoomakethu, 
I'm also very interested in fully functional async pymodbus. Assuming you are testing on software based slave simulator, are you interested in physical slaves (that may sometimes skip an answer)? This way the possible real world situations are better matched. I would happy to provide one physical RTU slave and one TCP server device to support your development, if it helps to resolve the serial problem faster. Pls write me neeme[at]takis.ee if interested.

Issue #135, opened 2016-10-18
Thread Test


OptrixAU:
It would be nice if PyModbus would quickly check to see if it was running on the main thread before calling 'Run' in StartTcpServer.

If it IS being launched in a thread, you could then add 'installSignalHandlers=False' to avoid an exception.

PR #242, opened 2017-11-24
#135: Thread Test

rahulraghu94:
In the start TCP, UDP and Seriel server, we are running a quick check to see weather the thread in which the helpers are executed are the
Main thread or the spawned thread.

This infomation is passed to reactor.run() to the flag
installSignalHandlers. True is passed if the thread is a main
thread, else false.

This is done to prevent errors when running the helpers from a
spawned thread.


dhoomakethu:
Fix the coverage

rahulraghu94:
@dhoomakethu , changed the base to `dev`

Issue #84, opened 2015-05-18
Stop async Server


Loris1123:
I didn't find a way to stop an async TCP Server. I need this to test the server with different configuration in a single script.

To stop the async server I added the following code to async.py:

```
def StopTcpServer(is_mainthread=True):
    from twisted.internet import reactor
    if is_mainthread:
        reactor.stop()
    else:
        reactor.callFromThread(reactor.stop)
```

It seems like it is working for me. Maybe you should include something similar to pymodbus.

tmcolby:
I needed the ability to stop the server also.  Thank you, Loris1123.  Your solution works for me as well.  I agree that it would be nice to have this incorporated into pymodbus.

MM8704:
Hello, I have a question if I have a button that stops the server with the Stopserver () function, how can I start it again without closing the program?

dhoomakethu:
The `reactor` is not restartable . So I am not sure if there is way to restart it again with out calling `StartTcpServer` method.

MM8704:
Thanks for your answer dhoomakethu, I have to implement a method able to change the IP address of the server, without closing the main program.

Initially try to stop the server and start it again with the StarTcpServer method with a new Ip address, but when you stop the reactor and start it again, an error appears. Any recommendation?.

dhoomakethu:
@MM8704 could you share the error and any code to reproduce your use case?

MM8704:
@dhoomakethu sure, is a small example of a server mounted on kivy

Please download it in google drive: 
https://drive.google.com/open?id=10_VHhRWViiBYwp_OLX20VHzIF8Jo7WVk

dhoomakethu:
You can take a look at [modbus-simulator](https://github.com/riptideio/modbus-simulator. It is also based on pymodbus and kivy supports both TCP and Serial servers (uses pymodbus sync server and threads though). You can find the backend code [here](https://github.com/riptideio/modbus-simulator/blob/master/modbus_simulator/utils/pymodbus_server.py)

MM8704:
Hi @dhoomakethu, without any positive, have you found something?

dhoomakethu:
@MM8704 sorry but your code throws me error while running
`     on_press: Server_TCP.stop(self)
 TypeError: unbound method stop() must be called with Server_TCP instance as first argument (got ToggleButton instance instead)`. I would suggest to use the modbus-simulator mentioned above for the time being. I will take a closer look in to your code when I get some free time.

[...]

PR #243, opened 2017-11-24
#84: Stop Asynchronous Server

rahulraghu94:
New helper function to help stop Asynchronous server are being added.
Calling this helper function will kill the main thread

### DIFF
--- a/pymodbus/server/async.py
+++ b/pymodbus/server/async.py
@@ -276,6 +276,9 @@ def StartSerialServer(context, identity=None,
 
 
 def StopServer():
+    """
+    Helper method to stop Async Server
+    """
     from twisted.internet import reactor
     if _is_main_thread():
         reactor.stop()
--- a/pymodbus/server/async.py
+++ b/pymodbus/server/async.py
@@ -274,9 +274,23 @@ def StartSerialServer(context, identity=None,
     SerialPort(protocol, port, reactor, baudrate)
     reactor.run()
 
+
+def _stop_server(is_mainthread=True):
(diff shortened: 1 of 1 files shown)


pymodbus/server/async.py
```diff
@@ -274,9 +274,23 @@ def StartSerialServer(context, identity=None,
...
     reactor.run()
 
+
+def _stop_server(is_mainthread=True):
+    from twisted.internet import reactor
+    if _is_main_thread():
+        reactor.stop()
+    else:
+        reactor.callFromThread(reactor.stop)
+    _logger.debug("Stopping main thread")
```
dhoomakethu: @rahulraghu94 This log message will be displayed irrespective of whether the current thread is main or child.

rahulraghu94:
@dhoomakethu , changed the base to `dev`

pymodbus/server/async.py
```diff
@@ -274,9 +274,23 @@ def StartSerialServer(context, identity=None,
     SerialPort(protocol, port, reactor, baudrate)
     reactor.run()
 
+
+def _stop_server(is_mainthread=True):
```
dhoomakethu: THe param `is_mainthread=True` is not being used, remove it if unnecessary

pymodbus/server/async.py
rahulraghu94: Isn't that expected behaviour?

rahulraghu94: Oh, sorry. I'm on some meds 📦 
Fixing...

Issue #245, opened 2017-11-28
Bring unit test code coverage to 90%


rahulraghu94:
### Current coverage:

```python
 # code and logs here.
Name                                             Stmts   Miss  Cover   Missing
------------------------------------------------------------------------------
pymodbus/__init__.py                                12      4    67%   25-28
pymodbus/bit_read_message.py                        74      0   100%
pymodbus/bit_write_message.py                       97      0   100%
pymodbus/client/__init__.py                          0      0   100%
pymodbus/client/async.py                            73      1    99%   157
pymodbus/client/common.py                           39      0   100%
pymodbus/client/sync.py                            182      9    95%   344, 370-372, 379, 384-385, 389, 403
pymodbus/compat.py                                  44      3    93%   94-96
pymodbus/constants.py                               39      0   100%
pymodbus/datastore/__init__.py                       5      0   100%
pymodbus/datastore/context.py                       56      0   100%
pymodbus/datastore/database/__init__.py              3      0   100%
pymodbus/datastore/database/redis_datastore.py      86      0   100%
pymodbus/datastore/database/sql_datastore.py        67      1    99%   113
pymodbus/datastore/remote.py                        35      1    97%   84
pymodbus/datastore/store.py                         69      0   100%
pymodbus/device.py                                 164      0   100%
pymodbus/diag_message.py                           227      8    96%   72-74, 122, 725-729
pymodbus/events.py                                  60      0   100%
pymodbus/exceptions.py                              30      2    93%   91-92
pymodbus/factory.py                                 78      0   100%
pymodbus/file_message.py                           147      0   100%
pymodbus/interfaces.py                              46      0   100%
pymodbus/internal/__init__.py                        0      0   100%
pymodbus/internal/ptwisted.py                       20      3    85%   29, 33-34
pymodbus/mei_message.py                             75      1    99%   137
pymodbus/other_message.py                          148      0   100%
pymodbus/payload.py                                154      0   100%
pymodbus/pdu.py                                     73      0   100%
pymodbus/register_read_message.py                  129      2    98%   47, 297
pymodbus/register_write_message.py                 133      2    98%   72, 149
pymodbus/server/__init__.py                          0      0   100%
pymodbus/server/async.py                           134     52    61%   58-63, 70-84, 91-96, 163-168, 175-189, 197-201, 270
pymodbus/server/sync.py                            240      7    97%   62-65, 303-304, 457
[...]

PR #249, opened 2017-12-04
#245 unittest coverage

rahulraghu94:
## All tests at 90% and above.


examples/contrib/message-generator.py
```diff
@@ -32,14 +32,14 @@
...
-
+from pymodbus.pdu import ExceptionResponse
 #--------------------------------------------------------------------------#
 # initialize logging
 #--------------------------------------------------------------------------#
 import logging
 modbus_log = logging.getLogger("pymodbus")
 
-
+exp = ExceptionResponse(0x01, 0x0B)
```
dhoomakethu: Is this being supported ? Generation of exception response?

rahulraghu94: No, this was supposed to go away in a cleanup.

PR #253, opened 2017-12-20
Release 1.4.0

dhoomakethu:
1.4.0 Pre release
Wait till #252 and #249 is merged in to dev


pymodbus/server/async.py
```diff
@@ -58,7 +58,9 @@ def dataReceived(self, data):
         if _logger.isEnabledFor(logging.DEBUG):
             _logger.debug(' '.join([hex(byte2int(x)) for x in data]))
         if not self.factory.control.ListenOnly:
-            self.framer.processIncomingPacket(data, self._execute)
+            unit_address = byte2int(data[0])
+            if unit_address in self.factory.store:
+                self.framer.processIncomingPacket(data, self._execute)
```
karlp: This completely breaks using unit ids with modbusTCP.  0 is only the correct offset for RTU framed packets.  You've added a fixed RTU offset into the core server code.

karlp: Why on earth isn't this just being handled in the framer, it seems to be already handled there anyway, and even if the address doesn'tmatch on serial, you still need to swallow all the remaining bytes on the wire.

dhoomakethu: @karlp there is an issue already for that #256  and there is also a branch with the fix https://github.com/riptideio/pymodbus/compare/%23256-Fix-TCP-Server-Transactions

If you think you have a better solution please raise a PR rather than complaining. You could have searched for the issues before commenting here. Also note, it will be part of the upcoming release . I hope this answers you.

karlp: If your merging and committing work had been cleaner, it would have been feasible to review your changes at the time.  There was no way anyone could sanely attempt to review things like PR #253 so it's not surprising that bugs crept in.  Please consider what your patches look like, not just commit and move forward.

dhoomakethu: ok, thanks for the feedback. I will keep that in mind. Your are free to fix them and raise a PR or raise an issue so others can take care of it when ever time permits.

Issue #82, opened 2015-05-03
Add more convenient API for checking if response is error?


estan:
In some of the examples I've seen e.g:

```
assert(rq.function_code < 0x80)
```

Is this the canonical way to check if a response is an error or not? If so, what's the best way to pinpoint what kind of error it was? Check the `exception_code` against `ModbusExceptions`?

Note sure how the API could be improved, but would be nice if there were some more examples of error handling in the examples (which are great!).

PR #258, opened 2018-01-08
#82: Error response API

rahulraghu94:
Added API in response to allow users to easily to check for errors in stead of having to manually compare response codes.

Eg:
        response = client.write_registers(...)
        assert(response.isError() == False)


dhoomakethu:
@rahulraghu94 Please update the examples with the new the error api's . I missed to update the comment. You can push the changes to the same branch and raise a PR. Thanks

Issue #250, opened 2017-12-05
To scrape a real PLC in realtime and use that data to create NEW updating server


MazoSergey:
### Versions

* Python: 2.7
* OS: Linux
* Pymodbus: 1.3.2
* Modbus Hardware (if used): RSIM 

### Description

I am trying to make a clone of a real PLC with a help of Scraper example, and then use that data to create MODBUS server... If i understand correct I have to: scrape  a real PLC --> save the data to file --> create server --> load data from file to server. It looks like I do not understand how this staff works :( If it is possible please help me, may be some examples of this kind of data flow. Thank you!

### Code and Logs

```python
 # code and logs here.

```

dhoomakethu:
@MazoSergey what exactly is the problem ? Did you run the scraper example agains the PLC and encountered any error ?

MazoSergey:
@dhoomakethu thank you for support! Yes, I tried to run example, I've got a file (dump of data from PLC). Now I am trying to run "Modbus Simulator Example" with -c option, and getting "Configuration Error: Invalid Configuration" :(

dhoomakethu:
Could you upload the saved the configuration here please ?

MazoSergey:
For sure! Here it is:
[conf.txt](https://github.com/riptideio/pymodbus/files/1531316/conf.txt)

rahulraghu94:
Hello @MazoSergey,

Could you please try [this branch](https://github.com/riptideio/pymodbus/tree/%23250-fix-scraper-simulator) on your config file?
I contains a fix that worked. Please let us know :)

MazoSergey:
Hello @rahulraghu94,
Thank you very much guys for your support. I tried to use an updated modbus-simulator. Now it is started, but it seems that simulator does not take correct date from map dump. I've make a dump of PLC with 400001 = 111, 400002 = 222, but in simulator i see only zeros :(

rahulraghu94:
Hello @MazoSergey , can you provide code excerpts, output and logs appended to the problem you're facing?

MazoSergey:
Again thanks guys! Tomorrow i will be on the work and will provide all needed information...

MazoSergey:
[...]

PR #259, opened 2018-01-08
#250 fix scraper simulator

rahulraghu94:
Works as expected:

```
(modbus) [~/pymodbus/examples/contrib]$ ./message-parser.py -b -t -p tcp -m "0x0 0xc 0x0 0x0 0x0 0x13 0x1 0x3 0x10 0x0 0xa 0x0 0xa 0x0 0xa 0x0 0xa 0x0 0xa 0x0 0xa 0x0 0xa 0x0 0xa"                            *[#250-fix-scraper-simulator]
================================================================================
Decoding Message 000c00000013010310000a000a000a000a000a000a000a000a
================================================================================
ServerDecoder
--------------------------------------------------------------------------------
ClientDecoder
--------------------------------------------------------------------------------
name            = ReadHoldingRegistersResponse
protocol_id     = 0x0
unit_id         = 0x1
skip_encode     = 0x0
registers       =
                . [10, 10, 10, 10, 10, 10, 10, 10]
check           = 0x0
transaction_id  = 0xc
documentation   =
    This function code is used to read the contents of a contiguous block
    of holding registers in a remote device. The Request PDU specifies the
    starting register address and the number of registers. In the PDU
    Registers are addressed starting at zero. Therefore registers numbered
    1-16 are addressed as 0-15.
```

screen shot 2018-01-08 at 4 00 13 pm


examples/common/synchronous-client.py
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env python
```
dhoomakethu: There is alreadty a file by name `synchronous_client.py` , Remove this from the commit.

rahulraghu94:
@dhoomakethu , not sure how [this commit](https://github.com/riptideio/pymodbus/pull/259/commits/968475c5d4300ce798d0aaded4938449107c2cd7) made into dev branch already. This PR will then consist of only new command line option to the message parser along with some minor improvement to error handling.

PR #268, opened 2018-01-23
Fix AsyncioModbusSerialClient (serial options and create_serial_connection usage)

cbergmiller:
The API of serial_asyncio.create_serial_connection is different from the async socket API.
Also, the serial kwargs (baudrate etc.) were missing in AsyncioModbusSerialClient.__init__ Method.

### DIFF
--- a/pymodbus/client/async/asyncio/__init__.py
+++ b/pymodbus/client/async/asyncio/__init__.py
@@ -659,7 +659,8 @@ class AsyncioModbusSerialClient(object):
     transport = None
     framer = None
 
-    def __init__(self, port, protocol_class=None, framer=None,  loop=None):
+    def __init__(self, port, protocol_class=None, framer=None,  loop=None,
+                 baudrate=9600, bytesize=8, parity='E', stopbits=1):
         """
         Initializes Asyncio Modbus Serial Client
         :param port: Port to connect
@@ -674,6 +675,10 @@ class AsyncioModbusSerialClient(object):
         #: Event loop to use.
         self.loop = loop or asyncio.get_event_loop()
         self.port = port
+        self.baudrate = baudrate
+        self.bytesize = bytesize
+        self.parity = parity
+        self.stopbits = stopbits
         self.framer = framer
         self._connected = False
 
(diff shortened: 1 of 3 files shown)


pymodbus/client/async/asyncio/__init__.py
```diff
@@ -707,11 +700,18 @@ def connect(self):
         """
         _logger.debug('Connecting.')
         try:
-            yield from self.loop.create_connection(self._create_protocol)
-            _logger.info('Connected to %s:%s.' % (self.host, self.port))
+            from serial_asyncio import create_serial_connection
```
dhoomakethu: Please move this change to serial factory, https://github.com/cbergmiller/pymodbus/blob/fix/serial_async_client/pymodbus/client/async/factory/serial.py#L82.

cbergmiller: Thanks for the review. I updated the pull request.

dhoomakethu:
@cbergmiller the tests are failing on python3 , please have a look.

pymodbus/client/async/asyncio/__init__.py
```diff
@@ -680,18 +680,27 @@ def __init__(self, port, protocol_class=None, framer=None,  loop=None,
...
-        self._connected = False
+        self._connected_event = asyncio.Event()
 
     def stop(self):
         """
         Stops connection
         :return:
         """
-        if self._connected:
+        if self._connected.is_set():
```
dhoomakethu: You can directly use the property `self._connected` here.

pymodbus/client/async/asyncio/__init__.py
```diff
@@ -718,8 +725,8 @@ def protocol_made_connection(self, protocol):
         Protocol notification of successful connection.
         """
         _logger.info('Protocol made connection.')
-        if not self._connected:
-            self._connected = True
+        if not self._connected_event.is_set():
```
dhoomakethu: You can directly use the property `_connected` here.

cbergmiller: I didn't want to break compatibility with possible subclasses or mixins. So i created the `_connected` property that returns the state of the connected event. But you are right it is cleaner to use the `_property` inside the class as well. I will update the pull request.

PR #280, opened 2018-03-14
Update concurrent_client.py - Python3 compatbility

dragoshenron:
iter.next() was removed in python 3 so the code is broken on line 218.
This edit brings compatibility to python 3.
Successfully tested with Pythin 2.7.13 and 3.5.3


dhoomakethu:
Thanks @dragoshenron , I will merge this as part of 1.5.0 release

PR #285, opened 2018-04-09
Pymodbus 1.5.0

dhoomakethu:
* Improve transaction speeds for sync clients (RTU/ASCII), now retry on empty happens only when retry_on_empty kwarg is passed to client during intialization

`client = Client(..., retry_on_empty=True)`

* Fix tcp servers (sync/async) not processing requests with transaction id > 255
* Introduce new api to check if the received response is an error or not (response.isError())
* Move timing logic to framers so that irrespective of client, correct timing logics are followed.
* Move framers from transaction.py to respective modules
* Fix modbus payload builder and decoder
* Async servers can now have an option to defer `reactor.run()` when using `Start<Tcp/Serial/Udo>Server(...,defer_reactor_run=True)`
* Fix UDP client issue while handling MEI messages (ReadDeviceInformationRequest) 
* Fix Misc examples


dhoomakethu:
@karlp @kierun @mrcreatd @memetb @bosnj . Please give a try/share your comments.

dhoomakethu:
I will wait for another week for others to share there review comments. The release will be on 04/23/2018 unless some serious issues come up.

kierun:
eek… Been pulled to do something else while the code I wrote is in user acceptance test which could last a few months. Due to horrid time pressure, I have no time to test the new release. I do apologies as it's a shitty thing to do to you when you have been nothing but kind.

Issue #47, opened 2014-05-16
Handling of malformed MEI responses


Tamarinen:
We have a heat pump that manages to confuse pymodbus when asked to identify itself through RTU. When asked for the mandatory information, the response is as follows:

0x01 00x2b 0x0e 0x01 0x01 0x00 0x00 0x01 0x00 0x04 NIBE 0x01 0x08 F1345-60 0x02 0x04 4150 0xab 0x9a

The pump indicate that the MEI response contain only one object, while three are sent, which calculateRtuFrameSize() in mei_message.py doesn't like at all.

I propose this method of verifying the MEI response structure:

``` python
@classmethod
def calculateRtuFrameSize(cls, buffer):
    ''' Calculates the size of the message

    :param buffer: A buffer containing the data that have been received.
    :returns: The number of bytes in the response.
    '''
    size = 8 # skip the header information

    while size + 2 < len(buffer):
        _, object_size = struct.unpack('>BB', buffer[size:size+2])
        size += object_size + 2

    return size + 2
```

Instead of trying to read the specified number of objects, this read objects until the input buffer is exhausted, returning a valid length for a certain class of malformed MEI responses. The same method of parsing the response is used in ReadDeviceInformationResponse.decode().

PR #293, opened 2018-04-26
Add support for MEI device identification encoding respecting PDU size and repeated Object IDs in the device identification

chintal:
This PR allows parsing of some kinds of non-standard MEI device information responses. In particular, if a device (modbus server) reply to an MEI device information request (0x2B 0x0E) includes multiple objects with the same ID. I am uncertain about whether any current devices do this, though given the fact that I have had good reason to add such behavior in my devices and the likes of #47, I would not be surprised if there are indeed a few which do so. 

This change included in this PR will cause no change in the behavior of any existing code with standard devices, and continues to return a dict with the object IDs as keys. However, in case an object ID is encountered twice, the value in the dictionary against that object ID (and only that object ID) is turned into a list. The objects in the list appear in the order in which they are encountered in the response. The performance cost to this approach should be minimal to the point of being negligible.

As an example of where I use such a mechanism, I dedicate one object ID to storing version numbers of included libraries in a semi-descriptive string. As multiple libraries are linked into the firmware, each adds a device information object with it's version string. Another example where I intend to use such behavior is to store per-channel calibration information in a client-readable format, in instances where the number of channels is not known before hand. 

I do currently have a functioning workaround, which subclasses ReadDeviceInformationResponse and makes the necessary changes. However, changing the Response parser and maintaining it separately does take a little more effort than I am quite comfortable with, given that ModbusSerialClient needs to subclassed as well and ClientDecoder needs to be replicated entirely.

### DIFF
--- a/pymodbus/mei_message.py
+++ b/pymodbus/mei_message.py
@@ -155,7 +155,14 @@ class ReadDeviceInformationResponse(ModbusResponse):
         while count < len(data):
             object_id, object_length = struct.unpack('>BB', data[count:count+2])
             count += object_length + 2
-            self.information[object_id] = data[count-object_length:count]
+            if object_id not in self.information.keys():
+                self.information[object_id] = data[count-object_length:count]
+            else:
+                if isinstance(self.information[object_id], list):
+                    self.information[object_id].append(data[count-object_length:count])
+                else:
+                    self.information[object_id] = [self.information[object_id],
+                                                   data[count - object_length:count]]
 
     def __str__(self):
         ''' Builds a representation of the response
--- a/pymodbus/mei_message.py
+++ b/pymodbus/mei_message.py
@@ -114,18 +114,14 @@ class ReadDeviceInformationResponse(ModbusResponse):
         ModbusResponse.__init__(self, **kwargs)
         self.read_code = read_code or DeviceInformation.Basic
         self.information = information or {}
-        self.number_of_objects = len(self.information)
-        for obj in self.information:
-            if isinstance(obj, list):
-                self.number_of_objects += (len(obj) - 1)
+        self.number_of_objects = 0
         self.conformity = 0x83 # I support everything right now
 
         # TODO calculate
         self.next_object_id = 0x00 # self.information[-1](0)
         self.more_follows = MoreData.Nothing
 
-    @staticmethod
-    def _encode_object(object_id, data):
+    def _encode_object(self, object_id, data):
         encoded_obj = struct.pack('>BB', object_id, len(data))
         if IS_PYTHON3:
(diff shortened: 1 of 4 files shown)


dhoomakethu:
@chintal would you mind adding an example (both for server/client) to  showcase this change ? Also please update unit tests for the same.

chintal:
@dhoomakethu The change is currently only on the client side. The change on the server side would be relatively more intrusive, but I shall make it anywajy and perhaps back out the changes later on if there are concerns.

I will look at the unit tests and examples. I think the unit tests should be manageable, but I'm not entirely certain how / where the examples would go. Any example would presumably need both a server and a client to showcase the change? Do you have recommendations of a current example or pair of examples which I can use as a starting point for one?

As as aside, now that I look at the server code for MEI, it looks like it would not actually be compatible with ModbusRTU, specifically due to the RTU frame size limit of 256 bytes. Is this on purpose?

chintal:
@dhoomakethu I have added the example you requested. I have also put in the superstructure to handle the case of splitting the device information across multiple responses, but see below.

pymodbus/mei_message.py
```diff
@@ -114,31 +119,55 @@ def __init__(self, read_code=None, information=None, **kwargs):
...
-            self.next_object_id, self.number_of_objects)
-
-        for (object_id, data) in iteritems(self.information):
-            packet += struct.pack('>BB', object_id, len(data))
-            if IS_PYTHON3:
-                if isinstance(data, bytes):
-                    packet += data
+        packet = struct.pack('>BBB', self.sub_function_code,
+                             self.read_code, self.conformity)
+        # TODO By the time the code gets here, self.space_left should be set
```
chintal: This thing needs to be done before split responses actually work.

pymodbus/mei_message.py
```diff
@@ -114,31 +119,55 @@ def __init__(self, read_code=None, information=None, **kwargs):
...
+        # to the length of the largest allowable PDU for the transport and
+        # framer, ie, 252 for RTU.
```
chintal: This thing needs to be done before split responses will actually work.

chintal:
@dhoomakethu This PR is done. It should fix the ReadDeviceInformationResponse for Servers and includes an example for how a client should handle it. As far I can tell if should be ready to merge and should cause no change in behavior for existing user code. 

I'm still not especially comfortable with adding a separate 'request' for the client which wraps multiple individual modbus transactions. This might create a number of other special cases which will be annoying to handle and to maintain. If you have a preexisting structure which already wraps multiple transactions within the codebase, you may use that to create another such instance if you desire it.

Issue #254, opened 2017-12-23
Prepare for 2.0.0


dhoomakethu:
Following issues would be covered and will be updated on need basis
This would cover as an Epic to cover all open tasks pending with PyModbus for 2.0.0 release. Please do not close this issue . All new Issues to be mentioned/referenced here.

- [x] Check #37
- [x] Check #47 
~~- [ ] Check #93~~ --> Moved to next release
- [x] Check #138
- [x] Fix #160 
- [x] Fix #212 
- [ ] Fix #220 
- [x] Fix #221 
~~- [ ] Fix #232~~ --> Moved to next release
- [x] Fix #250 

~~- [ ] Check and Merge PR #158~~ --> Moved to next release
- [x] Check and Merge PR #188
- [x] Check and Merge PR #246 
~~- [ ] Check and Merge PR #321~~  --> Moved to next release
- [x] Check and Merge PR #323 

- [x] Add Deprecating warning for old type class instantiation for async clients

PR #325, opened 2018-09-07
#254 backward comaptibility

dhoomakethu:
Ensure async client backward compatibility with older versions of pymodbus.

### DIFF
--- a/pymodbus/client/async/deprecated/__init__.py
+++ b/pymodbus/client/async/deprecated/__init__.py
@@ -1,5 +1,10 @@
-"""
-Async Modbus Client implementation based on Twisted, tornado and asyncio
+import warnings
+warnings.simplefilter('always', DeprecationWarning)
+
+WARNING = """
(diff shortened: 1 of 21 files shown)


pymodbus/client/async/deprecated/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+
+    from pymodbus.client.async.tcp import AsyncModbusTCPClient as Client
+    from pymodbus.client.async.serial import AsyncModbusSerialClient as Client
+    from pymodbus.client.async.udp import AsyncModbusUDPClient as Client
+
+    # For tornado based async client use
+    event_loop, future = Client(schedulers.IO_LOOP, port=5020)
+
+    # For twisted based async client use
+    event_loop, future = Client(schedulers.REACTOR, port=5020)
```
bhagn: `event_loop, deferred = Client(scheduler.REACTOR, port=5020)` - just to be consistent with the terminology of the framework being used.

PR #329, opened 2018-09-17
Pymodbus repl

dhoomakethu:
https://github.com/riptideio/pymodbus/blob/pymodbus-repl/pymodbus/repl/README.md


.readthedocs.yml
```diff
@@ -6,5 +6,6 @@ requirements_file: requirements-docs.txt
 python:
     extra_requirements:
         - twisted
+        - torndo
```
tracernz: Typo

dhoomakethu: Thanks for the catch.

PR #334, opened 2018-09-22
Async updates

dices:
Updates to support python 3.7. They were tested in multiple versions of python and they worked. Hoping they pass your CI tests there.


dices:
Closing this pull request because left out a commit

PR #335, opened 2018-09-22
Async updates

dices:
Updates to support python 3.7
Tested using tox


dices:
There seems to be an issue with the 3.7 environment with downloading python3.7 tar

dices:
Closing to remove unwanted commit

PR #339, opened 2018-09-26
Avoid unnecessary import of deprecated module #338

### LINKED ISSUE #338: `client.async` sneaking in dependencies
In importing `client.async.asyncio`, `client.async.__init__` triggers `from pymodbus.client.async.deprecated.async` ([source](https://github.com/riptideio/pymodbus/blob/master/pymodbus/client/async/__init__.py)), which then imports twisted. Twisted is a fairly heavy dependency and is not needed to use asyncio.

dhoomakethu:
Fix #338


dhoomakethu:
cc @patrickfuller

Issue #340, opened 2018-09-26
Exception thrown by BinaryPayloadDecoder


dzimine:
### Versions

* Python: 2.7
* OS: Ubuntu 16.04 + AWS Greengrass
* Pymodbus: 2.0.0
* Modbus Hardware (if used): No

### Pymodbus Specific
* Server: tcp - sync/async
* Client: tcp - sync/async

### Description

Parsing payload sometimes throws an exception: 
```
Error: 'ascii' codec can't encode character u'\xbb' in position 87: ordinal not in range(128)
```
The same code works with exactly the same input data on OS-X and Ubuntu 16.04 but strangely fails when run as Lambda on AWS Greengrass core (deployed on the same Ubuntu 16.04 box). 

The code line causing the exception is https://github.com/riptideio/pymodbus/blob/master/pymodbus/payload.py#L344: 

```
        _logger.debug(unicode_string(handle))
```
Commenting it out or removing `unicode_string` conversation fixes the problem. 

### Code and Logs

```
[2018-09-25T13:26:09.47Z][DEBUG]-transaction.py:112,Current transaction state - TRANSACTION_COMPLETE
[2018-09-25T13:26:09.472Z][DEBUG]-transaction.py:116,Running transaction 24
[2018-09-25T13:26:09.474Z][DEBUG]-transaction.py:206,SEND: 0x0 0x18 0x0 0x0 0x0 0x6 0x1 0x3 0x0 0x0 0x0 0x8
[2018-09-25T13:26:09.475Z][DEBUG]-sync.py:75,New Transaction state 'SENDING'
[2018-09-25T13:26:09.477Z][DEBUG]-transaction.py:209,Changing transaction state from 'SENDING' to 'WAITING FOR REPLY'
[2018-09-25T13:26:09.479Z][DEBUG]-transaction.py:284,Changing transaction state from 'WAITING FOR REPLY' to 'PROCESSING REPLY'
[2018-09-25T13:26:09.48Z][DEBUG]-transaction.py:214,RECV: 0x0 0x18 0x0 0x0 0x0 0x13 0x1 0x3 0x10 0x18 0x3f 0xbb 0x48 0x0 0xd 0x0 0xd 0x0 0xd 0x0 0xd 0x0 0xd 0x0 0xd
[...]

PR #341, opened 2018-09-26
Fix exception thrown by BinaryPayloadDecoder

dzimine:
This fixes bug #340. 

The problem is difficult to reproduce yet irritating, I hope loosing a tiny bit of debug info worth the fix.


dhoomakethu:
@dzimine  Thanks for the PR, the unicode error is too annoying to fix.

PR #345, opened 2018-09-28
Python 3.7 async keyword

dices:
Renamed pymodbus.server.async to pymodbus.server.asynchronous
Modified asyncio.async to use asyncio.ensure_future
Added missing prompt_toolkit dependency to requirements-docs.txt
Updated references in documentation
Updated examples


dices:
I am the suck

PR #350, opened 2018-10-08
Fixing endianness in Payload example.

EricDuminil:
builder and decoder don't have the same byteorder in this example:
https://pymodbus.readthedocs.io/en/latest/source/example/modbus_payload.html

`64float` comes out as `1.0489227558190733e+146` instead of `123.45`

My commits set both byteorder to `Endian.Big`. and `64float` is decoded to `123.45` again.


dhoomakethu:
@EricDuminil  Thanks for the PR.

PR #354, opened 2018-10-16
check self.socket

mpf82:
self.socket might be None at this point


pymodbus/client/sync.py
```diff
@@ -473,7 +473,8 @@ def connect(self):
             _logger.error(msg)
             self.close()
         if self.method == "rtu":
-            self.socket.interCharTimeout = self.inter_char_timeout
+            if(self.socket):
```
tracernz: No parens needed, canonical style would be `if socket is not None:`

dhoomakethu: Agree, please remove the parenthesis.
```suggestion
            if self.socket:
```

Issue #368, opened 2018-12-18
RTU broadcast doesn't function as expected using client.write_registers(add, val, unit=0)


muhlbaier:
### Versions

* Python: 2.7.15
* OS: Win 7
* Pymodbus: 2.1.0
* Modbus Hardware (if used): pymodbus on both sides

### Pymodbus Specific
* Server: rtu - sync
* Client: rtu - sync

### Description

When there are multiple slaves on a pymodbus server I would expect them all to receive a broadcast message but none receive the message. There doesn't seem to be any special handling for the broadcast address. If I were to create a slave context at address 0 it alone would respond to the broadcast address. The below example creates slave 1 and 2 and writes to slave 1 then slave 2 then writes to 0 to broadcast to both of them but this doesn't work as expected.

### Code and Logs

```python
import threading
from pymodbus.client.sync import ModbusSerialClient as ModbusClient
from pymodbus.datastore import ModbusSequentialDataBlock
from pymodbus.datastore import ModbusSlaveContext, ModbusServerContext
from pymodbus.server.sync import StartSerialServer
from pymodbus.device import ModbusDeviceIdentification
from pymodbus.transaction import ModbusRtuFramer


# custom datablock to print what is going on
class MyDataBlock(ModbusSequentialDataBlock):
    def __init__(self, address, values, slave_id):
        self.slave_id = slave_id
        super(MyDataBlock, self).__init__(address, values)

    def setValues(self, address, values):
        print "slave " + str(self.slave_id) + " write address: " + str(address) + " values: " + str(values)
        super(MyDataBlock, self).setValues(address, values)


# create two slaves with unit ids 1 and 2
store = dict()
store[1] = ModbusSlaveContext(di=None, co=None, hr=MyDataBlock(0, [0] * 100, 1), ir=None)
store[2] = ModbusSlaveContext(di=None, co=None, hr=MyDataBlock(0, [0] * 100, 2), ir=None)

context = ModbusServerContext(slaves=store, single=False)
# initialize the server information
identity = ModbusDeviceIdentification()
identity.VendorName = 'Pymodbus'
identity.ProductCode = 'PM'
identity.VendorUrl = 'http://github.com/riptideio/pymodbus/'
identity.ProductName = 'Pymodbus Server'
identity.ModelName = 'Pymodbus Server'
identity.MajorMinorRevision = '1.0'


def server_thread():
[...]

PR #372, opened 2019-01-10
#368 Fixes write to broadcast address

muhlbaier:
When writing to broadcast address (unit_id=0) there should be no response according to the Modbus spec. This fix changes expected_response_length to 0 when writing to unit_id=0. This will break any existing code that is improperly using unit_id 0 for a slave address.


dhoomakethu:
@muhlbaier Thanks for the PR, there are couple of things I would like to be fixed along with this PR.

The changes enables broadcast address for all requests (including read requests) but as per the specs the broadcast messages are available only for write requests.

From the [modbus specs](http://www.modbus.org/docs/Modbus_over_serial_line_V1_02.pdf)
 
> In broadcast mode, the master can send a request to all slaves.
> No response is returned to broadcast requests sent by the master. The broadcast requests are necessarily writing commands. All
> devices must accept the broadcast for writing function. The address 0 is reserved to identify a broadcast exchange. 

This also requires changes on the pymodbus server, Ideally I would like to have the changes for both the client and server go together. 

Let me know, if you can do these changes , otherwise I will keep this on hold and will work on the comments when I get some time.

muhlbaier:
@dhoomakethu I should be able to work on these features in the next week or two.

muhlbaier:
@dhoomakethu I added broadcast support and a broadcast_enable kwarg so it will work as it has in the past unless users use broadcast_enable=True on the client and or server. My changes should work for sync/async client and sync server. I wasn't sure the best way to add support for async server.

The below code is what I used to test both the client and server. My testing exposed one other weakness on the sync serial server which is it always waits for a timeout when receiving requests. Normally this issue doesn't cause problems except for slowing things down since the client waits for the response which comes after the timeout on the server read. With broadcast_enable=True the client doesn't wait for a response so it might send another request while the server is still waiting for the timeout. Thus in the following code I had to add a sleep equal to the server timeout otherwise the following transactions fail. Let me know your thoughts on this issue.

`
from __future__ import print_function
import threading
import time
from pymodbus.client.sync import ModbusSerialClient as ModbusClient
from pymodbus.datastore import ModbusSequentialDataBlock
from pymodbus.server.sync import StartSerialServer
from pymodbus.device import ModbusDeviceIdentification
from pymodbus.transaction import ModbusRtuFramer

class MyDataBlock(ModbusSequentialDataBlock):
    def __init__(self, address, values, slave_id):
        self.slave_id = slave_id
        super(MyDataBlock, self).__init__(address, values)

    def setValues(self, address, values):
        print("slave " + str(self.slave_id) + " write address: " + str(address) + " values: " + str(values))
        super(MyDataBlock, self).setValues(address, values)


store = dict()
store[1] = ModbusSlaveContext(di=None, co=None, hr=MyDataBlock(0, [0] * 100, 1), ir=None)
store[2] = ModbusSlaveContext(di=None, co=None, hr=MyDataBlock(0, [0] * 100, 2), ir=None)

context = ModbusServerContext(slaves=store, single=False)
# initialize the server information
identity = ModbusDeviceIdentification()
identity.VendorName = 'Pymodbus'
identity.ProductCode = 'PM'
identity.VendorUrl = 'http://github.com/riptideio/pymodbus/'
identity.ProductName = 'Pymodbus Server'
identity.ModelName = 'Pymodbus Server'
identity.MajorMinorRevision = '1.0'


def server_thread():
    StartSerialServer(context, framer=ModbusRtuFramer, identity=identity, port="COM83", timeout=0.1, baudrate=115200,
                      broadcast_enable=True)


server = threading.Thread(target=server_thread)
server.daemon = True
server.start()

client = ModbusClient(method='rtu', port="COM82", baudrate=115200, timeout=2, broadcast_enable=True)
client.connect()
# write to slave 1
registers = [2, 3, 4, 5]
request_result = client.write_registers(1, registers, unit=1)
print(request_result)
# write to slave 2
registers = [6, 7, 8, 9]
request_result = client.write_registers(1, registers, unit=2)
print(request_result)
# broadcast to both
registers = [1, 1, 1, 1]
request_result = client.write_registers(1, registers, unit=0)
print(request_result)
# must delay for however long the server timeout is otherwise the following requests get missed.
# todo fix server so it can process requests as they are received instead of waiting for the timeout every time
time.sleep(0.1)
# read from slave 1
request_result = client.read_holding_registers(1, 4, unit=1)
if "Error" not in str(request_result):
    print("slave 1 read values", request_result.registers)
# read from slave 2
request_result = client.read_holding_registers(1, 4, unit=2)
if "Error" not in str(request_result):
    print("slave 2 read values", request_result.registers)
`

dhoomakethu:
I have included this as part of [pymodbus-2.2.0](https://github.com/riptideio/pymodbus/commits/pymodbus-2.2.0) branch. I will be closing this PR .

Thanks.

PR #376, opened 2019-01-16
Fixed erronous CRC handling

JStrbg:
If the CRC recieved is not correct in my case my slave got caught in a deadlock, not taking any new requests. This addition fixed that.

### DIFF
--- a/pymodbus/framer/rtu_framer.py
+++ b/pymodbus/framer/rtu_framer.py
@@ -91,7 +91,12 @@ class ModbusRtuFramer(ModbusFramer):
             data = self._buffer[:frame_size - 2]
             crc = self._buffer[frame_size - 2:frame_size]
             crc_val = (byte2int(crc[0]) << 8) + byte2int(crc[1])
-            return checkCRC(data, crc_val)
+            if checkCRC(data, crc_val):
+                return True
+            else:
+                _logger.debug("CRC invalid, discarding header!!")
+                self.resetFrame()
+                return False
         except (IndexError, KeyError, struct.error):
             return False
 


pymodbus/framer/rtu_framer.py
```diff
@@ -91,7 +91,12 @@ def checkFrame(self):
             data = self._buffer[:frame_size - 2]
             crc = self._buffer[frame_size - 2:frame_size]
             crc_val = (byte2int(crc[0]) << 8) + byte2int(crc[1])
-            return checkCRC(data, crc_val)
+            if checkCRC(data, crc_val):
```
dhoomakethu: Ideally I would prefer the handling of failure in `processIncomingPacket` method. If you can move the `else` clause there, I would be happy to accept this change.

```
diff --git a/pymodbus/framer/rtu_framer.py b/pymodbus/framer/rtu_framer.py
index 8579a57..16af459 100644
--- a/pymodbus/framer/rtu_framer.py
+++ b/pymodbus/framer/rtu_framer.py
@@ -222,6 +222,9 @@ class ModbusRtuFramer(ModbusFramer):
                     _logger.debug("Not a valid unit id - {}, "
                                   "ignoring!!".format(self._header['uid']))
                     self.resetFrame()
+            else:
+                _logger.debug("CRC invalid, discarding header!!")
+                self.resetFrame()
         else:
             _logger.debug("Frame - [{}] not ready".format(data))
```

JStrbg: Not able to do that without changing the return type of `checkFrame()` which makes it kinda clunky, since we don't want to `resetFrame()` on every _False_ returned from `checkFrame()`. Help a noob out if you think there is another way, I don't see it :)

dhoomakethu: I believe we can use `resetFrame` if `checkFrame` fails, may be we can keep the debug messages generic, something like `Check frame failed ...` in `processIncomingPacket` method or  can be more explicit by logging appropriate messages inside `checkFrame` method.
Any how just for your reference.

From the [specs](http://www.modbus.org/docs/Modbus_over_serial_line_V1_02.pdf)
for Master

> When a reply is received, the Master checks the reply before starting the data processing. The checking may result in an error,
> for example a reply from an unexpected slave, or an error in the received frame. In case of a reply received from an unexpected
> slave, the Response time-out is kept running. In case of an error detected on the frame, a retry may be performed. 

For Slave:

> If the slave detects an error in the received frame, no respond is returned to the master.

JStrbg: I just tried it and the result on my system is that if the socket `timeout` is set to 0, and the `else` clause is in `processIncomingPacket` instead of `checkFrame`, my system spits out _CRC invalid_ close to every second request. If `timeout` is above 0.01 it happens less frequently, and above 0.2it seem to happen only when i purposely send it an invalid CRC. As opposed to when the `else` clause is located in `checkFrame` the `timeout` setting has no effect on the amount of _CRC invalid_ reported, it stays as when the timeout is set to above 0.2, only reporting actual invalid CRC's.

My theory is that the library relies on `checkFrame` returning _False_ whenever a partial message is recieved, of which we do _not_ want to reset completely as the remaining part is about to be read. Atleast this is how my setup seem to behave.

dhoomakethu:
@JStrbg sorry for the delay and thanks for the additional tests and explanation. I will try to move this PR to a closure asap after some tests at my end.

dhoomakethu:
I have[ cherry-picked ](https://github.com/riptideio/pymodbus/commit/5030514e851ef910632d6153de50570986c70d7c) this as part [pymodbus 2.2.0](https://github.com/riptideio/pymodbus/commits/pymodbus-2.2.0) release branch . I will be closing this PR.

Thanks

PR #400, opened 2019-03-24
asyncio server implementation

memetb:
This is a preliminary server implementation using core `asyncio` functionality.

I have successfully run the [examples/common/synchronous_client.py](https://github.com/memetb/pymodbus/blob/pymodbus-2.2.0/examples/common/synchronous_client.py) tests for TCP and UDP. I have not integrated other tests as of this commit.

Topics to be discussed:

- default asyncio event loop is used throughout the implementation, this may or may not be flexible enough depending on peoples' requirements. Open to revision.
- the serial implementation hasn't been done yet. As per your recommendation, I assume [pyserial-asyncio](https://github.com/pyserial/pyserial-asyncio) will be necessary to integrate it. It should not be very difficult to do so.
- There were some minor differences between the `handle` implementation for TCP and UDP. I'm not sure why those were and I've left a commented line [here](https://github.com/memetb/pymodbus/blob/pymodbus-2.2.0/pymodbus/server/asyncio.py#L128). Please comment.
- exception handling of the handle loop hasn't been looked at properly. I just copied over from existing code and made guesses. I probably need some guidance from unit tests.
- `unittest`: will require something like `asynctest` or the equivalent. I can see there are some helper tests but haven't fully investigated. I'm willing to contribute some more but I need some guidance and I need to work on my main project at the same time. Especially useful would be if there are tests I can wholesale copy over.

... any other comments?

(it goes without saying that this is not merge ready as is)

Thanks.


dhoomakethu:
@memetb Thanks for the PR, Could you please fix the failing tests ?

memetb:
Sure. At least one failures is that I have 0% code coverage. Like I said, I will tackle this time permitting this week but I would appreciate if you point me to existing tests that i could reuse.

Thanks.

dhoomakethu:
You can take a look in to existing [async_server tests](https://github.com/riptideio/pymodbus/blob/master/test/test_server_async.py)

memetb:
@dhoomakethu  sorry this has taken so long to advance. Do you have a problem with adding an `asynctest` dependency to your tests? This is the package I use for making asyncio native unittests.

memetb:
@dhoomakethu  I have added the unittests for the asycnio server implementation.

Known issues:
- Most obviously, this only works for python3.7+. Can you please help me out with the configuration so that the tests are properly skipped on relevant platforms? (I actually only see environments up to 3.6 in travis)
- as mentioned above, I have added `asyncttest` dependency
- The Serial server is not implemented. I do not have access to a test bed and also don't have time right now to do it. Someone else can contribute or I can return to it as a separate issue at a later date.
- I'm not sure why but I wasn't able to send garbage data to the framer to trigger an exception. The two tests `testTcpServerGarbage` and `testUdpServerGarbage` are commented out for this reason.
- Code coverage is 94%, mainly due to the `execute` method. This is in part related to the above issue. This method was lifted from the asynchronous implementation so it shouldn't be problematic.
- StopServer is deprecated since multiple servers can be run simultaneously in asyncio, and therefore stopping semantics are now instance based
- ModbusServerFactory is deprecated as it was an abstraction leakage from the twisted implementation. The equivalent of `twisted.internet.protocol.ServerFactory` doesn't exist in `asyncio`.


Minor point: commit [50c5117](https://github.com/riptideio/pymodbus/pull/400/commits/50c511791c3601cc5eb2f8fb081f01db10f5d656) was a minor detail to cleanup code base, as the project was failing to load on an install without serial.

dhoomakethu:
@memetb Thanks for the updates, I will take a look and will create a separate issue for serial server.

memetb:
No problem. I've just added two test cases that bumps up coverage to 98%.

I'll check back regularly to see if you have any further requests.

dhoomakethu:
@memetb pymodbus-2.2.0 is merged in dev, please re-raise your PR against dev branch. Sorry for the confusion.

memetb:
Does this work?

tracernz:
Pleased to see this work progressing! I think you need to add asynctest to requirements-tests.txt to fix the CI errors.

memetb:
Thanks @tracernz, I've added asynctest into the requirements. However, this code is only meant to work for 3.7+. The highest python environment I'm seeing is 3.6.

memetb:
I've added the asynctest to the requirements file and also skipped tests for outright unsupported targets (<3.5), however, on targets that don't support it (e.g. python2), the loading of the unittest fails because of the `async def` aren't recognize keywords yet.

Furthermore: on higher targets, there is also a version mismatch between earlier versions of asyncio and the current 3.7 which I have coded and tested against.

In the [travis config file](https://github.com/riptideio/pymodbus/blob/master/.travis.yml#L14), the 3.7 target is specifically commented out which I guess is because the overall library isn't ready for it.

Let me know what you guys think. I'm personally not enormously upset by lack of backward compatibility since asyncio is a new feature and pymodbus has worked without it before this. However, if the project requirement for this feature be that it's as far backward compatible as possible, I'm going to have to go back to the drawing board and this may take a few more weeks.

memetb:
Ok, I think a more extensive refactoring is required. I just pushed a commit removing 3.7 language keywords from the unittests and now there's issues with f-strings.

Still open to suggestions.

dhoomakethu:
@memetb I am yet to go through the code, but at present we skip the unittests for python<3.4 with this fixture `@pytest.mark.skipif(not IS_PYTHON3, reason="requires python3.4 or above")` . Refer [test_client_async_asyncio.py](https://github.com/riptideio/pymodbus/blob/master/test/test_client_async_asyncio.py#L19). I will give it a spin when I get some freecycles and share the feedback here.

memetb:
Thanks @dhoomakethu. Commit 985d0cf added just that fixture.

As an aside, I've been looking at some other projects (e.g. influxdb python client) and I might get some ideas. Also, pending my surplus time from work, I've come to think it's probably better for me to update the implementation itself so that it works on earlier asyncio versions given that this is a library and not an end-user application.

Comments are welcome, as always.

memetb:
Any updates on this PR?

dhoomakethu:
@memetb I am waiting for the tests to be fixed.

memetb:
@dhoomakethu I'll be going on leave for a short bit so won't be working on anything before September. 

However, please check my comments above: code coverage is upper 90's and there are unit tests for each feature. What is failing to build is that the project itself [does not have python 3.7 support](https://github.com/riptideio/pymodbus/blob/master/.travis.yml#L14) and this code is not meant for (or even useful for) any language level below that.

dhoomakethu:
@memetb thanks for the clarification, I will merge this and fix the tests later.

dhoomakethu:
@memetb did you close this by Mistake?

memetb:
@dhoomakethu: yes, sorry about that.

PR #411, opened 2019-04-22
Remove unused pycrypto dep

### LINKED ISSUE #402: Use pycryptodome instead of pycrypto.
### Versions

* Python: 2.7.12
* OS: Ubuntu 18.04
* Pymodbus: 2.1.0 [twisted]
* Modbus Hardware (if used): 

### Pymodbus Specific
* Server: tcp - async

### Description

I am trying to use Mod bus server on TCP protocol, but when I installed pymodbus and I saw it's installed pycrypto, which is deprecated and dead software. 

I already have installed pycryptodome in my application, which is a conflict with pycrypto, 
we can't have both pycrypto and pycryptodome at the same time,

Can we have pymodbus[twisted] release which can use pycryptodome instead of pycrypto?

tracernz:
It has not been needed by Twisted for a long time, and has been unmaintained for a long time.

Fixes #402.

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -6,7 +6,6 @@ mock >= 1.0.1
 pyserial-asyncio==0.4.0;python_version>="3.4"
 pep8>=1.7.0
 pyasn1>=0.2.3
-pycrypto>=2.6.1
 pyserial>=3.4
 pytest-cov>=2.5.1
 pytest>=3.5.0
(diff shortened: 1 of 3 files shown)


ssuvorov-fls:
@tracernz hello, do you know when it will be merged into master and available through pip?

tracernz:
I guess with 2.3.0 which might come next month: https://github.com/riptideio/pymodbus/issues/384#issuecomment-522398679

PR #424, opened 2019-07-25
Adding padding to the diagnostic decode

tcplomp:
When running the GetClearModbusPlusRequest to retrieve statistics from a PLC the modbus message is 115 bytes long.
After stripping teh Function Code the message is 113 bytes.
word_len was adjusted for the extra byte, but data wasn't.
The struct.unpack would fail since the buffer was 113 bytes, but the format was 114 bytes long.


dhoomakethu:
@tcplomp Thanks for the PR. Could you please raise the PR against the `dev` branch.

PR #431, opened 2019-08-08
Improve efficiency of TCP _recv()

nickb937:
Concatenating data to a string is a slow operation in Python because
strings are immutable and have to be re-allocated for each append.

Storing the data in a list then using join() is faster.  I measured a 1%
drop in CPU utilisation.

### DIFF
--- a/pymodbus/client/sync.py
+++ b/pymodbus/client/sync.py
@@ -260,18 +260,21 @@ class ModbusTcpClient(BaseModbusClient):
         else:
             recv_size = size
 
-        data = b''
+        data = []
+        data_length = 0
         time_ = time.time()
         end = time_ + timeout
         while recv_size > 0:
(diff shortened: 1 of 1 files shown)


dhoomakethu:
@nickb937 Thanks for the PR, the changes seems to have broken the tests. Would you mind fixing them, please ?

nickb937:
The test was broken before the change.  It's an infrastructure error rather than an artefact of this change.

tracernz:
The problem was fixed in https://github.com/riptideio/pymodbus/pull/440. If you rebase your branch on top of current dev it should go away.

dhoomakethu:
Have cherry picked the changes to `dev` branch via commit b97659d81a32119dc06ea4ed5f278d3e27139986. Will be closing this.

PR #444, opened 2019-09-13
Support multiple Python versions to fix test error from PR #400

starnight:
Here are the commits to fix the test error from PR #400
* server/asyncio.py: Create server with appropriate args and environment
* Create asyncio task with appropriate method and environment
* test: Make serve_forever() test only with Python 3.7+
* server/asyncio.py: Fix format string for older Python
* test: Make assert_called_once() test only with Python 3.6+

There is still one **SyntaxError** error from `yield from` in test with Python 2.7, but have no idea how to skip the test_server_asyncio.py for it.  The good new is Python 2.7 will be end of support at the end of this year.

### DIFF
--- a/pymodbus/client/sync.py
+++ b/pymodbus/client/sync.py
@@ -1,5 +1,6 @@
 import socket
 import select
+import serial
 import time
 import sys
 from functools import partial
--- a/pymodbus/server/asyncio.py
+++ b/pymodbus/server/asyncio.py
@@ -62,7 +62,10 @@ class ModbusBaseRequestHandler(asyncio.BaseProtocol):
             self.framer = self.server.framer(self.server.decoder, client=None)
 
             # schedule the connection handler on the event loop
-            self.handler_task = asyncio.create_task(self.handle())
+            if PYTHON_VERSION >= (3, 7):
+                self.handler_task = asyncio.create_task(self.handle())
+            else:
+                self.handler_task = asyncio.ensure_future(self.handle())
         except Exception as ex: # pragma: no cover
             _logger.debug("Datastore unable to fulfill request: "
                           "%s; %s", ex, traceback.format_exc())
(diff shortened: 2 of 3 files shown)


dhoomakethu:
These changes make the tests fail on python2.7 and mac. @starnight Could you please try to address this ?

starnight:
> These changes make the tests fail on python2.7 and mac. @starnight Could you please try to address this ?

Do you mean the SyntaxError `yield from` in Python 2.7?

dhoomakethu:
Yes, that's correct.

tracernz:
It looks like the tests used to be guarded by [0] but perhaps adding skips to individual functions in that class overrides the class level skip? [1] I couldn't find anything saying that in the pytest docs though...

[0] https://github.com/riptideio/pymodbus/blob/762ebfbae13a27330e6772af7b95b3157dc87652/test/test_server_asyncio.py#L38
[1] https://github.com/riptideio/pymodbus/pull/444/files#diff-4f350ea6424f6092cb332d04697d0563R79

starnight:
I tried to add skips to individual functions like [1].  But it still hits the same SyntaxError `yield from` in Python 2.7. [2]
I think this is build time error, not run time error.

[1] https://github.com/starnight/pymodbus/commit/f01cf8fc7668
[2] https://travis-ci.org/starnight/pymodbus/jobs/586598803#L387

dhoomakethu:
@starnight Please refer #452 and see if the changes there fixes the test failures in your fork.

starnight:
I rebased upon tags/v2.3.0rc1.  And it passes the tests.  So, Yes!

According to commit cc6e976b81aa "Fix asynci server test failures on python3.6 and below ", it seems the commits: "test: Make assert_called_once() test only with Python 3.6+" and "test: Make serve_forever() test only with Python 3.7+" in branch starnight:fix-serial-test can be reverted.  @dhoomakethu any coments?

dhoomakethu:
That's correct, asyncio server implementation uses 3
7+ specific asyncio constructs and hence disabled for 3.6 and below.

Issue #476, opened 2019-12-17
MyPy type annotations


lovasoa:
### Description

Type annotations allow programmers to automatically check whether code uses APIs with the right parameter types. They are very useful and often prevent bugs. It would be great if this project could provide [mypy type annotations](https://mypy.readthedocs.io/en/latest/stubs.html).

dhoomakethu:
@lovasoa thanks for the suggestion. Right now I do not have enough bandwidth for this. I will be more than happy to accept a PR.

lovasoa:
Great ! Can we keep the issue open until then ? It will give it more visibility. Maybe you could add a "help wanted" label ?

PR #478, opened 2019-12-18
Fix #477

### LINKED ISSUE #477: Asyncio server fails on every connection when the logging level is set to "debug"
### Pymodbus Specific
* Server: tcp - async with asyncio
* OS: MacOS 10.15.1

### Description

The asyncio server raises an error on every connection when setting it to listen on "localhost" on MacOS.

### Code and Logs

```python
import asyncio
from pymodbus.server.asyncio import StartTcpServer

import logging

logging.basicConfig()
logging.getLogger().setLevel(logging.DEBUG)

context = ModbusServerContext(slaves=ModbusSlaveContext(), single=True)
asyncio.run(StartTcpServer(context, address=("localhost", 5020), defer_start=False))
```

```
python  mbserver.py & nc localhost 5020
```

```
DEBUG:pymodbus.server.asyncio:Datastore unable to fulfill request: not all arguments converted during string formatting; Traceback (most recent call last):
  File ".../venv/lib/python3.7/site-packages/pymodbus/server/asyncio.py", line 59, in connection_made
    _logger.debug("Socket [%s:%s] opened" % transport.get_extra_info('sockname'))
TypeError: not all arguments converted during string formatting
```

```
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/events.py", line 88, in _run
  File ".../venv/lib/python3.7/site-packages/pymodbus/server/asyncio.py", line 247, in connection_made
    _logger.debug("TCP client connection established [%s:%s]" % self.client_address)
```

Edit: Removed inaccurate diagnostic.

lovasoa:
Fix error when connection to the asyncio client and logging is in "debug" verbosity level.

There may be other places where the same problem occurs. It would be great if the type of variables were somehow indicated (see #476). This would avoid this kind of bugs.

### DIFF
--- a/pymodbus/server/asyncio.py
+++ b/pymodbus/server/asyncio.py
@@ -244,12 +244,12 @@ class ModbusConnectedRequestHandler(ModbusBaseRequestHandler,asyncio.Protocol):
 
         self.client_address = transport.get_extra_info('peername')
         self.server.active_connections[self.client_address] = self
-        _logger.debug("TCP client connection established [%s:%s]" % self.client_address)
+        _logger.debug("TCP client connection established [%s]" % self.client_address)
 
     def connection_lost(self, exc):
         """ asyncio.BaseProtocol: Called when the connection is lost or closed."""
         super().connection_lost(exc)
-        _logger.debug("TCP client disconnected [%s:%s]" % self.client_address)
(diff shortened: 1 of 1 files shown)


lovasoa:
Sorry, I don't think this is the right fix for #477 ...

PR #504, opened 2020-05-19
Fix comment and assertions in synchronous_client.py

erlend-aasland:

PR #505, opened 2020-05-19
Fix asserts in synchronous_client.py

erlend-aasland:
* Line 131 and 147 asserts the wrong variable
* Add asserts for the remaining queries


erlend-aasland:
Will merge with PR #504

Issue #513, opened 2020-07-19
asyncio.coroutine has been deprecated since Python 3.8 and will be removed in future


tirkarthi:
### Versions

* Python: 3.8
* OS: Ubuntu
* Pymodbus: master branch
* Modbus Hardware (if used): 

### Description

asyncio.coroutine has been deprecated since Python 3.8 and will be removed in future. Use `async def` instead.

### Code and Logs

```
pymodbus/client/asynchronous/asyncio/__init__.py
249:    @asyncio.coroutine
286:    @asyncio.coroutine
329:    @asyncio.coroutine
379:    @asyncio.coroutine
462:    @asyncio.coroutine
509:    @asyncio.coroutine
554:    @asyncio.coroutine
611:    @asyncio.coroutine
710:    @asyncio.coroutine
760:@asyncio.coroutine
777:@asyncio.coroutine

test/test_server_asyncio.py
69:    @asyncio.coroutine
80:    @asyncio.coroutine
88:    @asyncio.coroutine
96:    @asyncio.coroutine
109:    @asyncio.coroutine
143:    @asyncio.coroutine
183:    @asyncio.coroutine
216:    @asyncio.coroutine
248:    @asyncio.coroutine
286:    @asyncio.coroutine
322:    @asyncio.coroutine
361:    @asyncio.coroutine
407:    @asyncio.coroutine
425:    @asyncio.coroutine
433:    @asyncio.coroutine
449:    @asyncio.coroutine
464:    @asyncio.coroutine
485:    @asyncio.coroutine
527:    @asyncio.coroutine
565:    @asyncio.coroutine

test/asyncio_test_helper.py
24:            @asyncio.coroutine
```

JamesJeffryes:
duplicates #467 but with more complete information.

swamper123:
This issue is already solved in branch 3.0.

dhoomakethu:
Please be aware, we are supporting 3.7 also so removing co-routines completely would break things for pymodbus on python3.7

swamper123:
Isn't this decorator not just a compatible thing to async/await syntax?
Since my PR is a while ago, I will take a look again if I have fixed that the way it was right.

swamper123:
Well, it looks like I have just replaced one decorator, but replaced it with modern syntax.
I guess this is for codecompatibility with older versions, but isn't used anymore and not important for >= 3.7.
Even you replaced one last year. ;)

dhoomakethu:
[...]

Issue #467, opened 2019-11-11
Deprecated coroutine


albireox:
### Versions

* Python: 3.8
* OS: macOS 10.15
* Pymodbus: 2.3.0
* Modbus Hardware (if used): WAGO

### Pymodbus Specific
* Client: tcp - async

### Description

Trying to import `from pymodbus.client.asynchronous.asyncio import AsyncioModbusTcpClient` produces a series of messages `[WARNING]: "@coroutine" decorator is deprecated since Python 3.8, use "async def" instead (DeprecationWarning)`. Python 3.8 has deprecated the `coroutine` decorator. Also, there is a filter in pymodbus that always shows these warnings so it's not possible to suppress them.

JamesJeffryes:
@dhoomakethu I'm happy to open a PR to address this. One thing to note is that converting to use of the async keyword would limit the library to 3.5+ but that seems like a pretty reasonable limitation to me.

dhoomakethu:
Yes, pymodbus 3 would be only available for Python 3.5 and above and it should be good to have that change. Feel free to raise a PR

swamper123:
This was cleaned up with #564 .
At least I couldn't find any decorator.

PR #534, opened 2020-09-23
Use async/await keywords in test_server_asyncio.py

JamesJeffryes:
I'm planning on splitting up a fix to #513 & #467 into a couple of PRs as there's more edits than I expected.

While I was replacing the old async syntax, I took a moment to do some reformatting, remove unused imports and replace the unittest methods with pytest syntax.

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -1,3 +1,4 @@
+attrs>=20.2.0
 bcrypt>=3.1.6
 capturer >= 2.2
 coverage >= 4.2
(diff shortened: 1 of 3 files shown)


dhoomakethu:
@JamesJeffryes Thanks for the PR, looks like this is breaking tests in all other python versions except 3.8. Would you be able to take a look please ?

JamesJeffryes:
Fixed most of the other builds by specifying a minimum version of 'attrs' but the xcode tests looks like they are timing out trying to install java. Maybe there's a more modern image that could be used?

JamesJeffryes:
More general question, if this code is bound for a 3.5+ release. Should I pull out the IS_PYTHON3 checking in the tests as I go?

JamesJeffryes:
@dhoomakethu Thoughts on the failing xcode build?

dhoomakethu:
Will take a look, osx builds are failing off late. Could be the xcode version that is causing the issue.

dhoomakethu (approved):

PR #535, opened 2020-10-01
Update requirements-checks.txt

mdismailmd44:
It looks wise in == as compared to >=


janiversen:
this will/can cause problems with programs that uses pymodbus and e.g. a newer version of flake8, so since pymodbus works fine with >= please do not change that.

dhoomakethu:
@mdismailmd44 Thanks for the PR. However this is not required keeping in mind the changes in [newer pip versions](https://discuss.python.org/t/announcement-pip-20-2-release/4863)

PR #556, opened 2020-11-01
Implement asyncio ModbusSerialServer

d21d3q:
Main goal when implementing ModbusSerialServer was to have "bulletproof" server - `autoreconnect` option set to `True` will cause server to start/run even with specified port is not present. The use case is to be able to embed modbus server to some embedded device and forget about it, so that if server is running and user disconnects serial port physically and plugs it back, server will continue to operate. Same as server will be restarted when serial device is not present, then server will wait for it to become available and continue to serve.  
Setting `autoreconnect` to False will cause it to fail on start if serial device is not present.

I am creating this PR as draft so that you can review it and let me know if it right direction of development.

### DIFF
--- a/examples/common/asyncio_server.py
+++ b/examples/common/asyncio_server.py
@@ -140,17 +140,16 @@ async def run_server():
     # #
     # await server.serve_forever()
 
-    # !!! SERIAL SERVER NOT IMPLEMENTED !!!
     # Ascii:
-    # StartSerialServer(context, identity=identity,
-    #                    port='/dev/ttyp0', timeout=1)
+    # await StartSerialServer(context, identity=identity,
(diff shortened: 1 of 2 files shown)


dhoomakethu (commented):
The Serial handler will not have `get_extra_info`. In `ModbusBaseRequestHandler` You will have to handle the logging in else clause. Please add a new clause for `ModbusSingleRequestHandler` .May be self.transport.serial.port here ?

dhoomakethu:
@d21d3q I have picked up your code and added it to  branch `2.5.0`. Please take a look or raise the PR against it so that if I have missed some new changes, that could be captured

dhoomakethu:

Issue #557, opened 2020-11-05
Create an AsyncModbusTCPClient from within a coroutine blocks


tiagocoutinho:
### Versions

* Python: 3.7, 3.8 or 3.9
* OS: Linux Mint 20 Ulyana
* Pymodbus: 2.4.0
* Modbus Hardware (if used): examples/common/synchronous_server.py

### Pymodbus Specific
* Server: tcp - sync
* Client: tcp - asyncio

### Description

Trying to create an AsyncModbusTCPClient from within a coroutine blocks.

I expect the AsyncModbusTCPClient to return a tuple <event_loop, client> but instead there is not even an attempt to connect with the server.

### Code and Logs

The following simple example blocks without even connecting to the server:

```python
import asyncio
import logging

from pymodbus.client.asynchronous import schedulers
from pymodbus.client.asynchronous.tcp import AsyncModbusTCPClient as ModbusClient

async def main():
    loop = asyncio.get_event_loop()
    _, client = ModbusClient(schedulers.ASYNC_IO, port=5020, loop=loop)
    print("Reading coils")
    rr = await client.read_coils(1, 1, unit=0x01)

asyncio.run(main())

```

PR #558, opened 2020-11-05
Be able to create an AsyncModbusTCPClient from within a coroutine

tiagocoutinho:
This PR solves #557 by making it possible to create an AsyncModbusTCPClient from within a coroutine.

The change is that, if called from within the event loop, the AsyncModbusTCPClient returns a coroutine which, when awaited, should return a ModbusTCPClient object.

Here is how it would look like in practice:

```python
import asyncio

from pymodbus.client.asynchronous import schedulers
from pymodbus.client.asynchronous.tcp import AsyncModbusTCPClient

async def main():
    client = await AsyncModbusTCPClient(schedulers.ASYNC_IO, port=5020)
    print("Reading coils")
    rr = await client.read_coils(1, 1, unit=0x01)

asyncio.run(main())
```
Missing:
* [ ] Documentation
* [ ] Unit test

### DIFF
--- a/pymodbus/client/asynchronous/factory/tcp.py
+++ b/pymodbus/client/asynchronous/factory/tcp.py
@@ -91,17 +91,19 @@ def async_io_factory(host="127.0.0.1", port=Defaults.Port, framer=None,
     """
     import asyncio
     from pymodbus.client.asynchronous.async_io import init_tcp_client
-    loop = kwargs.get("loop") or asyncio.new_event_loop()
+    loop = kwargs.get("loop") or asyncio.get_event_loop()
     proto_cls = kwargs.get("proto_cls", None)
     if not loop.is_running():
         asyncio.set_event_loop(loop)
         cor = init_tcp_client(proto_cls, loop, host, port)
         client = loop.run_until_complete(asyncio.gather(cor))[0]
     else:
-        cor = init_tcp_client(proto_cls, loop, host, port)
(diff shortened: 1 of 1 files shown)


dhoomakethu:
This might also be applied to all the other asyncio clients (udp, tls, serial).

memetb:
This change should probably have a consistent return signature by returning the `loop, client` tuple [here](https://github.com/riptideio/pymodbus/pull/558/commits/ee65a9093a734af87036b7aef0cc60e4a53d55ce#diff-af83448f906094a8f95088d50466232c67890743f4bf4fa653d0096fa19cd2bbR102).

Also, maybe I'm missing something but calling [get_event_loop()](https://github.com/riptideio/pymodbus/pull/558/commits/ee65a9093a734af87036b7aef0cc60e4a53d55ce#diff-af83448f906094a8f95088d50466232c67890743f4bf4fa653d0096fa19cd2bbR101) will/can create a new loop if none is present. As per [documentation](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop)

> Because this function has rather complex behavior (especially when custom event loop policies are in use), using the get_running_loop() function is preferred to get_event_loop() in coroutines and callbacks.

Emilv2:
This feature would be much appreciated and it seems like it's not that much work. What is holding it up and could I maybe help with this?

dhoomakethu:
@Emilv2 please refer to @memetb's comment.

dhoomakethu:

PR #562, opened 2020-11-30
other: reportSlaveId: allow slaves to provide custom responses

karlp:
At present, function code 17, "Report Slave ID" is hardcoded to reply
with a string joined form of the "device identification" fields from
function code 43, subcode 14 (0x2b, 0x0e)

There's nothing that actually implies these should be related.  This
change allows slave contexts to provide custom data to return via the
ModbusServerContext property reportSlaveIdData

If this is not provided, the existing fallback to the device
identification fields is used.

Signed-off-by: Karl Palsson <karlp@etactica.com>

I'm using this to make better fakes for existing devices, and those existing devices very much don't have the same data in both requests!

Open notes:
* Should this be done via extension of the "Datastore" instead?
* How should this be documented? changed to a function on IModbusSlaveContext instead?
* tests for this function don't have a context, should they?

### DIFF
--- a/pymodbus/other_message.py
+++ b/pymodbus/other_message.py
@@ -360,21 +360,26 @@ class ReportSlaveIdRequest(ModbusRequest):
         pass
 
     def execute(self, context=None):
-        ''' Run a read exeception status request against the store
+        ''' Run a report slave id request against the store
 
         :returns: The populated response
         '''
-        information = DeviceInformationFactory.get(_MCB)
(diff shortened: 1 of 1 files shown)


karlp:
Test failures don't seem to be related?

dhoomakethu (approved):

dhoomakethu:
@karlp Seems to be setup issue. Will check the test failures. I am merging this to `dev` for the time being.

PR #574, opened 2020-12-17
Disable strict mode for inter character timeout in Serial mode by default

carlosgs91:
The _strict argument should be defaulted to False since the inter_char_timeout feature is not working correctly. Changes in the past have tried to fix the issue but one way or another they do not seem to work consistently across all computers and systems. Therefore I think by default the _strict argument should be set to False unless specified to prevent the faulty behavior from producing false timeouts.

There are two possible reasons why I think this feature may never work correclty:
* One the OS handling the timeout incorrectly if more than one character arrives at the same time if it is busy doing something else.
* Another one as described here https://stackoverflow.com/a/37053542/1813498 is the fact that the timeout value must be a multiple of 0.1. At 9600 the inter char value is calculated to be 0.00171875 and then rounded to 0 so therefore it will not work correctly.

The conclusion from my side is that it is not possible to accurately measure this timeout and should perhaps be removed. At least it should be by default disabled. This small patch aims to do the later.

### DIFF
--- a/pymodbus/client/sync.py
+++ b/pymodbus/client/sync.py
@@ -552,7 +552,7 @@ class ModbusSerialClient(BaseModbusClient):
         self.parity = kwargs.get('parity',   Defaults.Parity)
         self.baudrate = kwargs.get('baudrate', Defaults.Baudrate)
         self.timeout = kwargs.get('timeout',  Defaults.Timeout)
-        self._strict = kwargs.get("strict", True)
+        self._strict = kwargs.get("strict", False)
         self.last_frame_end = None
         self.handle_local_echo = kwargs.get("handle_local_echo", False)
         if self.method == "rtu":


dhoomakethu:
Thanks for the PR @carlosgs91 , could you please raise this against `2.5.0`

carlosgs91:
@dhoomakethu What do you mean to raise against 2.5.0? Should I add the change to release branch for version 2.5.0? What about 3.0.0?

dhoomakethu:
@carlosgs91 you have raised this against `master` branch. You will have to raise this against branch `2.5.0` . I will have `3.0.0` released in the first quarter of 21 as there is still a lot more work pending in there.

carlosgs91:
@dhoomakethu Ah ok. Which versions should we put it then? I would assume if we put it in master it will travel all the way to the future versions that's why I tried to merge it there. Should I put it in 2.5.0, 3.0.0 and master then?

dhoomakethu:
2.5.0 or dev

ಸೋಮ, ಡಿಸೆಂ 21, 2020 19:01 ದಿನಾಂಕದಂದು carlosgs91 <notifications@github.com>
ಅವರು ಬರೆದಿದ್ದಾರೆ:

dhoomakethu:
@carlosgs91 I have cherry picked this change to 2.5.0. Thanks

carlosgs91:
@dhoomakethu I have cherry picked it into 3.0.0 and dev so it is ready for future versions:

* https://github.com/riptideio/pymodbus/pull/584
* https://github.com/riptideio/pymodbus/pull/585

Issue #538, opened 2020-10-06
Previous response from modbus device returned as current response when using sync TCP over network with lag


thormick:
### Versions

* Python: 3.8.2
* OS: Ubuntu 20.04.1 LTS
* Pymodbus: 2.4.0
* Modbus Hardware (if used): Hardware independent, verified with local testing against dummy sync server

### Pymodbus Specific
* Server: tcp - sync
* Client: tcp - sync

### Description
TCP sync requests over a laggy network connection against different ranges of registers eventually give the timed-out result of one register range performed previously as the reading of the other register range.

Because the transaction manager and socket framer ignore the transaction ID of the response (my client's app uses synchronized TCP requests) and just returns the response that has come in last, pymodbus can under certain circumstances silently return the result of the previous request made as if it was the result to the one being currently made. This is Bad for my client, whose application makes sequential modbus requests to different ranges of registers of a single unit and needs to communicate over networks with sporadic periods of high lag (problematic VPN connections, mobile networks, etc...).

A step-by-step description of what happens, where an application is set up to read two or more separate ranges of registers in sequential fashion against a single device, and there is network lag. What will happen sooner or later is:

1. Read against register range 1 fails due to timeout. `ModbusTransactionManager._recv` raises `InvalidMessageReceivedException`. This is caught in `ModbusTransactionManager._transact`, which closes the connection, and as such the data will not cause an issue when it *does* arrive. `ModbusTransactionManager.execute` registers this as that the device didn't respond in a timely fashion, and adds it to the `_no_response_devices` list. My client's application sees that there is no valid response, logs an error and sleeps a few seconds.
2. Read against register range 1. pymodbus sees the connection is closed and opens it. There is another timeout, but because the device this time is already on the `_no_response_devices`, there is no exception from `ModbusTransactionManager._recv`, even if it fails to read any data before the timeout, and the connection is *not* closed. `ModbusTransactionManager.execute` just returns an exception with "No Response received from the remote unit/Unable to decode response". My client's application sees that there is no valid response from this register range, logs an error and sleeps a few seconds.
[...]

PR #575, opened 2020-12-18
Fixes for synchronized client and transaction handling

thormick:
Fixes in the synchronized client and the parts of the transaction manager used by the synchronized client.

This has all the work I did in the sync client for [gapit.io](https://github.com/gapitio). Let me know if this should be split up, for instance with just the #538 fix. The other commits are for issues I discovered while digging to find and fix that issue.

Sorry for not getting around to this before, things have popped up so that even if this sat almost ready and just needed a once or twice over it still got punted a month, but at least now it's presentable enough for a PR.

No sweat about merging all of this, gapit.io is using a fork for pymodbus, has been doing so for a long time and can do so for a while longer.

### DIFF
--- a/pymodbus/client/sync.py
+++ b/pymodbus/client/sync.py
@@ -239,7 +239,12 @@ class ModbusTcpClient(BaseModbusClient):
         """ Reads data from the underlying descriptor
 
         :param size: The number of bytes to read
-        :return: The bytes read
+        :return: The bytes read if the peer sent a response, or a zero-length
+                 response if no data packets were received from the client at
+                 all.
+        :raises: ConnectionException if the socket is not initialized, or the
+                 peer either has closed the connection before this method is
+                 invoked or closes it before sending any data before timeout.
         """
         if not self.socket:
             raise ConnectionException(self.__str__())
@@ -270,6 +275,9 @@ class ModbusTcpClient(BaseModbusClient):
             ready = select.select([self.socket], [], [], end - time_)
             if ready[0]:
                 recv_data = self.socket.recv(recv_size)
+                if recv_data == b'':
+                    return self._handle_abrupt_socket_close(
+                        size, data, time.time() - time_)
                 data.append(recv_data)
                 data_length += len(recv_data)
             time_ = time.time()
@@ -285,6 +293,35 @@ class ModbusTcpClient(BaseModbusClient):
                 break
 
         return b"".join(data)
+
+    def _handle_abrupt_socket_close(self, size, data, duration):
+        """ Handle unexpected socket close by remote end
+
+        Intended to be invoked after determining that the remote end
+        has unexpectedly closed the connection, to clean up and handle
+        the situation appropriately.
+
+        :param size: The number of bytes that was attempted to read
+        :param data: The actual data returned
+        :param duration: Duration from the read was first attempted
+               until it was determined that the remote closed the
+               socket
+        :return: The more than zero bytes read from the remote end
+        :raises: ConnectionException If the remote end didn't send any
+                 data at all before closing the connection.
+        """
+        self.close()
+        readsize = ("read of %s bytes" % size if size
+                    else "unbounded read")
+        msg = ("%s: Connection unexpectedly closed "
+               "%.6f seconds into %s" % (self, duration, readsize))
+        if data:
+            result = b"".join(data)
+            msg += " after returning %s bytes" % len(result)
+            _logger.warning(msg)
+            return result
+        msg += " without response from unit before it closed connection"
+        raise ConnectionException(msg)
 
     def is_socket_open(self):
         return True if self.socket is not None else False
--- a/pymodbus/client/sync.py
+++ b/pymodbus/client/sync.py
@@ -261,9 +261,9 @@ class ModbusTcpClient(BaseModbusClient):
 
         timeout = self.timeout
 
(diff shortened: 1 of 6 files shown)


thormick:
I see that the diagnostic client in acdd9e9e61aa4e122cb8864e4e9315211115c195 caused those issues, that can easily be backed out.

pymodbus/client/sync.py
```diff
@@ -286,6 +294,35 @@ def _recv(self, size):
...
+               until it was determined that the remote closed the
+               socket
+        :return: The more than zero bytes read from the remote end
+        :raises: ConnectionException If the remote end didn't send any
+                 data at all before closing the connection.
+        """
+        self.close()
+        readsize = (f"read of {size} bytes" if size
+                    else "unbounded read")
+        msg = (f'{self.__str__()}: Connection unexpectedly closed '
```
dhoomakethu, line 316: fstrings are supported only in python3, this would essentially the PR valid for pymodbus 3.x. If you intend to support to python 2.x or want this to go in to pymodbus 2.5.0 please update accordingly. Otherwise please raise this against branch 3.0.0

pymodbus/client/sync.py
```diff
@@ -256,9 +261,9 @@ def _recv(self, size):
 
         timeout = self.timeout
 
-        # If size isn't specified read 1 byte at a time.
+        # If size isn't specified read up to 4096 bytes at a time.
         if size is None:
-            recv_size = 1
+            recv_size = 4096
```
dhoomakethu, line 266: This value is debatable. This would essentially block the `recv` till all `4096` bytes are received or timeout. Adding further to delay in reads.

pymodbus/client/sync.py
thormick: Thank you, good catch, this was a copy from the custom branch that Gapit uses, which is only 3.0. I don't have any thought of whether this should go into the 2.x or 3.x versions. Might as well update it to work for 2.x.

pymodbus/client/sync.py
thormick: That's not how `recv` works. From the documentation:
```
$ pydoc socket.socket.recv | more
Help on method_descriptor in socket.socket:

socket.socket.recv = recv(...)
    recv(buffersize[, flags]) -> data
    
    Receive up to buffersize bytes from the socket.  For the optional flags
    argument, see the Unix manual.  When no data is available, block until
    at least one byte is available or until the remote end is closed.  When
    the remote end is closed and all data is read, return the empty string.
```

So it never blocks when a single byte is available, regardless of what `buffersize` is provided. I.e. it's a maximum limit, the minimum limit is always 1 regardless.

thormick:
I've got the fstring thing sorted out, added and changed the unit tests a bit, and done a few other fixes too. But I'm not sure about intricacies regarding GitHub; is it correct to force push to the branch this PR is based off? Or should this PR be closed and another opened? The revised version resides in the branch at https://github.com/gapitio/pymodbus/tree/fix-sync-client-trans-2 for now.

pymodbus/client/sync.py
tracernz: @thormick That matches the underlying recv man page too:

https://man7.org/linux/man-pages/man2/recv.2.html
> If no messages are available at the socket, the receive calls wait for a message to arrive, unless the socket is nonblocking (see fcntl(2)), in which case the value -1 is returned and the external variable errno is set to EAGAIN or EWOULDBLOCK.  The receive calls normally return any data available, up to the requested amount, rather than waiting for receipt of the full amount requested.

dhoomakethu:
@thormick you can push the changes to the same PR, I will cherry-pick the changes to `2.5.0` and from there it gets merged to `dev` and subsequently to `master` once `2.5.0` is officially released. Currently I have the  release candidate `2.5.0rc2` 
 on pypi. May be this PR will go with `2.5.0rc3`.

pymodbus/client/sync.py
dhoomakethu: ah, I got this confused with `Serial`.Apologies for the mistake. Looks good.

thormick:
@dhoomakethu So done, I've rammed it in now. The major difference from the previous iteration is all the changes to the test_client_sync.py. Most of it is just to increase coverage enough that the CI doesn't complain, but I've also included the admittedly highly brazen change of patching out `time`. This will hopefully make the test more deterministic, and in my tests more than halves the time it takes to run the entire unittest suite. I can back that out again if it's too radical.

pymodbus/client/sync.py
thormick: Actually, it's good that you picked on it, it made me gave it a quick look over and I realize now that `ModbusTlsClient` does the same, and now those two will be inconsistent. I don't have a setup ready for testing modbus TLS, so I'm not sure about touching that part of the code. Not necessarily a problem, though?

thormick:
Made some small additions to the unit tests. Commit-to-commit diff relative to my Jan 8 push [here](https://github.com/gapitio/pymodbus/compare/gapitio:946297439374d8f11b1805174d534394e4c6000f..gapitio:24c0ffc975c40a9dcab9620847a195f5565eaad3).

dhoomakethu:
@thormick I have merged the changes in this PR in #605 . Will be closing this PR.

thormick:
@dhoomakethu Thank you very much, glad you could make use of it all.

PR #577, opened 2020-12-24
Fixed driver name and logical reference

meeas:


dhoomakethu:
@meeas I will be cherry picking this to 2.5.0. Thanks

Issue #581, opened 2021-01-02
Add Trio support


altendky:
I am interested in having PyModbus support Trio async in addition to asyncio, Twisted, and Tornado.  No promises I will succeed but I have started exploring the code and am trying to piece together some amount of Trio functionality.  I figured it would be good to publicly document the interest and whatever efforts I make in case anyone else is interested or knows anything on the topic.

In case there is curiosity about my context, my particular use will be for [SunSpec](https://sunspec.org/) communication within a Qt Trio application using [QTrio](https://github.com/altendky/qtrio).  There are [pySunSpec](https://github.com/sunspec/pysunspec) and [pySunSpec2](https://github.com/sunspec/pysunspec2) but given that I wasn't able to get any interest from them in my working Twisted support several years ago I hesitate to invest again.  I'm not sure why they wrote their own Modbus anyways...  My expectation is to develop PyModbus Trio support and then add a SunSpec layer on top of that using pySunSpec2's data model but not its communication code.

altendky:
The PR is in work and waiting on others to get merged so as to have responsive CI.  Also, I've been on other stuff for a bit.

altendky:
Still running around on other tasks and haven't made it back here yet.

altendky:
Well, the PR has finally been submitted for review so...  `:]`  (open sesame)

altendky:
meh...

dhoomakethu:
Adding label so that this wont get stale.

PR #583, opened 2021-01-03
Trio support

altendky:
#581

Draft for:
- [x] Being more than an exploration
- [x] Server tests
- [x] Client tests
- [x] Documentation
- [x] Examples


altendky:
`tc.py` does run against `examples/common/synchronous_server.py` and get answers.  Though, only the coil read seems correct.  Anyways, I just tend to share sooner than later.  I have no expectation that anyone will really want to read this at the moment.  `:]`

requirements.txt
```diff
@@ -29,6 +29,11 @@ six>=1.15.0
 # -------------------------------------------------------------------
 # pyserial-asyncio==0.4.0;python_version>="3.4"
 
+# -------------------------------------------------------------------
+# if you want to use the Trio asynchronous version, uncomment these
+# -------------------------------------------------------------------
+# async_generator ~= 1.10
```
dhoomakethu, line 35: can we add this as part of `extra_requires` in [setup.py](https://github.com/riptideio/pymodbus/blob/master/setup.py#L76)

altendky: Certainly, that's how I usually do it but managed to overlook that here.  `:[`  Thanks for catching it.

altendky:
Mostly just for linkage, I have a quick SunSpec layer running on top of this PR over at https://github.com/altendky/ssst/pull/19.  That code will likely move out to it's own project once this all settles down a bit.  Anyways, this functions in a basic scenario.  Maybe I'll take another pass at this in the next couple days.

test/test_server_trio.py
```diff
@@ -0,0 +1,189 @@
...
+        host=trio_tcp_server.host,
+        port=trio_tcp_server.port,
+    )
+
+    async with modbus_client.manage_connection() as protocol:
+        yield protocol
+
+
+@pytest.mark.trio
+async def test_read_holding_registers(trio_tcp_client, trio_tcp_server):
```
dhoomakethu, line 90: These tests looks more like functional tests than unit tests. I can move this to functional tests later to #2 when functional tests actually are in place. Also the test seems to be failing on mac osx with ` trio.TrioInternalError: internal error in Trio - please file a bug!`.

dhoomakethu:
@altendky sorry for picking on this late, can you also add an example for trio client?

pymodbus/client/asynchronous/factory/tcp.py
```diff
@@ -105,6 +105,25 @@ def async_io_factory(host="127.0.0.1", port=Defaults.Port, framer=None,
...
+    :param source_address: Bind address
+    :param timeout: Timeout in seconds
+    :param kwargs:
+    :return: tcp client
+    """
+    from pymodbus.client.asynchronous.trio import init_tcp_client
+    proto_cls = kwargs.get("proto_cls", None)
+    client = init_tcp_client(proto_cls, host, port)
+
+    return client
```
dhoomakethu, line 124: This is inconsistent with other factories which return two values.

altendky:
I was going to come back to this after the CI was redone in #592 so I could have functional CI while working on this.  Adding examples is already in the list in the first post explaining why this is a draft PR.  I'll address the other concerns when working on this again.  Thanks for the feedback.

pymodbus/client/asynchronous/factory/tcp.py
altendky: They do return two values, but I don't see the consistency in the meaning of those two values.  Can you help me understand what sort of thing the first value should be and what the second should be?  Then I can try to map that into the Trio stuff.

Twisted: protocol / deferred
Tornado: protocol / future
asyncio: event loop / client
Trio: ? / ?

I can see the similarity between Twisted and Tornado.  I don't see how asyncio fits.  Apparently when I coded this a couple months ago I didn't see what other useful thing I could return in the Trio case.  We'll see if I can find something now.

altendky:
(hopefully I didn't already comment on this and forget...)  @dhoomakethu, any interest in enabling read the docs builds for each PR?  If you go to https://readthedocs.org/dashboard/pymodbus/advanced/ there should be a checkbox labeled `Build pull requests for this project`.  Beyond just making sure the docs also build on RTD, this acts as a reminder to keep them up to date and let's contributors easily sea the rendered result.

pymodbus/client/asynchronous/factory/tcp.py
altendky: Relatedly, the following docstring seems to only get asyncio right.

https://github.com/riptideio/pymodbus/blob/c7a582b6796be211d693dba80367f2b26846c134/pymodbus/client/asynchronous/__init__.py#L15-L22

test/test_server_trio.py
altendky: I added a bunch more unity tests as well.

altendky:
Welp, @dhoomakethu I think this is ready for at least a second pass review.  I understand that I may not be testing sensibly or need a pile more docs or...  So, don't feel obliged to dig into the details if something big is out of line.  Anyways, let me know what you think.

Side note, I took this on so I could develop a Trio library for communication with [SunSpec](https://sunspec.org/) devices.  That ended up being [sundog](https://sundog.readthedocs.io/en/latest/).

dhoomakethu:
@altendky it will be sometime before I go through this. Apologies in advance and thanks for the efforts.

altendky:
I understand.  Took me awhile to get back to it myself and I've got a few other projects I'm playing catch up on.  No worries.

altendky:
The PyPy issues should get fixed by https://github.com/riptideio/pymodbus/pull/637.

janiversen:
As discussed on the issue, I close this for now.

PR #589, opened 2021-01-14
Warn when using deprecated clients

Emilv2:
Instead of the other way around, warning when not using them.

I think that makes way more sense, now I get warnings when I'm *not* using anything deprecated and it looks like I'm missing dependencies. At most that message should be `info` or `debug`.


dhoomakethu:
This is merged with #578 . Closing.

PR #590, opened 2021-01-15
Upgrade pip for CI

altendky:
Perhaps this will help with attrs etc versions


dhoomakethu:
This is merged with #578 . Closing.

altendky:
Note that this was marked draft and wasn't intended to be merged.  But the intent is that the Travis config gets removed entirely in #592 so maybe it doesn't matter.

PR #593, opened 2021-01-16
Test Windows and macOS

altendky:
Draft for:
- [x] Do not merge until ~~#592~~ #605 has been merged as this is built from and includes the changes in ~~#592~~ #605.
- [x] Also, get the tests passing...
- [x] Clean up the hacky exploratory "fixes"


altendky:
@dhoomakethu, I'm at a bit of a loss as to how this would have been working much recently at all.  Are Windows tests run much anywhere?  I know they weren't on Travis but I don't know what devs there are and if any use Windows.  That said, if you look on the Twisted side you end up seeing my name wound up in the serial stuff so...  Funny thing is that four years ago I was working on the same goal of a Qt+Modbus application and so I needed Twisted serial.  Got Twisted serial going, but not the application, and now I'm back at it again trying to use Trio and getting bit by some Twisted serial-related stuff.  `:|`

For example, why is a serial port created for any of these except the RTU test?

https://github.com/riptideio/pymodbus/blob/d4f5f9e515e589a863ba2148edf6cdf656e551df/test/test_client_async.py#L178-L182

Are you expected to have a serial adapter attached to be able to run the tests?  Maybe even two hooked to each other to talk back and forth?  Though this bit makes it look like Windows just wasn't ever setup to be testable.

https://github.com/riptideio/pymodbus/blob/d4f5f9e515e589a863ba2148edf6cdf656e551df/test/test_client_async.py#L32-L39

Presumably I can do at least some of the nitty gritty but it would probably be good for me to understand what the present state and any history of the Windows support.  If it weren't for the `setup.py` I guess I might assume that Windows support simply needed to be developed.  `:]`  Though, I develop in Linux, I do need Windows support along with Trio work I've done.

https://github.com/riptideio/pymodbus/blob/efb90fb05882d606933a97435dd37ae87ead5449/setup.py#L72

altendky:
I should probably clarify...  I need pymodbus/Trio/TCP support on Windows...  which so far seems fine on my branch (used in my application tests across Linux/macOS/Windows).  But, I also would like pymodbus/Trio/RTU support.  And, I don't super like leaving the rest of a lib hanging around while only making sure my little corner works.  So I'm still interested in putting in some effort on this point.

altendky:
And...  green.  \*sigh\*  Now to go back and remember what all the changes were for and conclude if they are good or a mess...  Also maybe extract the PyPy support changes.

altendky:
@dhoomakethu, not sure if you are notified as I remove the draft status from the PR (and there's still the draft tag) so here's a quick note that this is ready for review.  Thanks for the progress.  `:]`

dhoomakethu:
@altendky Thanks for the headsup. I will merge this to `dev` over the weekend.

altendky:
I hope it isn't a problem, but I went ahead and combined #594 in here.  I don't think there's any remaining need for them to be separate.

PR #596, opened 2021-01-20
Enable coveralls

altendky:
Draft for:
- [ ] Do not merge until #592 has been merged as this is built from and includes the changes in #592.
- [ ] Reach a conclusion about forked PRs and 'secret' tokens etc...  and implement it.


janiversen:
I think the current state of sources is not something we should publish, WDYT ?

Coverage is excluded from not a small part of the sources, and on top of that we only get 85% coverage.

janiversen:
Closing this for now, should you get a some spare cycles I am more than happy to reopen it, and work with you to implement coverall.

PR #597, opened 2021-01-20
Test pypy

altendky:
Draft for:
- [ ] Do not merge until #593 has been merged as this is built from and includes the changes in #593.


altendky:
This separation didn't make sense.  Being addressed in #593.

PR #599, opened 2021-01-29
BUGFIX?:  self._header needs to be set to {}

frawau:
self._header needs to be set to {}. If not, the first frame would always be missed.

populateHeader() is called in isFrameReady() and checkFrame(). In isFrameReady,
populateHeader is only called 'if not self._header', and checkFrame is only called
'if isFrameReady()'.

Reset also sets self._header to {}.

I did test with a async serial connection. Does fix my problem, was only checking for one answer.

### DIFF
--- a/pymodbus/framer/rtu_framer.py
+++ b/pymodbus/framer/rtu_framer.py
@@ -60,7 +60,7 @@ class ModbusRtuFramer(ModbusFramer):
         :param decoder: The decoder factory implementation to use
         """
         self._buffer = b''
-        self._header = {'uid': 0x00, 'len': 0, 'crc': '0000'}
+        self._header = {}
         self._hsize = 0x01
         self._end = b'\x0d\x0a'
         self._min_frame_size = 4


sfarbotka:
Please, take a look at PR #608. It looks like they both a trying to fix the same issue in RTU framer.

pymodbus/framer/rtu_framer.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, decoder, client=None):
         :param decoder: The decoder factory implementation to use
         """
         self._buffer = b''
-        self._header = {'uid': 0x00, 'len': 0, 'crc': '0000'}
+        self._header = {}
```
sfarbotka, line 63: It looks like `isFrameReady()` can still work incorrectly in some conditions.

If framer instance receives incomplete frame with only first byte of some frame, 
`populateHeader()` properly sets `self._header['uid']` and then raises `IndexError` when trying to read second byte from buffer. After that `isFrameReady()` will not call `populateHeader` because `self._header` is not empty, but instead will fail on line `return self._header and len(self._buffer) >= self._header['len']`, as `len` key was not added to `self._header`

sfarbotka (commented):
Please take a look at my comments

dhoomakethu:
@frawau Please check #608 and see if that fixes the issue.

frawau:
Yes I checked and it very much look like the same issue.. only solved better. I have not had time to check it yet... but it does look like it will solve my issue.

dhoomakethu:
@frawau The changes in #608 is merged in to #605. I will be closing this PR. If you think the changes do not fix your issue, please feel free to re-open this.

PR #600, opened 2021-01-30
"Reconnection retries" was added into Modbus TCP Client 

agn-7:
I've added reconnect retries due to the handle of the connection failed at the first time into ModbusTcpClient.


dhoomakethu:
@agn-7 Thanks for the PR . This is a good addition however this is something we can do outside pymodbus library as well. I will be declining this PR to be inconsistent with other pymodbus clients (both sync and async)

PR #615, opened 2021-03-10
Add server requiring client's cert in TLS handshake feature

starnight:
This patch adds server requiring client's certificate feature which is
mentioned in the 6th step CertificateRequest to 9th step
VerifyClientCertSig in Table 5 TLS Full Handshake Protocol of MODBUS/TCP
Security Protocol Specification [1].

This patch implements the feature within both sync and async_io version.

* Server side:
Add an optional argument "reqclicert" of StartTlsServer(). So, users can
force server require client's certificate for TLS full handshake, or
according to the SSL Context's original behavior [2].

* Client side:
Add optional arguments "certfile" and "keyfile" for replying, if the
server requires client's certificate for TLS full handshake.

Besides, also add an optional argument "password" on both server and
client side for decrypting the private keyfile.

This fixes part of https://github.com/riptideio/pymodbus/issues/606

[1]: http://modbus.org/docs/MB-TCP-Security-v21_2018-07-24.pdf
[2]: https://docs.python.org/3/library/ssl.html#ssl.SSLContext.verify_mode

### DIFF
--- a/examples/common/asyncio_server.py
+++ b/examples/common/asyncio_server.py
@@ -131,7 +131,13 @@ async def run_server():
 
     # Tls:
     # await StartTlsServer(context, identity=identity, address=("localhost", 8020),
-    #                      certfile="server.crt", keyfile="server.key",
+    #                      certfile="server.crt", keyfile="server.key", password="pwd",
+    #                      allow_reuse_address=True, allow_reuse_port=True,
+    #                      defer_start=False)
+
+    # Tls and force require client's certificate for TLS full handshake:
+    # await StartTlsServer(context, identity=identity, address=("localhost", 8020),
+    #                      certfile="server.crt", keyfile="server.key", password="pwd", reqclicert=True,
     #                      allow_reuse_address=True, allow_reuse_port=True,
     #                      defer_start=False)
 
--- a/examples/common/synchronous_server.py
+++ b/examples/common/synchronous_server.py
@@ -120,8 +120,14 @@ def run_server():
     #                framer=ModbusRtuFramer, address=("0.0.0.0", 5020))
 
     # TLS
-    # StartTlsServer(context, identity=identity, certfile="server.crt",
-    #                keyfile="server.key", address=("0.0.0.0", 8020))
+    # StartTlsServer(context, identity=identity,
+    #                certfile="server.crt", keyfile="server.key", password="pwd",
+    #                address=("0.0.0.0", 8020))
+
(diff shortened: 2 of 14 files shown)


starnight:
Cool!  There is the Code Analysis for security!!!

It shows `ssl.PROTOCOL_TLS` is insecure at [here](https://github.com/riptideio/pymodbus/pull/615/files#diff-73d8fe18c2b9a4ce5490305b39953aecd7c61d751fab6a0c7fbd2ea85f094636R21) for example.
The description:
```
Weak SSL/TLS protocols should not be used

Vulnerability
Major
Main sources

    cwe, owasp-a3, owasp-a6, privacy, sans-top25-porous
    Available SinceSep 17, 2019
    SonarQube (Python)
    Constant/issue: 2min

Older versions of SSL/TLS protocol like "SSLv3" have been proven to be insecure.

This rule raises an issue when an SSL/TLS context is created with an insecure protocol version, i.e. when one of the following constants is detected in the code:

    OpenSSL.SSL.SSLv3_METHOD (Use instead OpenSSL.SSL.TLSv1_2_METHOD)
    ssl.PROTOCOL_SSLv3 (Use instead ssl.PROTOCOL_TLSv1_2) 

Protocol versions different from TLSv1.2 and TLSv1.3 are considered insecure.
```
However, the older SSL/TLS protocols are disabled by the following way in the [code](https://github.com/riptideio/pymodbus/pull/615/files#diff-73d8fe18c2b9a4ce5490305b39953aecd7c61d751fab6a0c7fbd2ea85f094636R22):
```
        # According to MODBUS/TCP Security Protocol Specification, it is
        # TLSv2 at least
        sslctx.options |= ssl.OP_NO_TLSv1_1
        sslctx.options |= ssl.OP_NO_TLSv1
        sslctx.options |= ssl.OP_NO_SSLv3
        sslctx.options |= ssl.OP_NO_SSLv2
```
So, I think this is okay.

starnight:
Since there is no `PROTOCOL_TLSv1_3` in Python's SSL library, and  [`PROTOCOL_TLSv1_2`](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLSv1_2) is the latest one for now.  To avoid the warning message from the Code Analysis, I changed all of the `PROTOCOL_TLS` to `PROTOCOL_TLSv1_2`.

starnight:
Any other things I can help for this PR?

dhoomakethu:

PR #637, opened 2021-05-26
select proper bitness from pypy externals for windows

altendky:
PyPy went to 64 bit between 7.3.3 and 7.3.4 which happened between the last good build riptideio/pymodbus@de0cc3e and the first bad one at riptideio/pymodbus@a22932e.  We need to update to use the 64-bit PyPy externals reference to get the needed libraries for building cryptography.  Also, 7.3.4 dropped Python 3.6 support so our 3.6 tests will continue with 32-bit PyPy 7.3.3.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,7 +158,7 @@ jobs:
           mkdir --parents $(dirname $PYPY_EXTERNALS_PATH)
           hg clone https://foss.heptapod.net/pypy/externals/ $PYPY_EXTERNALS_PATH
           dir $PYPY_EXTERNALS_PATH
-          cd $PYPY_EXTERNALS_PATH && hg update win$(pypy -c 'import struct; print(struct.calcsize("P") * 8)')_14x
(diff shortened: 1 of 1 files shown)


altendky:
@dhoomakethu, here's a Windows PyPy fix up.

dhoomakethu:
@altendky its still failing for windows 3.7x.

altendky:
It is.  I should probably take another pass at clearing that flakiness, but it is not related to the PyPy fix implemented here.

dhoomakethu:
@altendky I was talking about this job https://github.com/riptideio/pymodbus/pull/637/checks?check_run_id=2670955558

altendky:
@dhoomakethu, yes, that is CPython and thus unrelated to the new PyPy release being 64-bit.  This PR fixes the PyPy issue.  I would recommend merging this fix and addressing the flaky asyncio test setup separately.  I'll see if I can find some time today to look at the flaky stuff.

altendky:
For linkage, see https://github.com/riptideio/pymodbus/pull/639 for work on the flaky CPython asyncio test.

PR #639, opened 2021-06-01
improve flaky asyncio test

altendky:


altendky:
@dhoomakethu, I had to cut the sleep time down to `0.01` locally to get the error but it seems like it  may just be these tests where 'we' knew there were just sleeps that were hopefully long enough.  We don't really want to just make them all longer and have tests take longer all the time, but we do want to give them more time.  The new one second on that particular test may or may not be 'sufficient' and other tests may need it as well.  If you want I could make this functionality (sleep with occasional check) a function that takes a callable so we can reuse it in each of the non-zero sleep locations.

altendky:
Kick CI to test for remaining flakiness

altendky:
Kick CI to test for remaining flakiness

altendky:
Kick CI to test for remaining flakiness

altendky:
Welp, there's a few runs anyways without any (related) failures.

PR #643, opened 2021-06-10
Lazy Load Serial Module Instead of Requiring it For Asyncio

StephenNorCalControls:
Fixes an issue where using any aspect of the async_io server module requires having a serial module installed, even to use things like the TCP Server.


dhoomakethu:

PR #653, opened 2021-07-25
sync/ModbusTlsServer: Fix ModbusTlsServer's default framer

starnight:
ModbusTlsServer's framer is set as ModbusTlsFramer by caller
StartTlsServer. However, if developers invoke ModbusTlsServer directly
without assigned framer, the framer will be assigned as
ModbusSocketFramer. Because, ModbusTlsServer inherits from
ModbusTcpServer. This parses payload failed for ModbusTlsServer.
This patch assigns ModbusTlsServer's default framer as ModbusTlsFramer.
And, adds corresponding unittest. (Add sync/ModbusTlsClient's framer
test as well)

Fixes #644

PR #652, opened 2021-07-18
add python 3.10 tests

Emilv2:
Also test python 3.10 since it will be released soon.

PR #650, opened 2021-07-04
Added missing support for broadcast messages on Asyncio client.

NevoSolar:

PR #658, opened 2021-08-01
3.0.0

dhoomakethu:
This PR contains all the existing PR's intended for 3.x branch. 
This PR will be held open against the other PR's

Closes

- #653 
- #556 
- #558 
- #604
- #615 
- #643 
- #650 
- #652 
- #653


dhoomakethu:
If you think some of the changes from the PR's are not merged correctly or have been overidden by subsequent merges, please raise a new PR against `3.0.0` branch

dhoomakethu:
@starnight please take a look in to this PR

starnight:
Have a PR #661 to fix the conflicts.  @dhoomakethu please take a look.

pymodbus/client/asynchronous/factory/serial.py
```diff
@@ -103,11 +108,12 @@ def async_io_factory(port=None, framer=None, **kwargs):
 
     client = AsyncioModbusSerialClient(port, proto_cls, framer, loop, **kwargs)
     coro = client.connect()
```
kellerza, line 110: This coroutine is never awaited if the following is true: loop.is_running() and loop is asyncio.get_event_loop()

kellerza: Proposed fixes if you create from within another coroutine,- option 3 seems the cleanest!

### option 1 
inside the factory, create the task and return
```
   coro = client.connect()
   if loop.is_running() and loop is asyncio.get_event_loop():
        asyncio.create_task(coro)
        return loop, client
```
the the caller should just pause its execution to allow the connect schedule on the event loop to complete

```
async def main():
    _loop, client = AsyncSerialClient(...)
    asyncio.sleep(0)  # this allows this excution to wait 
```

### option 2
from the factory, return the coro and client (instead of the loop - here the signature is different)
```
   coro = client.connect()
   if loop.is_running() and loop is asyncio.get_event_loop():
        return coro, client
```
the the caller you can then await the coro

```
async def main():
    coro, client = AsyncSerialClient(...)
    await coro
```

### option 3
inside the factory, dont create the coro
```
    # coro = client.connect()
    if loop.is_running() and loop is asyncio.get_event_loop():
        return coro, client
   
    coro = client.connect()
```
the the caller you can then await client.connect()

```
async def main():
    loop, client = AsyncSerialClient(...)
    await client.connect()
```

dhoomakethu: The problem with `asyncio` code is not most of the users will be knowing how to use it and end up creating more tickets. I will see what could be done. Thanks for the suggestion.

kellerza: Indeed, that makes option 1&2 above rather cryptic.

My suggestion would be NOT to connect during the creation, but always leave it to the user to call `client.connect()`

Then they can do either one of the following, depending on where they start it from

```
loop, client = AsyncSerialClient(...)

# Option a - Loop has yet to start
asyncio.run_until_complete(client.connect())

# Option b - if the loop is already running and they are inside a coroutine/async def function
await client.connect()

# Option c - from python 3.7 replaces the ensure_future code currently there..
# although this works from a running coroutine, you might have to await it, so awaiting the client.connect (option a) is better)
asyncio.create_task(client.connect())

kellerza: I think https://github.com/riptideio/pymodbus/pull/558/files was an attempt to fix this for TCP. In that case they also left client.connect() to the caller, but this was not clearly mentioned in the PR

Also, you really want to same interface, irrespective of where you call things from. So believe splitting out the connect is the way to go. Let me know if you want me to do a PR in the 3.0.0 branch (its not clear when/if you will merge this PR?)

dhoomakethu: That's a good question, it was supposed to be merged and available a long time ago but due to time constraints and the pandemic this is getting nowhere. I have no idea of the timeline but I am hopeful to get a 3.x release by the end of year.

PR #662, opened 2021-08-10
Treat exception codes as valid responses

christiansandberg:
When an exception response is received the function code is different from the request but the frame should still be considered valid.

### DIFF
--- a/pymodbus/transaction.py
+++ b/pymodbus/transaction.py
@@ -123,7 +123,7 @@ class ModbusTransactionManager(object):
             return False
 
         mbap = self.client.framer.decode_data(response)
-        if mbap.get('unit') != request.unit_id or mbap.get('fcode') != request.function_code:
+        if mbap.get('unit') != request.unit_id or mbap.get('fcode') & 0x7F != request.function_code:
             return False
 
         if 'length' in mbap and exp_resp_len:


pymodbus/transaction.py
```diff
@@ -123,7 +123,7 @@ def _validate_response(self, request, response, exp_resp_len):
             return False
 
         mbap = self.client.framer.decode_data(response)
-        if mbap.get('unit') != request.unit_id or mbap.get('fcode') != request.function_code:
+        if mbap.get('unit') != request.unit_id or mbap.get('fcode') & 0x7F != request.function_code:
```
dhoomakethu, line 126: what is the rational behind comparing `function_code` with 127 (0x7F) ? From the modbus spec
```The exception response message has two fields that differentiate it from a normal response:
Function Code Field: In a normal response, the server echoes the function code of the
original request in the function code field of the response. All function codes have a most–
significant bit (MSB) of 0 (their values are all below 80 hexadecimal). In an exception
response, the server sets the MSB of the function code to 1. This makes the function code
value in an exception```

The title of the PR is a little confusing as well. Would you please mind clarifying if you are trying to handle the function code in the exception response or the exception code (valid range of which is between (0-0x0B)?

christiansandberg: Sorry for the vague description. I'll try to explain with an example.

If the request uses function code 0x03, but the server responds with an exception, it will set the MSB to 1 in the function code, so the response function code will be 0x83. This is however a valid response, so when checking the response we must accept both 0x03 and 0x83. One way to do this is to mask out the MSB. 0x83 & 0x7F == 0x03.

dhoomakethu: ah, sorry, I confused `&` with logical `and` , side effects of working on multiple languages :D  . Now it's clear. Thanks for the clarification.

dhoomakethu:
@christiansandberg I will cherry pick this to `2.5.3` branch and will close this PR.

PR #665, opened 2021-09-05
Fix async unit

### LINKED ISSUE #624: Unit ID not handled properly in async clients
@dhoomakethu, sorry for long response. Below is the log file of failing unit id validation.
```
2021-03-25 23:47:13,091 MainThread      DEBUG    selector_events:53       Using selector: SelectSelector
2021-03-25 23:47:13,097 MainThread      DEBUG    __init__       :789      Connecting.
2021-03-25 23:47:13,106 MainThread      DEBUG    __init__       :106      Client connected to modbus server
2021-03-25 23:47:13,106 MainThread      INFO     __init__       :806      Protocol made connection.
2021-03-25 23:47:13,106 MainThread      INFO     __init__       :798      Connected to COM57
2021-03-25 23:47:13,107 MainThread      DEBUG    mdbs_master    :27       Read input registers
2021-03-25 23:47:13,107 MainThread      DEBUG    __init__       :140      send: 0x1 0x4 0x0 0x1 0x0 0xe 0x20 0xe
2021-03-25 23:47:13,107 MainThread      DEBUG    transaction    :448      Adding transaction 1
2021-03-25 23:47:13,148 MainThread      DEBUG    __init__       :149      recv: 0x1 0x4 0x1c 0x55 0x55 0x55
2021-03-25 23:47:13,149 MainThread      DEBUG    rtu_framer     :240      Frame - [b'\x01\x04\x1cUUU'] not ready
2021-03-25 23:47:13,164 MainThread      DEBUG    __init__       :149      recv: 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55
2021-03-25 23:47:13,164 MainThread      DEBUG    rtu_framer     :240      Frame - [b'UUUUUUUU'] not ready
2021-03-25 23:47:13,181 MainThread      DEBUG    __init__       :149      recv: 0x55 0x55 0x55 0x55 0x55 0x55 0x55
2021-03-25 23:47:13,181 MainThread      DEBUG    rtu_framer     :240      Frame - [b'UUUUUUU'] not ready
2021-03-25 23:47:13,196 MainThread      DEBUG    __init__       :149      recv: 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55
2021-03-25 23:47:13,196 MainThread      DEBUG    rtu_framer     :240      Frame - [b'UUUUUUUU'] not ready
2021-03-25 23:47:13,213 MainThread      DEBUG    __init__       :149      recv: 0x55 0x55 0xf3 0xbf
2021-03-25 23:47:27,496 MainThread      DEBUG    rtu_framer     :234      Not a valid unit id - 1, ignoring!!
2021-03-25 23:47:32,495 MainThread      DEBUG    rtu_framer     :120      Resetting frame - Current Frame in buffer - 0x1 0x4 0x1c 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0x55 0xf3 0xbf

```
If in the line 230 in pymodbus/framer/rtu_framer.py remove unit validation:
```
                if True: #if self._validate_unit_id(unit, single):
```
messages will start to receive correctly.
The reason, as I said, in pymodbus/client/asynchronous/async_io/__init__.py code:
```
        # here decode_data assigns unit to the first byte of last part of message, 0x55 in the example above
        unit = self.framer.decode_data(data).get("unit", 0)  
        self.framer.processIncomingPacket(data, self._handleResponse, unit=unit)
```

_Originally posted by @olddudealex in https://github.com/riptideio/pymodbus/issues/608#issuecomment-807501686_

mdavidsaver:
An attempt to fix #624, at least for asyncio.  I'm not certain how unit ID should be handled here.  Can a client have requests to multiple unit IDs in-flight?  I've tried to account for this possibility by storing all unit IDs requested since the last (re)connect.  If this isn't necessary, a simpler solution would be to only remember the most recently sent unit ID.

If this is seen as a reasonable approach, I will expand to the other async backends (although I'm only using/testing with asyncio).

### DIFF
--- a/pymodbus/client/asynchronous/async_io/__init__.py
+++ b/pymodbus/client/asynchronous/async_io/__init__.py
@@ -45,6 +45,7 @@ class BaseModbusAsyncClientProtocol(AsyncModbusClientMixin):
         """
         self.transport = transport
         self._connectionMade()
+        self.__units_send = set()
 
         if self.factory:
             self.factory.protocol_made_connection(self)
@@ -135,9 +136,10 @@ class BaseModbusAsyncClientProtocol(AsyncModbusClientMixin):
         Starts the producer to send the next request to
         consumer.write(Frame(request))
         """
+        self.__units_send.add(request.unit_id)
         request.transaction_id = self.transaction.getNextTID()
         packet = self.framer.buildPacket(request)
-        _logger.debug("send: " + hexlify_packets(packet))
+        _logger.debug("send: u%d " + hexlify_packets(packet), request.unit_id)
         self.write_transport(packet)
         return self._buildResponse(request.transaction_id)
 
@@ -146,9 +148,8 @@ class BaseModbusAsyncClientProtocol(AsyncModbusClientMixin):
 
         :param data: The data returned from the server
         '''
-        _logger.debug("recv: " + hexlify_packets(data))
-        unit = self.framer.decode_data(data).get("unit", 0)
-        self.framer.processIncomingPacket(data, self._handleResponse, unit=unit)
+        _logger.debug("recv: %r " + hexlify_packets(data), self.__units_send)
+        self.framer.processIncomingPacket(data, self._handleResponse, unit=list(self.__units_send))
 
     def _handleResponse(self, reply, **kwargs):
         """
--- a/pymodbus/framer/rtu_framer.py
+++ b/pymodbus/framer/rtu_framer.py
@@ -230,8 +230,8 @@ class ModbusRtuFramer(ModbusFramer):
                 if self._validate_unit_id(unit, single):
                     self._process(callback)
                 else:
-                    _logger.debug("Not a valid unit id - {}, "
-                                  "ignoring!!".format(self._header['uid']))
+                    _logger.debug("Not a valid unit id - {} not in {}, "
+                                  "ignoring!!".format(self._header['uid'], unit))
                     self.resetFrame()
             else:
                 _logger.debug("Frame check failed, ignoring!!")


dhoomakethu:
@mdavidsaver thanks for the PR and apologies for the delay. I am going to keep this in review for some more time. I do not have a strong opinion for storing the unit id's . This is something that a modbus application should handle which uses pymodbus as a driver and not pymodbus itself. Please correct me if I am not understanding this PR properly.

janiversen:
I think there are a bigger misunderstanding here (Or it is in my head). UnitId in tcp is the same as slaveid in serial, so why do we want to keep track of that ???

I believe #624 is true but a whole different problem.

janiversen:
According to the modbus protocol there can only be one request "in flight" at one connection. Please unitId addresses the slave (device) on the open connection, so we have to wait until a response is received before continuing.

We should try with a simpler solution. I can see that unitId is missing is some calls (at least in the documentation).

tracernz:
> According to the modbus protocol there can only be one request "in flight" at one connection

Is this still valid for Modbus TCP? That protocol specification equips the server to have multiple active transactions, and specifies "Several MODBUS transactions can be activated simultaneously on the same TCP Connection." It also includes the "Transaction Identifier" field in the MBAP header which would seem to facilitate this.

P.S. Great to see all the work you're doing lately 👍 .

janiversen:
The server must be able to handle multiple transactions because it can have multiple clients connected, that is true for a number of the transport protocols. The server do not need to store the transactions since it is part of the request, which are used to form the response.

A client however connect to 1 server with a socket (tcp etc.) or a physical line, and can therefore only have one transaction outstanding. A client can support multiple devices using multiple sockets or multiple serial lines, but in that case each connection can use the same transaction. In our case 1 instance of modbusClient can only handle 1 connection so no need to store transactions.

janiversen:
thanks for the nice words.

tracernz:
These are the key passages from the "MODBUS Messaging on TCP/IP Implementation Guide V1.0b" I refer to:
image
image
image

A limit of 1 is no problem; the spec allows this, so I guess it's fine 😀 . In any case, I think this is a much bigger topic than the original issue referenced here, and the more I read the spec the more vague the descriptions appear 😆 .

janiversen:
Your reference is correct, but you are referencing a gateway which is more correctly named a forwarder, this is a different topic, because that can e.g. have 1 server (tcp) and multiple clients (serial).

I have yet to see a end-device that allows multiple requests, nor do I understand the need for that....think about it send the following:
```
read_register()
write_register()
```
the device (server) might decide to do the write first...because there are no demand that transactions ids are handled in ascending sequence.

But for a forwarder the following example makes a lot of sense:
```
read_register(slave=1)
write_register(slave=2)
read_register(slave=3)
```
and the forwarder must be able to handle that.

We have hard coded limit=1, and no intention of changing that for the above reasons.

The description is not vague, you just need to see it together with the different device types.

PR #668, opened 2021-09-29
Replace distutils which is deprecated in 3.10

alexrudd2:
See [PEP632](https://www.python.org/dev/peps/pep-0632/).  

Replaced uses of `distutils.version` with `pkg_resources.parse_version` per their recommendation.  Replace `distutils.core.Command` with `setuptools.Command`.  Was unsure of a suitable replacement for `disutils.log`

Fix several hundred lint warnings from running `flake8` against `\examples`


examples/gui/gtk/simulator.py
```diff
@@ -124,34 +128,7 @@ def run(self):
         """ Run the network reset """
         os.system("/etc/init.d/networking restart")
 
-# --------------------------------------------------------------------------- #
```
alexrudd2, line 127: GTK 2.12 is ancient

dhoomakethu:
Thanks @alexrudd2 , this includes a lot of changes and will take some time before I hit the merge button.

alexrudd2:
> Thanks @alexrudd2 , this includes a lot of changes and will take some time before I hit the merge button.

I'll split the linting into a separate PR.

PR #683, opened 2021-11-13
Use assertEqual instead of assertEquals for Python 3.11 compatibility.

tirkarthi:
The deprecated aliases have been removed in python/cpython#28268 .

### DIFF
--- a/examples/contrib/message_parser.py
+++ b/examples/contrib/message_parser.py
@@ -105,7 +105,7 @@ class Decoder:
                         "  %-12s => %s" % (k_item, v_item)  # pylint: disable=consider-using-f-string
                     )
 
-            elif isinstance(v_dict, collections.Iterable):  # pylint: disable=no-member
+            elif isinstance(v_dict, collections.Iterable):
                 print("%-15s =" % k_dict)  # pylint: disable=consider-using-f-string
                 value = str([int(x) for x in v_dict])
                 for line in textwrap.wrap(value, 60):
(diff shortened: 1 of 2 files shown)


dhoomakethu:
This would break the tests in older python versions (3.7 and above). At present `3.x` branch supports only till 3.9, there is work to support 3.10. This will be in backseat for sometime.

tirkarthi:
The assertEqual is present in both Python 2 and 3. The PR to cpython only removes an alias.

janiversen:
I rebased the PR and added a commit that convert assertEqual(x,y) to assert x == y, which is version independent.

dhoomakethu:
Please hold this PR, we should still be using `self.assertEqual` or whatever the assertion helpers `unittest.TestCase` class provides . These helpers provides more inbuilt way of representing an error when a test case fails.

dhoomakethu:
## Test case
```
import pytest
import unittest


class Tests(unittest.TestCase):
    def test_assert_equal(self):
        self.assertEqual(1, 2, "no way")

    def test_assert(self):
        assert 1 == 2, "no way"

```

## Output without traceback.
**NOTE** the short summary with `assertEqual`
```
(pymodbus) ➜  pymodbus git:(dev) ✗ pytest --tb=no -q -s test/test_assert.py
FF
================================================== short test summary info ===================================================
FAILED test/test_assert.py::Tests::test_assert - AssertionError: no way
FAILED test/test_assert.py::Tests::test_assert_equal - AssertionError: 1 != 2 : no way
2 failed in 0.03s
```

swamper123:
More Info is better than no info, indeed.

janiversen:
But then we change the ton of "assert x == y" to assertEqual(s), so that we at least are consistent.

I have worked with both and do not really have preferences.

PR #693, opened 2021-12-15
Add support for byte strings in the device information fields

sytsereitsma:
I'm trying to emulate a modbus device that has non ASCII data in the vendor name part of the device information. The library cannot correctly encode this, because when encoding this info it is encoded as utf-8. For example, a vendor name of "\xF0\x12\x34" ends up as b"\xC3\xB0\x12\x34" in the encoded payload.

This PR changes the encoding of the device information fields to support both ```bytes``` objects and ```str``` objects.

The appropriate unit tests have been added.

### DIFF
--- a/pymodbus/other_message.py
+++ b/pymodbus/other_message.py
@@ -373,7 +373,7 @@ class ReportSlaveIdRequest(ModbusRequest):
             # Support identity values as bytes data and regular str data
             id_data = []
             for v in information.values():
-                if type(v) is bytes:
+                if isinstance(v, bytes):
                     id_data.append(v)
                 else:
                     id_data.append(v.encode())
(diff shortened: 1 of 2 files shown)


pymodbus/other_message.py
```diff
@@ -369,7 +369,16 @@ def execute(self, context=None):
             reportSlaveIdData = getattr(context, 'reportSlaveIdData', None)
         if not reportSlaveIdData:
             information = DeviceInformationFactory.get(_MCB)
-            identifier = "-".join(information.values()).encode()
+
+            # Support identity values as bytes data and regular str data
+            id_data = []
+            for v in information.values():
+                if type(v) is bytes:
```
dhoomakethu, line 376: Can we just use `isinstance(b, bytes)` instead ? Since we are going to support only 3.x going forward.

sytsereitsma: @dhoomakethu Any chance this will make it to master?

Issue #688, opened 2021-12-01
Partial reads cause RTU framer to discard whole frame with asyncio


marcosatti:
### Versions

* Python: 3.10
* OS: Windows 10
* Pymodbus: 2.5.3
* Modbus Hardware (if used):  USB RS-485 adapter to custom hardware

### Pymodbus Specific
* Client: rtu - async

### Description

Using the asyncio client, partial reads will cause the RTU framer to discard the whole incoming packet, complaining about the unit ID not being valid:

(Logs have been scrubbed of sensitive data.)

```
2021-12-01 09:11:28,424 - pymodbus.client.asynchronous.async_io - DEBUG - send: 0x1 [...]
2021-12-01 09:11:28,424 - pymodbus.transaction - DEBUG - Adding transaction 1
2021-12-01 09:11:28,456 - pymodbus.client.asynchronous.async_io - DEBUG - recv: 0x1 [...]
2021-12-01 09:11:28,456 - pymodbus.framer.rtu_framer - DEBUG - Frame - [b'\x01 ...'] not ready
2021-12-01 09:11:28,471 - pymodbus.client.asynchronous.async_io - DEBUG - recv: 0x65 [...]
2021-12-01 09:11:28,471 - pymodbus.framer.rtu_framer - DEBUG - Frame - [b'e ...'] not ready
2021-12-01 09:11:28,499 - pymodbus.client.asynchronous.async_io - DEBUG - recv: 0x6c [...]
2021-12-01 09:11:28,499 - pymodbus.framer.rtu_framer - DEBUG - Frame - [b'l ...'] not ready
2021-12-01 09:11:28,514 - pymodbus.client.asynchronous.async_io - DEBUG - recv: 0xfe [...]
2021-12-01 09:11:28,514 - pymodbus.framer.rtu_framer - DEBUG - Not a valid unit id - 1, ignoring!!
2021-12-01 09:11:28,514 - pymodbus.framer.rtu_framer - DEBUG - Resetting frame - Current Frame in buffer - 0x1 [...] 0xfb 0x7a
```

The problem is a combination of the following code:

`client/asynchronous/async_io/__init__.py`:
```python
    def _dataReceived(self, data):
        _logger.debug("recv: " + hexlify_packets(data))
        unit = self.framer.decode_data(data).get("unit", 0)
        self.framer.processIncomingPacket(data, self._handleResponse, unit=unit)
```

`framer/rtu_framer.py`:
```python
[...]

PR #707, opened 2022-01-31
Update rtu_framer.py

rckoepke:
Apparent fix for https://github.com/riptideio/pymodbus/issues/688

This was only tested on asyncio serial, not sure whether it will regress other use cases.


rckoepke:
This should help anyone who is having the exact issue in #688. Happy to help any way I can if this needs additional thought, testing, or development before it can actually be merged. Thank you so much for this software.

janiversen:
It breaks our tests, because illegal unit id are no longer caught.

Issue #708, opened 2022-02-01
ReadFileRecordRequest.execute() doesn't wait for complete pdu


theobarkerh:
### Versions

* Python: 3.8.10
* OS: Ubuntu 20.04, Windows 10
* Pymodbus: 2.3.5
* Modbus Hardware (if used): Hach Company reference implementation - slave

### Pymodbus Specific
* Server: none
* Client: /rtu - sync

### Description

When trying to read a file using the ReadFileRecordRequest (Function code 20), the pymodbus.transaction transitions from WAITING FOR REPLY to PROCESSING REPLY before all of the response has been received. This results in the client application receiving many "Modbus Error: [Input/Output] No Response received from the remote unit/Unable to decode response". The application must continue the retry roulette until it receives a complete record response.

One problem is that there is no get_response_pdu_size() member defined for ReadFileRecordRequest. 

### Code and Logs

```python

             rec = FileRecord(reference_type=self.REFERENCE_TYPE, file_number=file_number, record_number=(i*RECORD_LENGTH), record_length=rec_len)
            records.append(rec)

            request = ReadFileRecordRequest(records=records, unit=self.slave_unit)
            response = self.client.execute(request)


DEBUG:pymodbus.transaction:SEND: 0x1 0x14 0x7 0x6 0x0 0x18 0x0 0x0 0x0 0x1a 0x58 0xed
DEBUG:pymodbus.framer.rtu_framer:Changing state to IDLE - Last Frame End - None, Current Time stamp - 1643687749.313818
DEBUG:pymodbus.client.sync:New Transaction state 'SENDING'
DEBUG:pymodbus.transaction:Changing transaction state from 'SENDING' to 'WAITING FOR REPLY'
DEBUG:pymodbus.transaction:Changing transaction state from 'WAITING FOR REPLY' to 'PROCESSING REPLY'
DEBUG:pymodbus.transaction:RECV: 0x5e 0x20 0x8e 0x69 0xc2 0x29 0xb2 0x94 0xbf 0xcc 0x5b 0xf6 0x60 0x1e 0x51 0x5b 0xc2 0x5 0xc5 0xba 0x40 0x0
DEBUG:pymodbus.framer.rtu_framer:Frame check failed, ignoring!!
[...]

PR #709, opened 2022-02-01
708 Implement ReadFileRecordRequest.get_response_pdu_size() to ensure complete record

theobarkerh:
Possible fix to #708 

ReadFileRecordRequest does not include an implementation of get_response_pdu_size() so the rtu_framer completes prior to a full response received. Causes many errors during a file read operation.

### DIFF
--- a/pymodbus/file_message.py
+++ b/pymodbus/file_message.py
@@ -144,7 +144,7 @@ class ReadFileRecordRequest(ModbusRequest):
         :return: 
         """
         for record in self.records:
-            self.count += record.record_length
+            self.count+=record.record_length
 
         return 4 + self.count
 
--- a/pymodbus/file_message.py
+++ b/pymodbus/file_message.py
@@ -140,13 +140,16 @@ class ReadFileRecordRequest(ModbusRequest):
     
     def get_response_pdu_size(self):
         """
-        Func_code (1 byte) + Byte Count(1 byte) + Byte Count2 (1 byte) + Reference Type (1 byte) + registers to read for file
+        Func_code (1 byte) + Response Byte Count(1 byte) +
+            N * (File Response Byte Count (1 byte) + Reference Type (1 byte) + 2 * registers to read for file record)
+            where N = record sub-request count.
         :return: 
         """
+        self.count = 0
         for record in self.records:
-            self.count+=record.record_length
+            self.count+= record.record_length * 2 + 2 if record.record_length else 0
 
-        return 4 + self.count
+        return 2 + self.count
 
 class ReadFileRecordResponse(ModbusResponse):
     """Read file record response.
--- a/pymodbus/file_message.py
+++ b/pymodbus/file_message.py
@@ -90,7 +90,8 @@ class ReadFileRecordRequest(ModbusRequest):
         :param records: The file record requests to be read
         """
         ModbusRequest.__init__(self, **kwargs)
-        self.records = records or []
+        self.records  = records or []
+        self.count = 0
 
     def encode(self):
         """Encode the request packet.
@@ -136,7 +137,16 @@ class ReadFileRecordRequest(ModbusRequest):
         # is too big, return an error.
         files = []
         return ReadFileRecordResponse(files)
+    
+    def get_response_pdu_size(self):
+        """
+        Func_code (1 byte) + Byte Count(1 byte) + Byte Count2 (1 byte) + Reference Type (1 byte) + registers per record
+        :return: 
+        """
+        for record in self.records:
+            self.count += record.record_length
 
+        return 4 + self.count
 
 class ReadFileRecordResponse(ModbusResponse):
     """Read file record response.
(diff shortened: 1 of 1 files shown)


pymodbus/file_message.py
```diff
@@ -128,7 +129,16 @@ def execute(self, context):
         # is too big, return an error.
         files = []
         return ReadFileRecordResponse(files)
+    
+    def get_response_pdu_size(self):
+        """
+        Func_code (1 byte) + Byte Count(1 byte) + Byte Count2 (1 byte) + Reference Type (1 byte) + registers to read for file
```
theobarkerh, line 135: Per https://modbus.org/docs/Modbus_Application_Protocol_V1_1b3.pdf page 32, this should change to:
`Func_code (1 byte) + Response Byte Count(1 byte) + N * (File Response Byte Count (1 byte) + Reference Type (1 byte) + registers to read for file record) where N = record sub-request count. `

pymodbus/file_message.py
```diff
@@ -128,7 +129,16 @@ def execute(self, context):
...
+        :return: 
+        """
+        for record in self.records:
+            self.count+=record.record_length
 
+        return 4 + self.count
```
theobarkerh, line 141: This needs to change per https://modbus.org/docs/Modbus_Application_Protocol_V1_1b3.pdf.

dhoomakethu:
I haven't used the `ReadFileRecordRequest` so just dropping the spec for future reference
image

theobarkerh:
Suggestions on how to resolve??

christiansandberg:
I really think pymodbus should use the response data length (byte after the function code) instead to determine how many bytes to read next. I don't think you should rely on the server returning the requested length because some files may behave like in a file system where reading past EOF will return less bytes than requested.

theobarkerh:
> I really think pymodbus should use the response data length (byte after the function code) instead to determine how many bytes to read next. I don't think you should rely on the server returning the requested length because some files may behave like in a file system where reading past EOF will return less bytes than requested.

Have you tried transferring files from devices produced by multiple vendors?

christiansandberg:
This change will for sure break compatibility with certain devices. The method used currently, although unstable, is correct since it doesn't make assumptions about the response length based on the request. I'm guessing this was the reason why this method is missing from file records in the first place.

Using the byte count instead will not change the current behavior, as long as the byte count given by the device is correct of course.

theobarkerh:
> This change will for sure break compatibility with certain devices. The method used currently, although unstable, is correct since it doesn't make assumptions about the response length based on the request. I'm guessing this was the reason why this method is missing from file records in the first place.
> 
> Using the byte count instead will not change the current behavior, as long as the byte count given by the device is correct of course.

The current method is unusable. Thus the PR. 
This was the best solution I found based on my limited knowledge of the code base and specification. Please add code changes you would suggest to the diff on this PR.

janiversen:
@theobarkerh can you please test against the current dev, there have been multiple corrections, that leads me to believe that your PR might no longer be relevant, however you are the best to say so.

theobarkerh:
> @theobarkerh can you please test against the current dev, there have been multiple corrections, that leads me to believe that your PR might no longer be relevant, however you are the best to say so.

Thanks Jan! I'll test it as soon as a get a break to do that. Hopefully today or tomorrow.

theobarkerh:
@janiversen, I was able to test with the code which tripped over the implementation in the last release. 
The 'dev' branch ran much better. I didn't have a chance to instrument it to verify that the timing is optimal.

theobarkerh:
@janiversen, I was able to test with logging.setLevel(logging.DEBUG) and compare this implementation with 'dev'. They do appear to have the same efficiency.

janiversen:
I am confused you talk about efficiency, but your pr is about errors when reading files, I do not see the relation.

I never said that dev was more efficient, but partial reads have been reduced.

theobarkerh:
> I am confused you talk about efficiency, but your pr is about errors when reading files, I do not see the relation.
> 
> I never said that dev was more efficient, but **partial reads have been reduced**.

Partial reads is the efficiency problem to which I referred. Thank you. As you noted this PR may not add any improvement over the current 'dev' implementation.

janiversen:
Ok, but it still makes sense to have a correct get_response_pdu_size()

christiansandberg:
Unfortunately this will cause communication with our devices to fail since the size indication encoded in the response gets ignored. I know that the specification says that there probably should be a relation between request and response but the protocol also has an explicit byte count for the actual response size and this should take priority over any implicit calculations.

One example of a use case where the spec is not always followed is when transferring large files using Write File Record. The response should contain the exact same data as in the request but this wastes a lot of bandwidth and many implementations choose to not send the data back. The Modbus protocol supports this behavior both by indicating the byte size and the underlying framing layer.

In this regard the previous behavior was better since it was more flexible in response sizes it accepted. Now it will cause a timeout on perfectly valid responses.

PR #715, opened 2022-02-17
Asyncio server will set serving=False if it fails to start

couling:
Asyncio server should notify that it is NOT serving if the server fails to start.

At this time, the server has no `server.start_serving()` method only a `server.serve_forever()` method.  This means the only way to detect if the server is ready to receive connections `await server.serving()` with `defer_start=False`.

This is particularly useful when using the modbus server in a unit test.
```python
import asyncio

async def use_server_for_test(context):
    server = ModbusTcpServer(context)
    server_task = asyncio.get_running_loop().create_task(server.serve_forever())
    try:
        assert await server.serving
        # Test a client by connecting to this server.
        test_my_client(server.server.sockets[0].getsockname())
    finally:
        server_task.cancel()
```

At the moment the above code would just hang and I can see no other future that will return both when the server starts and if it fails to start.

This PR then sets the `serving` to `False` if the server fails to startup.

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -483,7 +483,11 @@ class ModbusTcpServer:
 
     async def serve_forever(self):
         if self.server is None:
-            self.server = await self.server_factory
+            try:
+                self.server = await self.server_factory
+            except:
+                self.serving.set_result(False)
+                raise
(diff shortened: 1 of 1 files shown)


janiversen:
This code is merged into dev.

PR #730, opened 2022-03-29
Solve macOS problems in test_async_asyncio.py

janiversen:
The failing tests mentioned in
https://github.com/riptideio/pymodbus/discussions/729
is solved. It was mostly a matter of a missing "async" keyword.

However with this PR it became clear, that there are a loose task (the event loop) that are never closed, that is however for another PR.

PR #731, opened 2022-03-31
Remove not maintained example directories

janiversen:
A major release is a good point in time to remove unsupported code, and by removing it we safe ourself problems later with pylint, black y mypy.

These will still live in the 2.x releases, but are removed for the 3.x
releases.

I have raised this PR against dev3.x, because I work on #671 so we can start 3.0.0 with clean examples.


swamper123:
Tbh I have never inspected these subfolders of example before. ^^
What was the purpose of those? Stuff for Integrationtests?

And what I can't understand (besides the PR), why github actions stuck mostly at macOS PyPy 3.7 tests...and not showing a reason why it was cancelled....

janiversen:
I wonder if this is actually the async problem I fixed in #730 

I detected that some async tests would hang on my mac with python 3.10. This PR fails on windows, but one reason why a job is cancelled is that it is hanging.

dhoomakethu:
This `examples/twisted/plugins/pymodbus_plugin.py` could be a good starting point for supporting pymodbus plugins.

janiversen:
I will make a new issue “make plugin framework for async factories” and add the file there.

janiversen:
new issue https://github.com/riptideio/pymodbus/issues/734

PR #732, opened 2022-03-31
rebase #671, and add pylint setup.

janiversen:
The PR rebases #671 (keeping the original commits with author etc). I had to do it as a new PR, since only maintainers are allowed to update PRs

Having the examples pep8 complaint is a nice thing to have for version 3.0.0

Based on this PR, I am working my way through all directories to ensure pylint declares them "clean", once this PR is merged I will finish and submit those, in the hope they too can make it into v3.0.0

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,5 @@
+[BASIC]
+
+# Good variable names which should always be accepted, separated by a comma
+good-names=rr,rq
+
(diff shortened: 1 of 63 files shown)


dhoomakethu:
@janiversen I started something on similar lines, refer branch https://github.com/riptideio/pymodbus/tree/3.0.0-black

janiversen:
No need for this PR anymore, updating direct instead.

PR #733, opened 2022-03-31
Solve async test problems (this is #730 but rebased also in GitHub)

janiversen:
The failing tests mentioned in
https://github.com/riptideio/pymodbus/discussions/729
is solved. It was mostly a matter of a missing "async" keyword.

However with this PR it became clear, that there are a loose task (the event loop) that are never closed, that is however for another PR.

### DIFF
--- a/test/test_client_async_asyncio.py
+++ b/test/test_client_async_asyncio.py
@@ -278,7 +278,7 @@ class TestAsyncioClient(object):
             assert protocol.factory.protocol_lost_connection.call_count == 1
 
     @pytest.mark.parametrize("protocol", protocols)
-    async def testClientProtocolDataReceived(self, protocol):
+    async def test_client_protocol_data_received(self, protocol):
         ''' Test the client protocol data received '''
         protocol = protocol(ModbusSocketFramer(ClientDecoder()))
         transport = mock.MagicMock()
@@ -319,7 +319,7 @@ class TestAsyncioClient(object):
         assert d == f
 
     @pytest.mark.parametrize("protocol", protocols)
-    async def testClientProtocolHandleResponse(self, protocol):
+    async def test_client_protocol_handle_response(self, protocol):
         ''' Test the client protocol handles responses '''
         protocol = protocol()
         transport = mock.MagicMock()
@@ -340,7 +340,7 @@ class TestAsyncioClient(object):
         assert result == reply
 
     @pytest.mark.parametrize("protocol", protocols)
-    async def testClientProtocolBuildResponse(self, protocol):
+    async def test_client_protocol_build_response(self, protocol):
         ''' Test the udp client protocol builds responses '''
         protocol = protocol()
         # if isinstance(protocol.create_future, mock.MagicMock):
--- a/test/test_client_async_asyncio.py
+++ b/test/test_client_async_asyncio.py
@@ -152,19 +152,22 @@ class TestAsyncioClient(object):
         mock_protocol_class = mock.MagicMock()
         mock_loop = mock.MagicMock()
         client = ReconnectingAsyncioModbusTcpClient(protocol_class=mock_protocol_class, loop=mock_loop)
-
         assert not client.connected
         assert client.protocol is None
+
+
+        # fake client is connected and *then* looses connection:
+        client.connected = True
+        client.host = None
+        client.protocol = mock.sentinel.PROTOCOL
         client.protocol_lost_connection(mock.sentinel.PROTOCOL_UNEXPECTED)
         assert not client.connected
 
-        # fake client ist connected and *then* looses connection:
+
         client.connected = True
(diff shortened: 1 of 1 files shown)


janiversen:
hmmm I had that working on the de3.x branch, let me do a dive.

test/test_client_async_asyncio.py
```diff
@@ -239,7 +242,7 @@ def testClientProtocolConnectionLost(self, protocol):
             assert protocol.factory.protocol_lost_connection.call_count == 1
 
     @pytest.mark.parametrize("protocol", protocols)
-    def testClientProtocolDataReceived(self, protocol):
+    async def test_client_protocol_data_received(self, protocol):
```
swamper123, line 245: I am not into the code deeply atm.
Why does this method (and the other changes to async) not containing an await?
Does it need to be async then? Or did I missed sth?

janiversen: Await can only be used in async functions, but are not needed to be used. Pytest detects it is an async function and does await internally.

In general all functions dealing with async (e.g. calling async functions or yielding) should be declared as async, this replaces the old coroutine decorator.

janiversen: with 3.10 you can experience function hanging if not declared correctly.

janiversen:
hmmm it fails in exactly the same step. tox on macOS (latest) x64. Funny thing is that I work on a Mac with i7 (x64) and latest OS, I cannot get tox to fail (with 3.7)

swamper123:
https://github.com/riptideio/pymodbus/actions/runs/2076917788

grafik

We should take more time. ^^

swamper123:
In the notations we got some more issues as well we should take a look at:

https://pipelines.actions.githubusercontent.com/serviceHosts/3d9d7158-a6a4-49fd-883b-a523e82feb58/_apis/pipelines/1/runs/1055/signedlogcontent/8?urlExpires=2022-04-01T13%3A02%3A34.1932351Z&urlSigningMethod=HMACV1&urlSignature=9D3PQkuTPZ1LpoQCIB%2BaW87JeZ0UgL%2FUlMADvhteIVc%3D

grafik

swamper123:
https://github.com/riptideio/pymodbus/blob/d92516d7cf57d5500c04636510edb417ea6424b4/.github/workflows/ci.yml#L24

Try changing this one to 15 or 20min. On one hand we could check if building is okay and we should observe why it takes so long in general.

dhoomakethu:
Let's move python version on mac osx to python3.8 or later and see if it helps. Not a solution but just a ugly hack if it works.

dhoomakethu:
```
Processing ./.tox/.tmp/package/1/pymodbus-2.5.3.zip
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Requirement already satisfied: six>=1.15.0 in ./.tox/pypy37/site-packages (from pymodbus==2.5.3) (1.16.0)
Requirement already satisfied: pyserial>=3.4 in ./.tox/pypy37/site-packages (from pymodbus==2.5.3) (3.5)
```
interesting to know that the pymodbus version being reported is `2.5.3`

dhoomakethu:
Sorry, the `dev` branch is on `2.5.3`. May be we should move to `dev.3.x`

janiversen:
Please remark the problem is only with the pypy interpreter. 

@swamper123 changing the time does not solve the issue, it is simply hanging but only on GitHub, not my machine.

dhoomakethu:
Same with me, I tested on mac with out any issues, I am worried both `dev` and `dev3.x` are floating around and might cause confusion.

janiversen:
Count me in on the confusion, I used dev3.x until you asked me to use dev. dev3.x have been merged into dev if I am correct, so we can delete dev3.x and continue the main development on dev.

If one day we need to patch 2.5.3 we open a branch based on the tag.

Could you secure that all of dev3.x is in dev ? and then remove dev3.x

dhoomakethu:
Yes , I am doing it now.

janiversen:
Just for test !! I have remove pypy test on macOS, and also updated tox.ini to not do py38, py39.

janiversen:
@swamper123 look at the other run times, all a couple of minutes, no reason why pypy3.7 should take so much longer. But I can give it a try.

janiversen:
@swamper123 keep your eyes on #671 that runs (for now) with timeout=15minutes.

janiversen:
@swamper123 my apologies, you were right !!! timing was the issue, but I still do not understand why this takes so much more.

swamper123:
grafik

It's the building phase. In the logs, you can activate the timestamps.
`cryptograhpy ` is complicated and needs time.

In other versions cryptography is installed before, like in Linux. So it hasn't to be rebuild. 
This seems one of the biggest timekiller in this whole thing.

PR #736, opened 2022-04-01
Remove python 2.7 and 3.6.

janiversen:
Remove CI for 2.7 and 3.6.
Update README.rst to reflect we support python >= 3.7

### DIFF
--- a/doc/INSTALL
+++ b/doc/INSTALL
@@ -1,8 +1,8 @@
 Requirements
 -------------
 
-* Python 2.7 or later.
-* Python Twisted, Tornado or asyncio (For async client and server)
+* Python 3.7 or later.
+* Python asyncio (preferred), Twisted or Tornado (For async client and server)
 * Pyserial
 
 On Windows pywin32 is recommended (this is built in to ActivePython,
--- a/doc/INSTALL
+++ b/doc/INSTALL
@@ -2,7 +2,8 @@ Requirements
 -------------
 
 * Python 3.7 or later.
-* Python asyncio (preferred), Twisted or Tornado (For async client and server)
+* Python Twisted or Tornado (if not using asyncio for async client and server),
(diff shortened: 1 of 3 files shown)


janiversen:
@dhoomakethu please have a look, I think I did it correctly, but for CI I really would like a second opinion.

doc/INSTALL
```diff
@@ -1,8 +1,8 @@
 Requirements
 -------------
 
-* Python 2.7 or later.
-* Python Twisted, Tornado or asyncio (For async client and server)
+* Python 3.7 or later.
+* Python asyncio (preferred), Twisted or Tornado (For async client and server)
```
swamper123, line 5: Python asyncio is builtin, so it may not be a requirement anymore.
For Twisted and Tornado  we should mention that stuff is in maintainance mode at the moment and other solutions (like plugins) may be provided in future or even dropped.

swamper123 (approved):
The rest seems to be fine :)

doc/INSTALL
janiversen: ups, you are quite right, only excuse is that I have not had my coffee yet :-)

Added a commit.

janiversen:
The failing 3.7 test is outside our control, it seems installing dependencies etc. takes too long.

swamper123:
It smells suspicious...in other running actions it got at least to the tests and have been cancelled after ~ 8 min.
This time it was cancelled before the tests .... there is sth. fishy.

swamper123:
I hate such things, where you can't traceback well ... there was a warning in the logs, but this shouldn't be the cause of the "now I want to run, now I cancel"-Action.
From my side this PR is fine and maybe we find the root of all macOS-Pypy-3.7 evil at some time. ^^

janiversen:
And as you can see, just restarting the job did the trick, there are indeed something fishy with the 3.7 tests.

PR #737, opened 2022-04-01
Update tox to only handle supported python versions

janiversen:
Let update tox to be identical to ci.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # directory.
 
 [tox]
-envlist = py{27,py27,36,37,38,39,py36,py37}
+envlist = py{37,38,39,py36,py37}
 
 [testenv]
 deps = -r requirements-tests.txt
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -4,7 +4,7 @@
 # directory.
 
 [tox]
-envlist = py{27,py27,36,37,38,39,py36,py37}
+envlist = py{37,38,39,py36,py37}
```
swamper123, line 7: Guess py36 can be removed as well or better replaced by py38 and py39?

swamper123 (changes requested):

tox.ini
janiversen: nice catch, done.

janiversen: please approve if you have no further comments.

swamper123 (approved):

PR #746, opened 2022-04-03
Add python 3.10 to CI

janiversen:
Let´s test python 3.10 as well.


janiversen:
hmmm seems there are 2 tests failing on windows, I will try to debug this (without having a windows machine).

dhoomakethu (approved):

PR #751, opened 2022-04-04
version 3.10 problem

janiversen:
During testing with 3.10 I found a missing try/except in async_io.py

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -300,7 +300,11 @@ class ModbusConnectedRequestHandler(ModbusBaseRequestHandler,asyncio.Protocol):
         self.receive_queue.put_nowait(data)
 
     async def _recv_(self):
-        return await self.receive_queue.get()
+        try:
+            result = await self.receive_queue.get()
+        except RuntimeError:
+            result = None
+        return result
 
     def _send_(self, data):
         """ tcp send """


dhoomakethu:
Any idea what was causing `RuntimeError` @janiversen ?

janiversen:
Yes it is documented, that the function raises a runtime error when the loop is already closed. In my mind it wrong to use runtime Error for that, but that is something we cannot change.

There are also no easy way to check is a loop is active (unless you have the loop object).

dhoomakethu:
Can we use [get_running_loop](https://docs.python.org/3.7/library/asyncio-eventloop.html#asyncio.get_running_loop) to identify these scenarios and output appropriate error message so that its easier to debug ?

dhoomakethu:
On the second thought, We still will have to catch `RuntimeError` though, which is what you fixed. May be using `get_running_loop` is not required.

janiversen:
it too does runtime error:
https://docs.python.org/3/library/asyncio-eventloop.html

There is an internal function _get_running_loop() that does what we need, but I am very much against using internal functions as they are bound to cause problems as versions change.

janiversen:
I actually tested with get_running_loop, before reading the manual 😬

dhoomakethu:
+1 for not using any internal methods from asyncio library. Let's just stick to a simple error or warning message.

PR #753, opened 2022-04-04
Add pypy3.8, 3.9, 3.10 to ci

janiversen:
we test Cpython on all platforms with 3.7, 3.8, 3.9, 3.10

But we only test pypy with 3.7

Let's change that.


swamper123:
3.10 can be removed, since it isn't supported atm.

https://www.pypy.org/download.html

PR #754, opened 2022-04-05
Use pytest_configure instead of import directly

janiversen:
Based on late review comments, contest is changed to use pytest_configure.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,15 +1,17 @@
+import pytest
 import platform
 from pkg_resources import parse_version
 
 
-IS_DARWIN = platform.system().lower() == "darwin"
-IS_WINDOWS = platform.system().lower() == "windows"
+def pytest_configure():
+    pytest.IS_DARWIN = platform.system().lower() == "darwin"
+    pytest.IS_WINDOWS = platform.system().lower() == "windows"
 
-if IS_DARWIN:
-    # check for SIERRA
-    if parse_version("10.12") < parse_version(platform.mac_ver()[0]):
-        SERIAL_PORT = '/dev/ptyp0'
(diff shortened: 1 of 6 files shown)


janiversen:
@dhoomakethu I think this is what you wanted ?

dhoomakethu (approved):

pymodbus/client/asynchronous/serial.py
```diff
@@ -38,6 +38,9 @@ def _framer(cls, method):
 
         raise ParameterException("Invalid framer method requested")
 
+    def __iter__(self):
```
dhoomakethu, line 41: Why do we have to make `AsyncModbusSerialClient` an iterable ?

janiversen: Without the __iter__ SonarCloud reported 2 bugs and asked that we use an iterable object, therefore I added __iter__

Actually sometimes I do not understand SonarCloud, I think it reports a lot of false positives.

dhoomakethu: Thats a false positive , it assumes it requires an iterable for having the variable `loop` in the examples.
image

dhoomakethu: Please remove that, I will mark that as false positivie

janiversen: Yes it is, but I cannot find a away to disable the specific line, expected to be able to do something like #sonarcloud: disable=iter but it seems sonar cloud do not support such features.

janiversen: I will remove it, but please tell me how to mark it as a false positive.

PR #755, opened 2022-04-05
Add _logger message when catching runtime error

janiversen:


janiversen:
@dhoomakethu this should make debugging easier ?

PR #756, opened 2022-04-06
pylint batch 1

janiversen:
The inclosed files passes pylint !

I have decided to keep the PRs with few files, because when merged into one commit it makes bisecting and reviewing a lot easier than if I put 100 files in each PR.

There are quite a lot more "pylint: disable=" than I normally like, but I decided only to do the changes that could be done without refactoring code, this makes review easier, and later we can address the "pylint: disable=".

### DIFF
--- a/pymodbus/server/reactive/main.py
+++ b/pymodbus/server/reactive/main.py
@@ -3,24 +3,25 @@ Copyright (c) 2020 by RiptideIO
 All rights reserved.
 """
 import os
+import sys
 import asyncio
 import time
 import random
 import logging
-from pymodbus.version import version as pymodbus_version
-from pymodbus.pdu import ExceptionResponse, ModbusExceptions
-from pymodbus.datastore.store import (ModbusSparseDataBlock,
-                                      ModbusSequentialDataBlock)
-from pymodbus.datastore import ModbusSlaveContext, ModbusServerContext
-from pymodbus.device import ModbusDeviceIdentification
 
 try:
     from aiohttp import web
 except ImportError as e:
     print("Reactive server requires aiohttp. "
           "Please install with 'pip install aiohttp' and try again.")
-    exit(1)
+    sys.exit(1)
 
+from pymodbus.version import version as pymodbus_version
+from pymodbus.pdu import ExceptionResponse, ModbusExceptions
+from pymodbus.datastore.store import (ModbusSparseDataBlock,
+                                      ModbusSequentialDataBlock)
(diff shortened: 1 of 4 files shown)


janiversen:
@dhoomakethu this is one for you CI 3.7 get "Error: Bad credentials".

janiversen:
@dhoomakethu "bad credentials" went away.

pymodbus/server/reactive/main.py
```diff
@@ -193,44 +196,43 @@ def manipulate_response(self, response):
...
+            err_response = ExceptionResponse(response.function_code, error_code)
+            err_response.transaction_id = response.transaction_id
+            err_response.unit_id = response.unit_id
+            response = err_response
+            self._counter += 1
+        elif response_type == "delayed":
+            delay_by = self._manipulator_config.get("delay_by")
+            txt = f"Delaying response by {delay_by}s for all incoming requests"
+            logger.warning(txt)
+            time.sleep(delay_by)
```
dhoomakethu, line 220: Note for self : This method should ideally be `async` and we should be using . TODO

dhoomakethu:
My thoughts are around fixing linter warning that are appropriate and disable only the things that does not matter. Let's include pylint but not fail tests and at the same time lets open some tasks to fix these linter errors.

janiversen:
I agree only to disable things that do not matter, but I prefer to do it in 2 steps:
step 1) have a clean run with pylint.
step 2) solve the important “pylint disable”
and then get on with black etc.

whenever I submit a PR I make sure all tests pass, that is a major item for me.


thanks for reviewing

pymodbus/server/reactive/main.py
```diff
@@ -193,44 +196,43 @@ def manipulate_response(self, response):
...
-                                   f"Using default lenght 10")
-                    data_len = 10
-                response = os.urandom(data_len)
-                self._counter += 1
-                skip_encoding = True
-            return response, skip_encoding
+
+        clear_after = self._manipulator_config.get("clear_after")
+        if clear_after and self._counter > clear_after:
+            txt = f"Resetting manipulator {clear_after} responses"
```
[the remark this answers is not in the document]
janiversen: good catch (copy/paste error), corrected.

pymodbus/server/reactive/main.py
janiversen: you are quite right, I have a note of changing all time.sleep, this will however be easier when there is a decision to remove tornado and twister.

janiversen: 👍 thanks did not know tat.

PR #768, opened 2022-04-11
pylint batch 10

janiversen:


dhoomakethu:
@janiversen thank you for taking time to cover these things alone. Apologies I am not able to support you with these tasks.

PR #771, opened 2022-04-12
Pylint batch 13

janiversen:


pymodbus/file_message.py
```diff
@@ -14,7 +14,7 @@
 #---------------------------------------------------------------------------#
 # File Record Types
 #---------------------------------------------------------------------------#
-class FileRecord(object):
+class FileRecord():
```
dhoomakethu, line 17: We can get rid of `()` as  well , `class FileRecord: ` will do fine as well.

PR #773, opened 2022-04-13
Change class name(): -> class name:.

janiversen:


janiversen:
@dhoomakethu thanks for the hint.

PR #797, opened 2022-04-21
Please SONAR (logging.basicConfig is a problem)

janiversen:


swamper123:
What happen?

janiversen:
I updated a lot of files, CI for pylint_28 was ok, but in CI sonar cloud only looks at changes.

When the commit is merged into dev SONARCLOUD controls the whole file, meaning "old" errors surface, that is what happened.

PR #805, opened 2022-04-21
Fix #716

PaulKGrimes:
Change the history file of the REPL from hardcoded `../.pymodhis` to hardcoded `.pymodhis`, so that REPL client will run from any working directory where the user has write permissions.


pymodbus/repl/client/main.py
```diff
@@ -178,7 +178,7 @@ def _process_args(args, string=True):
                             complete_while_typing=True,
                             bottom_toolbar=bottom_toolbar,
                             key_bindings=use_keys,
-                            history=FileHistory('../.pymodhis'),
```
dhoomakethu, line 181: I was planning to use a directory in users `HOME` directory and save the history file there. So that it would be available no matter from where the REPL is invoked.

pymodbus/repl/client/main.py
```diff
@@ -178,7 +178,7 @@ def _process_args(args, string=True):
...
+                            history=FileHistory('.pymodhis'),
```
janiversen, line 181: First of all thanks for the PR !
Your solution however do cause problems on Mac if you install it as an application (as I do) because the bin dir (macOS) is protected against changes.

For that reason I suggest this change, which is inline with @dhoomakethu :
```suggestion
                            history=FileHistory('~/.pymodhis'),
```
If you agree accept the suggestion and I will merge the PR.

janiversen (changes requested):
Please consider my suggestion.

pymodbus/repl/client/main.py
dhoomakethu: we will have to make sure  `FileHistory` can resolve `~` properly on windows as well.

PaulKGrimes: Using `os.path.normpath(os.path.expanduser("~/.pymodhis")` will make the path resolve correctly to the user directory across both Windows and Linux, if that's the desired behavior.  Sorry, don't have a Mac to test that on.

PaulKGrimes: I've made a commit with that added - I've also move the definition of the history file path to the top of `def cli():` block so that it's easier to change or to move to a configuration option or command line flag in the future.

pymodbus/repl/client/main.py
```diff
@@ -111,6 +113,7 @@ def convert(self, value, param, ctx):
 def cli(client): #NOSONAR pylint: disable=too-many-statements,too-many-branches
     """Client definition."""
     use_keys = KeyBindings()
```
janiversen, line 115: to me this is overkill, but I do not have a strong opinion (your suggestion work).

Moving it is a good idea, but then why not also make it a command line option as you suggested.

PaulKGrimes: That's certainly a better long term solution, but it also makes sense for this to be the default.  You will still want to do the normpath and expand user with the path received from command line option, just to make sure that you are handling the path correctly for the platform.

I'm not familiar with the click library or how you are handling command line options, so it'll be easier for someone else to implement that.

dhoomakethu: +1 on having the ability to change the path of the history file from command line and at the same time we should also be debating if this change needs to be persisted.

PaulKGrimes: Command line options really should always fall back to (or equivalently override) a default, rather than persisting between calls - you don't know how long it's going to be between calls, and whether the user will remember what they set it to last (I use detached tmux sessions a lot - sometimes it's weeks between me looking at a particular terminal session).  

But having an environment variable or config file that sets the default would be a good long term aim.

janiversen:
Thanks.

PR #806, opened 2022-04-22
Python 3.7 End Of Life.

janiversen:
Python 3.7 has been marked "End Of Life", therefore we no longer
support it or test it.

Python 3.7 is only receiving security patches, and the official EOL is 2023-06-27 with only 1 more security patch scheduled.

Python 3.7 is currently causing huge delays in our CI (15minutes runtime instead of 2-3minutes) and is also causing a number of "if PYTHON_VERSION > (3, 7):" in our code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,6 @@ jobs:
             runs-on: ubuntu-latest
             matrix: linux
             container:
-              3.7: docker://python:3.7-buster
               3.8: docker://python:3.8-buster
               3.9: docker://python:3.9-buster
               3.10: docker://python:3.10-buster
@@ -50,12 +49,6 @@ jobs:
               x86: win32
               x64: win64
         python:
-          - name: CPython 3.7
-            tox: py37
-            action: 3.7
-            docker: 3.7
-            matrix: 3.7
-            implementation: cpython
           - name: CPython 3.8
             tox: py38
             action: 3.8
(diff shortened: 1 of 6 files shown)


janiversen:
@dhoomakethu this is a PR, where I really want your approval before merging.

swamper123:
Where is it marked as EOL? It is still marked as security. https://devguide.python.org/#status-of-python-branches

Do you have the link for the schedule plan? This would be a good to know thing for me and further planing with other projects. ^^

I agree, that this "is_python_version" stuff is annoying and meh, but wouldn't remove it now. Yes the CI/CD is choking on it, but is this caused by the tests or by the building setup? If I remember right, cryptography needs ~ 5min. to be built in some tests. So the enviroment has to be prepared better.

We could flag/mention that this support will be dropped, when python drop it as well.

janiversen:
You are right it is currently marked as "security", but that means "only security fixes are accepted and no more binaries are released, but new source-only versions can be released" so in practice only developers can install this version (and only have been able to for a while). Therefore I do not see a lot of reason to keep it around.

I also see more docker images etc drop their 3.7 versions.

Here you have a link to EOL dates:
https://endoflife.date/python

There is a PEP for each release with its schedule, that is the official documentation, if you want to dig deeper you need to follow the dev conversations:
https://peps.python.org/pep-0537/

Reworking the code, I am struggling with the "if version". But a compromise could be to let 3.0.0 be the last release support 3.7 and drop it imidiatly after.

dhoomakethu:
I will take a look by the end of the week. Thanks

dhoomakethu:
This looks good to me. Thanks. Please merge if you have no more commits to add.

dhoomakethu (approved):

PR #807, opened 2022-04-22
Remove asynchronous/deprecated directory

janiversen:
Cleanup code before 3.0.0 release.

I cannot see any part of the code referring "deprecated", so it looks like a remembrance from old days.


janiversen:
I have a little more work on this one.

janiversen:
@dhoomakethu any comments ?

dhoomakethu:
@janiversen the deprecated folder is used here https://github.com/riptideio/pymodbus/blob/dev/pymodbus/client/asynchronous/__init__.py#L40

This is required to get the `twisted` version of asynchronous library working. I need to really check if this is still required but this can potentially break the twisted client,.

janiversen:
Strange all tests passed, but your point it very valid, I will let this wait until we remove twister/tornado (something I am slowly starting to think about).

janiversen:
No need for this.

PR #812, opened 2022-04-22
Jump timeout to 25 minutes, pypy3.7 fails too often on timeout

janiversen:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     container: ${{ matrix.os.container[matrix.python.docker] }}
     # present runtime seems to be about 2 minutes
     # with pypy being the exception (10 minutes)
-    timeout-minutes: 15
+    timeout-minutes: 25
     strategy:
       fail-fast: false
       matrix:


swamper123:
I am not very deep inside CI/CD topics. Raising up the timeout to 25min. is one way, but not suer if this is the best solution (well at least everything is fresh and tested, but still needs much time...hope that we will never use a more complex library than cryptography to be build).

Is it possible to say that only the testing part shall be timeouted after x min?
Or cache the building setup somehow for 12/24h and reuse it for testing?

janiversen:
Why do you want to have different timeouts for each test ? I do not understand what it would bring. Also I do not think CI permits it.

Cache the building setup is something that are done in many project and something we should consider, but someone needs to do it (are you volunteering?). Again I am not sure what your goal is, we do not pay for the CI, so optimizing it is not really a benefit for us. If we were a large project like e.g. Libreoffice (which I used to maintain) it is a demand because a complete build is 3-4 hours.

Apart from that the problem is solved since we do not test 3.7 anymore, and pypy build a factor faster with 3.8, so now the timeout can be reduced to something like 5 minutes.

swamper123:
I try to be more concrete what I mean. My wording doesn't fit sometimes so well. Sorry for that :/

At the moment, if you push a commit, Github starts all the checks/Jobs:
grafik

At that moment, the 15 min. are running.
Each check contains different steps (Set up job, rustu, Test ...):

grafik

Now the closer look on the time. The Step "Test" needed in this case 6min. and 26 sec.
If you check the logs, you see that there is much creating, copying, buildinhg, downloading, installing and at the end the testing itself.

It seems that we have got two substeps here:
1) Build everything
2) Test everything

If you turn on the timestamps (upright corner), where tiem was spent:

Start of the "Test" step:
grafik

And the start of the Tests itself:
grafik

You see that ~ 6min. of the 7min. 54sec. were spent by that building substep (and longer in other, where the job was cancelled).

As (possible) solutions I got:
- we seperate the "Test" step in two steps and put a max. timeout time on the 2. step. Also disable the Job-Timeout or extend it to ~ 1hr (?)
- expend the timeout each time, if a job was cancelled due to to such a long building step
- we cache the setup in intervalls and run the tests on them

But I agree, that a larger project, like libreoffice, needs more time and optimizations on CI side compared to our small pymodbus lib. :)
If I got some time to dig in the topic, I will take a look if I can do sth, but I can't promise to provide a prompt solution.

janiversen:
Actually your description matches what I understood first time, what I still do not get is why you want to fine tune "timeout", that does not make sense to me.

We can just set it to 1 hour if we want to be sure. Remember the only purpose of timeout is to catch a endless job.

Maybe you mix timeout with actual execution time...at least your ideas of caching affects the runtime and not really the timeout.

PR #841, opened 2022-04-25
Reduce timeout to 15minutes

janiversen:
This is possible because we longer test version 3.7.


janiversen:
@swamper123 reduced the timeout while drinking my morning coffee.

PR #865, opened 2022-04-28
Solve flake8 problems, part 1.

janiversen:
It seems flake8 have been active in CI for quite a while, but the problems reported by flake8 seems to have gone unnoticed.

In order to standardize the code, and make the use of flake8 make sense, the report will be solved in some PRs.

This is a big step in making the code pep8 compatible, and should reduce the errors from black quite a lot.


janiversen:
Moved to a fork.

PR #866, opened 2022-04-29
Revert pep8 work, but keep fix #716, #804 and drop support for 3.7.

janiversen:
With this PR the state should be as was before the pep8 work started.

The work done until 29 April 2022 is kept in dev_pep8, any commit wanted in dev, can be cherry-picked from there.

Ongoing work will most likely continue in https://github.com/janiversen/pymodbus, until such time where the code is pep8 compliant (meaning when dev_pep8 is integrated into dev, again).


janiversen:
@dhoomakethu as you can see the "old" dev have quite a lot of sonar cloud problems, which have not surfaced earlier because the code was not changed.

I suggest merging this PR as it is, since the problems reported are not caused by this PR.

PR #871, opened 2022-05-01
(merge 3): clean flake8

janiversen:
This is a very large PR but complete as wished.

To be honest I would never review a PR that large, but have fun.

DO NOT merge before #870 is merged, in order to keep important history.


swamper123:
Mostly spaces, nice addition in comments....nothing wild as far as I see... :heavy_check_mark:

.flake8
```diff
@@ -0,0 +1,5 @@
+# flake8 configuration.
+[flake8]
+exclude = .tox,.git,__pycache__
+ignore = D211,D400,E251,E731,W503
```
dhoomakethu, line 4: `E251 - unexpected spaces around keyword / parameter equals`  Can be fixed easily (??)

janiversen: Yes but it is massive change, so I assumed it was better to leave it.

However I can add that, preferable in a later PR.

dhoomakethu (approved):

dhoomakethu:
Looks like this PR is conflicting with some of the files.

PR #872, opened 2022-05-02
(merge 4): Allow pylint to check with all messages active.

janiversen:
The only "missing" control is "duplicate code", but that is calling for massive changes, so that will be done later.

This concludes all the semi-automatic changes to the code, until black is activated.

DO NOT merge before https://github.com/riptideio/pymodbus/pull/871 is merged, in order to keep important history.


pymodbus/repl/server/cli.py
```diff
@@ -17,13 +19,13 @@
 _logger = logging.getLogger(__name__)
 
 TITLE = (
```
dhoomakethu, line 21: We need to get a better way of representing this banner. This still gets mangled up.

PR #875, opened 2022-05-03
Simplify requirements*.txt

janiversen:
Cleaned installation procedure:

requirements.txt contains all dependencies, mainly because it is only used by people who clone the repo, and thus are very likely to be developers who need all dependencies (or is capable of adjusting to Specific needs).

Setup.cfg contain all fixed setup information.

Setup.py reads requirements.txt to generate dependencies as either fixed for opcional (allowing Pymodbus[<option>]

README.rst have been updated to reflect the above changes.


janiversen:
The CI errors are not caused by this PR, but probably somewhere in managing a stack of 6 PR on top of each other (waiting for approval).

Once approved I will rebase each PR and let CI verify.

requirements.txt
```diff
@@ -1,44 +1,49 @@
...
+# used by REPL.
+click>=6.7
+prompt-toolkit==2.0.4
+
+# used to enable serial communication.
+pyserial>=3.5
+pyserial-asyncio==0.4.0
+
+# used by datastore.
+sqlalchemy>=1.1.15
```
dhoomakethu, line 30: We do not have to install these modules unless users explicitly wants to use these features. For e.g if somebody wants to install pymodbus with `REPL` they can simply to `pip install pymodbus[repl]`.

dhoomakethu: Refer setup.py

requirements-tests.txt
```diff
@@ -1,25 +1,41 @@
...
+# Which are all part of a normal CI run.
+#
+# REMARK: if you have commented out packages in requirements.txt it will fail.
+-r requirements.txt
+
+
+# -------------------------------------------------------------------
+# Python packages needed to run tests
+# -------------------------------------------------------------------
 pytest>=3.5.0
```
dhoomakethu, line 20: We can also upgrade pytest to latest (6.x and above)

requirements.txt
janiversen: I agree it is not a demand, but it makes everything a lot easier, and what is the drawback...they occupy a bit of space in "site-packages". I wrote in the text that they can be commented out.

requirements-tests.txt
```diff
@@ -1,25 +1,41 @@
...
-# using platform_python_implementation rather than
-# implementation_name for Python 2 support
-Twisted[conch,serial]>=20.3.0; platform_python_implementation != "PyPy" or sys_platform != "win32"
-# pywin32 isn't supported on pypy
-# https://github.com/mhammond/pywin32/issues/1289
-Twisted[conch]>=20.3.0; platform_python_implementation == "PyPy" and sys_platform == "win32"
-zope.interface>=4.4.0
+pytest-cov>=2.5.1
+pep8>=1.7.0
 asynctest>=0.10.0
```
dhoomakethu, line 23: I remember there was an issue with `asynctest` for some versions of pytest. Will have to check that out.

janiversen: ok

requirements.txt
dhoomakethu: I am ok with commented out `requirements.txt` for easy reference except for pyserial and pyserial-async packages

janiversen: My preference was to go the other way, install all unless the end-user comments it out. The reason is that many users to not view the content of the files but simply install them.

requirements-tests.txt
dhoomakethu: Refer https://github.com/riptideio/pymodbus/issues/681

swamper123: In the actual versions of python, unittest support async stuff.
I would either recommened to switch to [builtin stuff](https://docs.python.org/3/library/unittest.html?highlight=unittest#unittest.IsolatedAsyncioTestCase) or as an alternative use the pytest plugin "[pytest-asyncio](https://github.com/pytest-dev/pytest-asyncio)".

['asynctest'  seems to be abandoned (last commit 2019)](https://github.com/Martiusweb/asynctest)

janiversen: That is a good hint !!!

I will take a look later and see how easy it is to remove, stay tuned.

janiversen: I think I am able to remove asynctest, I will do that before this one gets merged.

@dhoomakethu I removed package "attrs", because I could not find any references, but I just saw a commit in your "black" branch that upgrade the version, have I missed something ?

dhoomakethu: I do not remember using `attrs` any where explicitly, I believe it came in as part of the `black` package dependency.

janiversen: thanks, yes it is a requisite for black, but we should not list the dependencies for other packages.

janiversen:
It turned out removing asynctest was not as easy as I thought (pytest structure is better than unittest!!!), so I left that for now.

I have updated all test/check packages to their newest version (including for REPL!). My preference is to do it in a "bump" PR, but...

We still need to update
- click
- prompt-toolkit
- SQLAlchemy
- redis
but they seem to demand code changes.

janiversen:
@dhoomakethu, @swamper123 
I have now remate this PR, please see PR-text and README.rst

This is a simpler way of handling all the different needs, and ensuring dependency versions are kept identical (Currently different Versions are used in the different setup files).

Please have a look, and let me know what you Think.

README.rst
```diff
@@ -125,15 +122,14 @@ Examples Directory structure
 If you are looking for UI,checkout `Modbus Simulator <https://github.com/riptideio/modbus-simulator>`_ or
 `Modbus Cli <https://github.com/dhoomakethu/modbus_sim_cli>`_
 
-Also, if you have questions, please ask them on the mailing list
-so that others can benefit from the results and so that I can
-trace them. I get a lot of email and sometimes these requests
+Also, if you have questions, please create an issue so that others can benefit from the results and so that we can
```
swamper123, line 125: Better:

Also, if you have a question, please [create an post in discussions q&a topic](https://github.com/riptideio/pymodbus/discussions/new?category=q-a), so that others can benefit from the results [...].

If you think, that something in the code is broken/not running well, please [open an issue](https://github.com/riptideio/pymodbus/issues/new), read the Template-text first and then post your issue with your setup informations.

setup.cfg
```diff
@@ -1,3 +1,90 @@
...
+keywords         = 'modbus, asyncio, scada, client, server, simulator'
+long_description = file: README.rst, CHANGELOG.rst, LICENSE
+url              = 'https://github.com/riptideio/pymodbus/'
+classifiers      = 
+    Development Status :: 4 - Beta
+    Programming Language :: Python :: 3.8 :: 3.9 :: 3.10
+    Environment :: Console
+    Framework :: Twisted :: Tornado :: Asyncio
+    Intended Audience :: Developers :: Modbus testers
+    License :: OSI Approved :: BSD License
```
swamper123, line 19: I am not so deep in that Licence thing....do we have a BSD Licence? What kind of Licence is this one https://github.com/riptideio/pymodbus/blob/dev/LICENSE ? @dhoomakethu

swamper123: Just saw, that this is the old state as well...

swamper123 (changes requested):
We should add the difference between discussion and issue here with a link to avoid mixing up both.
My comment about  the license thing is obsolete .. saw it to late that this seems to be utd.

README.rst
janiversen: Done.

swamper123 (approved):

dhoomakethu:
Looks good to me.

dhoomakethu (approved):

PR #876, opened 2022-05-03
Fix version check when importing CancelledError

### LINKED ISSUE #690: AttributeError: module 'asyncio' has no attribute 'exceptions'
```
  File "/home//.local/bin/pymodbus.server", line 6, in <module>
    from pymodbus.repl.server.main import server
  File "/home//.local/lib/python3.7/site-packages/pymodbus/repl/server/main.py", line 16, in <module>
    CANCELLED_ERROR = asyncio.exceptions.CancelledError
AttributeError: module 'asyncio' has no attribute 'exceptions'
```

in the code 

```
if IS_PYTHON3 and PYTHON_VERSION > (3, 7):
    CANCELLED_ERROR = asyncio.exceptions.CancelledError
else:
    CANCELLED_ERROR = asyncio.CancelledError

```
python 3.7 at least 3.7.3 doesn't have `asyncio.exceptions`
Maybe it was supposed to be >= 3.8.

velovix:
This ensures that all Python 3.7.x versions and earlier do not depend on the existence of `asyncio.exceptions`.

This is just the PR form of the suggestion made by @dhoomakethu in #690. Admittedly very low hanging fruit as far as contributions go :)

Fixes #690

### DIFF
--- a/pymodbus/repl/server/main.py
+++ b/pymodbus/repl/server/main.py
@@ -14,7 +14,7 @@ from pymodbus.server.reactive.main import (
 from pymodbus.server.reactive.default_config import DEFUALT_CONFIG
 from pymodbus.repl.server.cli import run_repl
 
-if sys.version_info > (3, 7):
+if sys.version_info >= (3, 8):
     CANCELLED_ERROR = asyncio.exceptions.CancelledError
 else:
     CANCELLED_ERROR = asyncio.CancelledError  # pylint: disable=invalid-name


janiversen:
on dev we no longer support python < 3.8 thus this is not relevant, however if one day version 2.5.4 is made this should be included.

Thanks for your PR.

janiversen (approved):

swamper123:
Shouldn't this be better merged in  `2.5.3` instead of `dev`?

janiversen:
That is why I put the 2.5.0 label, right now it will not be merged...only if we decide to make 2.5.4

PR #884, opened 2022-05-09
Allow coroutine call from sync __init__

janiversen:
server/async_io calls coroutine functions in __init__ which pr definition is sync, this causes problems!!

Added the sync. variant of await.

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -477,14 +477,14 @@ class ModbusTcpServer:  # pylint: disable=too-many-instance-attributes
         # constructors cannot be declared async, so we have to
         # defer the initialization of the server
         self.server = None
-        self.server_factory = self.loop.create_server(
+        self.server_factory = asyncio.ensure_future(self.loop.create_server(
             lambda: self.handler(self),
             *self.address,
             reuse_address=allow_reuse_address,
             reuse_port=allow_reuse_port,
             backlog=backlog,
             start_serving=not defer_start
-        )
+        ))
 
     async def serve_forever(self):
         """Start endless loop."""
@@ -593,7 +593,7 @@ class ModbusTlsServer(ModbusTcpServer):  # pylint: disable=too-many-instance-att
         # defer the initialization of the server
         self.server = None
         # start_serving is new in version 3.7
-        self.server_factory = self.loop.create_server(
+        self.server_factory = asyncio.ensure_future(self.loop.create_server(
             lambda: self.handler(self),
             *self.address,
             ssl=self.sslctx,
@@ -601,7 +601,7 @@ class ModbusTlsServer(ModbusTcpServer):  # pylint: disable=too-many-instance-att
             reuse_port=allow_reuse_port,
             backlog=backlog,
             start_serving=not defer_start
-        )
+        ))
 
 
 class ModbusUdpServer:  # pylint: disable=too-many-instance-attributes


pymodbus/server/async_io.py
```diff
@@ -477,14 +477,14 @@ def __init__(self,  # pylint: disable=too-many-arguments
         # constructors cannot be declared async, so we have to
         # defer the initialization of the server
         self.server = None
-        self.server_factory = self.loop.create_server(
+        self.server_factory = asyncio.ensure_future(self.loop.create_server(
```
dhoomakethu, line 480: What exactly was the issue ? `loop.create_server ` returns a [server](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_server) object which is awaited when the server is started https://github.com/riptideio/pymodbus/blob/dev/pymodbus/server/async_io.py#L492 . With this change, the result in the future needs to be explicitly retrieved on line 492.

May be a little more context will help. Thanks.

janiversen: test_asyncio.py fails with “object not awaited”, during my removal of asynctest.

Create_server is async (coroutine), and is called from sync method (__init__). This is in general not adviceable and it most cases leads to dangling futures (which happens here).

swamper123: Yes, create_server is a coroutine and has to be scheduled to the loop.
I would prefere: `asyncio.create_task(self.loop.create_server(.... ` instead of `ensure_future`...the docs say it's the prefered way since 3.8 upwards.

swamper123: Still something has to await `self.server_factory` anywhere in the end. This just schedules the coroutine in the eventloop.

janiversen: It actually does a lot more, because the Code starts to execute, and the future becomes dangling. Create_task changes the whole idea but is a solution.

Basically when not awaiting the créate_server, you have a timing issue to get an await in place, before the Code internally reached a Point where a await is needed, and that is exactly what happens in the test.

janiversen: (venv) jan@MacMini test % pytest test_server_asyncio.py 
============================= test session starts ==============================
platform darwin -- Python 3.10.4, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/jan/repos/pymodbus, configfile: setup.cfg
plugins: timeout-2.1.0, cov-3.0.0
collected 17 items                                                             

test_server_asyncio.py .................                                 [100%]

============================== 17 passed in 0.46s ==============================
sys:1: RuntimeWarning: coroutine 'AsyncMockMixin._execute_mock_call' was never awaited
Coroutine created at (most recent call last)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 633, in run_until_complete
    self.run_forever()
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 600, in run_forever
    self._run_once()
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 1888, in _run_once
    handle._run()
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/events.py", line 80, in _run
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/async_case.py", line 101, in _asyncioLoopRunner
    ret = await awaitable
  File "/Users/jan/repos/pymodbus/test/test_server_asyncio.py", line 62, in test_async_start_tcp_server
    server = await StartTcpServer(context=self.context, loop=self.loop, identity=identity)
  File "/Users/jan/repos/pymodbus/pymodbus/server/async_io.py", line 840, in StartTcpServer
    server = ModbusTcpServer(context, framer, identity, address, **kwargs)
  File "/Users/jan/repos/pymodbus/pymodbus/server/async_io.py", line 480, in __init__
    self.server_factory = self.loop.create_server(
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/mock.py", line 1104, in __call__
    return self._mock_call(*args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/mock.py", line 1108, in _mock_call
    return self._execute_mock_call(*args, **kwargs)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
(venv) jan@MacMini test %

janiversen: But I have an alternative solution, which I Will push in a moment.

swamper123: What if we put the async stuff in a `@classmethod` ?

Just found that snippet which looks nice in some way, but would change the usage of the class a bit:

import asyncio

```python
class Foo(object):
    @classmethod
    async def create(cls, settings):
        self = Foo()
        self.settings = settings
        self.pool = await create_pool()
        return self

async def main(settings):
    settings = "..."
    foo = await Foo.create(settings)

```

Or could that leed to problems somehow?

janiversen: That is the ideal solution, but it demands that all examples and real Life Applications be changed. I prefer a solution where the external interface is kept unchanged.

janiversen:
The Code is wrong, but for now I Think it is easier to adapt the tests.

pymodbus/server/async_io.py
dhoomakethu: I believe it needs to be handled as part of the test itself.  In a normal server startup , `serve_forever` is always to be called and is awaited, this i where the `server_factory` is being awaited [Refer](https://github.com/riptideio/pymodbus/blob/dev/pymodbus/server/async_io.py#L489). We can move `self.server_factory = self.loop.create_server(` on line 480 inside `serve_forever` as a workaround if that fixes the issue,

janiversen: "is always" is actually "should always be" because there are no guarantee that it happens (I saw this Problem first time  when rewriting HA-modbus tests to use real).

But your idea is a good one, I do not see that as a workaround, but merely grouping the Code in a more logical way.

PR #889, opened 2022-05-13
Move create_server to serve_forever()

janiversen:
In order to avoid the python error "object not awaited", create_server is moved from __init__ to serve_forever.

Earlier the code looked like:
´´´
startTcpServer()
      with factory = create_server()
....unpredictable application code...
create_task(serve_forever())
     with await factory
´´´
Without any guarantee that serve_forever() would be called.

Now the code will look like:
´´´
startTcpServer()
....unpredictable application code...
create_task(serve_forever())
      with factory = await create_server()
´´´
and thus no error is possible, due to bad application code.

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -61,7 +61,10 @@ class ModbusBaseRequestHandler(asyncio.BaseProtocol):
             _logger.error(
                 "Handler for serial port has been cancelled")
         else:
-            sock_name = self.protocol._sock.getsockname()  # pylint: disable=protected-access,no-member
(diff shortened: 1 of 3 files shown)


dhoomakethu (approved):

swamper123:
It seems to stuck on `DeprecationWarning`. https://github.com/janiversen/pymodbus/blob/8876a9bcd32c825da92d051c7e75e6671a960c04/test/test_server_asyncio.py#L621

janiversen:
No it is simply hanging in factory, for some reason.

PR #903, opened 2022-05-20
Remove tornado/twister from test and examples

janiversen:
Isolate examples that uses tornado/twister to a new subdirectory.

Remove tornado/twister from test, these should be rewritten as asyncio tests.

This is the very first step in no longer supporting tornado/twister. It also allows me to start clean/amend test cases.

Removed files containing twister/tornado from coverage, in order to keep the %.


janiversen:
@dhoomakethu any comments ?

dhoomakethu:
We can exclude the source from coverage so as to keep the coverage above 88/90%.
Refer https://github.com/riptideio/pymodbus/blob/dev/.coveragerc

janiversen:
I will update the PR shortly.

janiversen:
Lets get this in before the next dev release.

PR #905, opened 2022-05-22
Fix readmethe docs generation

dhoomakethu:
* Fix read the docs building with python3.5
* Fix make file referencing wrong requirements file

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,10 +2,10 @@
 formats:
     - epub
     - pdf
-requirements_file: requirements-docs.txt
+requirements_file: requirements.txt
 python:
     extra_requirements:
         - twisted
         - tornado
         - documents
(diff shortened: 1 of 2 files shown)


janiversen:
Sorry this was my fault, I was no5 aware of the . file.

The Makefile either needs to go away, or be repaired….because all the new tools pylint etc, are to be run with “tox -e <tool>”.

Do you use the makefile? today it is not very common in python projects.

dhoomakethu:
No prob, can you check if these generated docs for `dev` branch is good enough ? https://pymodbus.readthedocs.io/en/dev/

janiversen:
I miss dev4 in CHANGELOGS.

Apart from the docs look ok, but now I can see that I need the pylint check for :param: in order to secure there are match.

PR #909, opened 2022-05-22
Fix parameter documentation of methods.

janiversen:
This PR correct all missing/false parameter documentations.

At the moment there are circa 800 errors, so I believe it is safe to say that our documentation lack a bit.

I have submitted the PR to raise awareness that I have started on this work, however help is more than welcome.

To see the problems, download this PR "gh pr checkout 909", and run "tox -e pylint".


janiversen:
@swamper123 This could be something for you,  a nice way of getting into the code.

PR #913, opened 2022-05-25
Try again for test coverage checking

altendky:


janiversen:
Now it works as intended (I added a small correction):
https://github.com/riptideio/pymodbus/runs/6599071687?check_suite_focus=true

And there are (to my surprise) quite a lot of issues in test! THANKS for your idea.

janiversen:
@altendky do you want to add more, or is it ready to be merged ? I would like to see it merged, so I can start solve the issues.

altendky:
I would like to see the settings in the config, not on the command line.  Then they apply more universally.  But, this is getting a report so no need to hold it up.

janiversen:
I agree to that, but that is a problem because the plugin does not use setup.cfg, only command line.

altendky:
The plugin is just a wrapper around coverage so I would expect it to be possible.  I also know I've had trouble getting the right incantations in the config in some cases.  `:]`

PR #914, opened 2022-05-26
Fix doc generation errors and re-organize examples

dhoomakethu:


janiversen:
I have updated your PR, because a lot of the changes was already done.

I have no problem moving the examples back, but then they should be changed to be asyncio only. Please do not forget the idea to move them in the first place, was to get rid of twisted/tornado.

Why is it important to have e.g. examples/contrib/tornado_twisted/sunspec_client.py direct in examples ?

A compromise could be to let the documentation point at the tornado_twisted directory.

dhoomakethu:
`sunspec` client I did not check but I was moving the synchronous and updating server examples which are based on synchronous server (socketserver). I am OK with keeping only twisted and tornado examples separate and the same time lets also try to keep the `.py` extension as it is till we completly get rid off them.

janiversen:
Now the doc can be generated here (after a rebase and an extra commit).

I will take a closer look now at the files you moved.

Sunspec was my bad, I was confused by something else.

janiversen:
- synchronous_server.pyX
Was already moved on dev.
- updating_server.pyX and dbstore_update_server.pyX
uses twisted, see line 16
```
from twisted.internet.task import LoopingCall
```

Now I can see that there are missing rst files, let me update that.

janiversen:
@dhoomakethu Please exclude the examples directory from Sonar, or at the very least remove the check for duplicate code (which is quite normal between examples).

janiversen (approved):

PR #918, opened 2022-05-26
Add doc generation to CI

janiversen:
Test if documents can be generated.


dhoomakethu:
Instead of this we can enable [build docs for PR](https://docs.readthedocs.io/en/stable/pull-requests.html) on readthedocs. Which should add an extra step in the build pipeline.

dhoomakethu:
@janiversen I have enabled build docs for PR on the readthedocs. JFYI

janiversen:
@dhoomakethu Your solution is the better one, closing this one.

PR #930, opened 2022-05-31
__data should be an object and not class variable

### LINKED ISSUE #694: Modbus Device Identification details overwrite
### Versions

* Python: 3.7.0
* OS: Windows 10
* Pymodbus: 2.5.2

### Pymodbus Specific
* Server: tcp, sync
* Client: tcp sync

### Description

I am running two Pymodbus servers on the same machine in two threads. I have given two sets of identification information for each server. But Device information gets overwritten. (i.e. I always get the device information of the last created server even I request information from the first server)

### Server Code 

#!/usr/bin/env python
from pymodbus.server.sync import StartTcpServer
from pymodbus.device import ModbusDeviceIdentification
from pymodbus.datastore import ModbusSequentialDataBlock
from pymodbus.version import version
from threading import Thread 

class ModbusServer:

    def __init__(self,_ip,_port,_vendorName,_ProductCode,_VendorUrl,_ProductName,_ModelName):
        self._ip = _ip
        self._port = _port
        store = ModbusSlaveContext(
            di = ModbusSequentialDataBlock(0, [0]*100),
            co = ModbusSequentialDataBlock(0, [0]*100),
            hr = ModbusSequentialDataBlock(0, [0]*100),
            ir = ModbusSequentialDataBlock(0, [0]*100))
        self.context = ModbusServerContext(slaves=store, single=True)

        # initialize the server information
        self.identity = ModbusDeviceIdentification(self)
        self.identity.VendorName = _vendorName
        self.identity.ProductCode = _ProductCode
        self.identity.VendorUrl = _VendorUrl
        self.identity.ProductName = _ProductName
        self.identity.ModelName = _ModelName
        self.identity.MajorMinorRevision = version.short()

    def _run(self):
        StartTcpServer(self.context, identity=self.identity, address=(self._ip, int(self._port)))
        
    def run(self):
        Thread(target=self._run).start()

Device1=ModbusServer("127.0.0.1",1500,"sample1","sample1","sample1","sample1","sample1")
Device1.run()
Device2=ModbusServer("127.0.0.1",1501,"sample2","sample2","sample2","sample2","sample2")
Device2.run()

# client code

from pymodbus.client.sync import ModbusTcpClient
from pymodbus.mei_message import ReadDeviceInformationRequest
from pymodbus.device import *

client = ModbusTcpClient('127.0.0.1','1500')
rq = ReadDeviceInformationRequest(unit=1,read_code=3)
rr = client.execute(rq)
print(rr.information)
client.close()

# result
{0: b'sample2', 1: b'sample2', 2: b'2.5.2', 3: b'sample2', 4: b'sample2', 5: b'sample2'}

But this is the result I wanted  {0: b'sample1', 1: b'sample1', 2: b'2.5.2', 3: b'sample1', 4: b'sample1', 5: b'sample1'}

janiversen:
fixes #694


janiversen:
Sadly enough, this one needs more changes, because ModbusControlBlock is also wrong.

janiversen:
Actually this Problem proved to need a bigger rewrite of the core. ModbusControlBlock is a singleton, and therefore cannot really handle 2 servers in the same Code.

zmatlik117:
...hell, that might explain one weird as **** issue i encountered when i launched approx. 18 modbus servers - mixed TCP and RTU - in single process, multiple threads in an attempt to perform a black-box test of a certain industrial device...
anyway thanks for trying at least

janiversen:
@dhoomakethu do you see it as safe to mode identity from _data to context ? There are exactly one context for each server, to it would allow to use threads without depending on a singleton.

dhoomakethu:
> @dhoomakethu do you see it as safe to mode identity from _data to context ? There are exactly one context for each server, to it would allow to use threads without depending on a singleton.

@janiversen Identity is set once per server and ideally it should be read-only, I am not sure what advantage it would bring by moving it to context ?

janiversen:
Because currently context only exist once pr process, due to the use of a singleton around _data. This means you cannot have 2 servers in the same process.

Context is stored in the server object, so that would allow to have exactly one identity pr server.

dhoomakethu:
What's the use case for having two servers in the same process? Also when you say same process do they use the same port or different port?

janiversen:
look at #694.

Of course they use different ports, quite complicated to have 2 listen on the same port.

As I see it someone designed a singleton to secure data was accessed in a serialised way (which is good), but forgot the implication that you cannot have 2 instances in the same process.

I am currently looking into a problem with examples/serial_forwarder.py and it could be an identical problem not with identity but with other parts of _data.

PR #951, opened 2022-07-05
Store method name in framer itself

janiversen:
Currently texts like "ascii" is distributed out in the code, put the name (text) in the framer itself, makes a fixed relationship, which can be easier controlled.

This is the first part of a more generic change where the different client/server classes only will allow framers valid for the type of communication and otherwise throw an exception. The generic change also includes support for a custom framer to be supplied by the calling app, without overwriting our code.

### DIFF
--- a/pymodbus/framer/__init__.py
+++ b/pymodbus/framer/__init__.py
@@ -16,6 +16,8 @@ TLS_FRAME_HEADER = BYTE_ORDER + "B"
 class ModbusFramer(IModbusFramer):
     """Base Framer class."""
 
+    name = ''
+
     def _validate_unit_id(self, units, single):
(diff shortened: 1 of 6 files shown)


dhoomakethu:
@janiversen , just so that we are clear, pymodbus is flexible enough to mix framers with different client implementations like try framer with tcp client etc. I am hoping these changes do not affect those capabilities.

janiversen:
It will not, that is a feature that are important!

I have however since seen, that "method=" for the serial clients should really be "framer=" to be consistent across all. I am trying to find out why "method=" was added instead of being consistent, but I cannot find the reason.

There are some combinations that do not really make sense like fx. ModbusTlsFramer with anything else than the TLS client/server.

PR #972, opened 2022-07-23
Remove twisted

janiversen:


janiversen:
@dhoomakethu would you please have a look at this.

I am aware that schedulers is not removed, I want to do that in a separate PR, because it affects a lot of files.

janiversen:
Big breath !!! this is going to make some waves in the pond.

shajek:
Hello, i think that this change broke [updating_server.py](https://github.com/riptideio/pymodbus/blob/dev/examples/common/updating_server.py) example

janiversen:
This PR have been merged for quite a while and thus closed. This is not a place to discuss new problems, open an issue if you have problem but please do tell why it is broken and add a log showing the problem.

As always Pull requests are more than welcome.

PR #976, opened 2022-07-24
Remove schedulers

janiversen:


janiversen:
Last hurdle of removing tornado/twisted...the rest is pure code efficient.

PR #991, opened 2022-07-29
Revert "Remove async get_factory"

janiversen:
Reverts riptideio/pymodbus#990

Have to check why pypi started to complain.


janiversen:
not needed

PR #1020, opened 2022-08-08
Fix port typo

alexrudd2:
I'm trying to keep abreast of breaking changes for a downstream driver [numat/productivity](https://github.com/numat/productivity) and ran across a typo.


janiversen:
Good catch, thanks.

PR #1032, opened 2022-08-10
Automate client/server examples.

janiversen:
The automated tests found a couple of problems, which needs to be addressed.

There are 3 principal problems:
a) get tls to work without a certificate
b) get serial to work on all platforms (which port to use on which platform)

### DIFF
--- a/examples/client_async.py
+++ b/examples/client_async.py
@@ -43,9 +43,12 @@ from pymodbus.transaction import (
 )
 
 
-def setup_async_client():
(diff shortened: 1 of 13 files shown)


janiversen:
@dhoomakethu this PR should please you. I hope it can be used as basis for the manual tests you do, please make another test file.

janiversen:
Once this file is ok, and yours is done we can merge them and see if they can be simplified.

PR #1040, opened 2022-08-16
Fix docs typos

alexrudd2:
Fix typos.  Mostly just me learning codespell :)
I separated the changelog ones in case you want that to be immutable


janiversen (changes requested):
Changelog.rst is a generated document from the commits, and once released it should be changed. It should be excluded from codespell.

janiversen (approved):
LGTM.

PR #1041, opened 2022-08-17
Add unit to arguments

pkubanek:
Fix #673. Adds unit to argument list for selected functions.

PR #1043, opened 2022-08-17
Fix #628: hex values for repl arguments

### LINKED ISSUE #628: Add support for hex values in REPL client.write_registers
Modbus registers are 16-bit and sometimes it is easier to understand their values when expressed in hex.

However, in the REPL tool "pymodbus.console", it looks like you cannot use hex for the values:

> client.write_registers unit=1 address=9 values=0x0331
ValueError("invalid literal for int() with base 10: '0x0331'")

It would be nice if you could use hex like depicted above.

pkubanek:
Also fixed some issues with repl client for unit handling

### DIFF
--- a/pymodbus/repl/client/main.py
+++ b/pymodbus/repl/client/main.py
@@ -146,9 +146,9 @@ def cli(client):  # noqa: C901 pylint: disable=too-complex
                 if not string:
                     if "," in val:
                         val = val.split(",")
-                        val = [int(v) for v in val]
+                        val = [int(v, 0) for v in val]
                     else:
-                        val = int(val)
+                        val = int(val, 0)
                 kwargs[arg_name] = val
             else:
                 arg_name, val = arg, args[i + 1]
@@ -156,9 +156,9 @@ def cli(client):  # noqa: C901 pylint: disable=too-complex
                     if not string:
                         if "," in val:
                             val = val.split(",")
-                            val = [int(v) for v in val]
+                            val = [int(v, 0) for v in val]
                         else:
-                            val = int(val)
+                            val = int(val, 0)
                     kwargs[arg_name] = val
                     skip_index = i + 1
                 except TypeError:
--- a/pymodbus/bit_write_message.py
+++ b/pymodbus/bit_write_message.py
@@ -5,7 +5,7 @@ TODO write mask request/response
 # pylint: disable=missing-type-doc
 import struct
 
-from pymodbus.constants import ModbusStatus
+from pymodbus.constants import Defaults, ModbusStatus
 from pymodbus.pdu import ModbusExceptions as merror, ModbusRequest, ModbusResponse
 from pymodbus.utilities import pack_bitstring, unpack_bitstring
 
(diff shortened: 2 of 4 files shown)


janiversen:
I think you forgotten to update your dev, with Pymodbus dev, before starting this PR. I see at least parts of your previous commit in this commit.

Please update your dev with the latest from Pymodbus and rebase this PR.

Btw it is a good idea to develop in a branch not named dev, because it makes it easier to do:
```
   git checkout dev
   git pull origin
   git checkout my_feature
   git rebase dev
   coding...
   git push 
```
When you develop on dev, you easily run into conflicts with Pymodbus dev (e.g. I updated your commit before merging it).

pkubanek:
Made fix_628 branch..shall I close this PR and start a new one?

janiversen:
It is not a demand from my side, but it might be easier for you.

janiversen:
The CI error is something that is fixed on dev.

janiversen:
Please look at your commit it still contains e.g. the super() change.

pymodbus/repl/client/main.py
```diff
@@ -146,19 +146,19 @@ def _process_args(args, string=True):
                 if not string:
                     if "," in val:
                         val = val.split(",")
-                        val = [int(v) for v in val]
+                        val = [int(v, 0) for v in val]
```
dhoomakethu, line 149: How is this helping?

From the [docs](https://docs.python.org/3/library/functions.html#int) `Base 0 means to interpret exactly as a code literal, so that the actual base is 2, 8, 10, or 16, and so that int('010', 0) is not legal, while int('010') is, as well as int('010', 8).`

dhoomakethu: Can you share some outputs with out this change and with this change please?

janiversen: I did a short review of your code, and I agree with @dhoomakethu it do look strange. Your idea is OK, but you need to work on the implementation (and solve your rebase problem).

janiversen (changes requested):
You need to address the concerns from @dhoomakethu and solve the rebase problem.

pymodbus/repl/client/main.py
pkubanek: int(<str>, 0) takes any sensible value as numerical. The original request was for accepting hex values. int(<str>, 0) accepts any value, with the cost that int('010', 0) throws ValueException - that can be handled by catching it and trying int(<str>), perhaps in a function, as this is on multiple places in the code.

I am not sure about rebasing issues. I discovered problems in my previous PR (#1041), which I fixed with this. Do you want to make an extra PR for improving those?

janiversen: As I already told you, your problems come from the fact that you use dev to submit PRs. You need to submit a PR based on the latest dev, whether you do it in this PR or open a new PR is up to you. I gave you a recipe how to work with dev and branches for each PR.

You need to prove that your PR accept hex values without causing problems. I believe that '010' is accepted today (without having tested it), so it needs to be accepted in this PR as well.

The best way to show your changes works as expected, is to write some test cases in test/test_<something>.py

pkubanek: Ok. Let's close this one, I will make two new PRs, one for fixing problems created by #1041, second to address #628, with still accepting 010 and providing tests to show that.

janiversen: I was not aware that there are any problems due to #1041, I suggest you solve your dev problem first. If there are problems I look forward to see a PR.

#628 is a long standing issue, so it will be nice to have that fixed.

janiversen:
Did you get your dev fixed ? We are still looking forward to see new PRs from you.

pkubanek:
Please see #1075 for new PR.

PR #1075, opened 2022-09-01
Hex values for repl arguments

### LINKED ISSUE #628: Add support for hex values in REPL client.write_registers

pkubanek:
Fix #628. int(x, 0) accept decimal (11), hex (0x11=17), as well as
binary (ob1100=12) and octo (0o11=9) values.

### DIFF
--- a/pymodbus/repl/client/main.py
+++ b/pymodbus/repl/client/main.py
@@ -121,6 +121,51 @@ class NumericChoice(click.Choice):
         return None
 
 
+def _process_args(args: list, string: bool=True):
+    """Internal function to parse arguments provided on command line.
+
+    :param args: Array of argument values
+    :param string: True if arguments values are strings, false if argument values are integers
+
+    :return Tuple, where the first member is hash of parsed values, and second is boolean flag
+        indicating if parsing succeeded.
+    """
+    kwargs = {}
+    execute = True
+    skip_index = None
+
+    def _parse_val(arg_name, val):
+        if not string:
+            if "," in val:
+                val = val.split(",")
+                val = [int(v, 0) for v in val]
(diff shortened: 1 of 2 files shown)


janiversen (approved):
Looks good to me, and the documentation support your claim, it would be nice though to see a couple of test cases.

Since @dhoomakethu if the expert on repl, lets hear his opinion.

pkubanek:
I added tests for _process_args, which I have to move to module (shall not cause an issue). repl itself is sparsely covered by tests (to be decent), but that shall perhaps be another PR (to add testing for repl client and server).

dhoomakethu (approved):

dhoomakethu:
Looks good,  although I would ideally want to have support only for decimal and hex and not support binary or octal as thats going to cause some confusion or not really used.

pkubanek:
Binary values make sense for various flags encountered in Modbus devices. I would leave it that way, binary support is handy, and octal wouldn't make a big difference.

janiversen (approved):
LGTM.

janiversen:
You are right repl does not have a lot of tests, so feel free to add tests. I am trying to generally cleanup the tests (and the code) but I am stuck with the client/server parts.

Feel free to ping me if you want to ping-pong on how to write the tests (I like pytest.mark.parametrice a lot)

PR #1106, opened 2022-10-10
Remove call to configure logging

JonasNorling:
This call configures logging for the whole process, not only the Pymodbus module, so it affects the application using this library.

### DIFF
--- a/pymodbus/constants.py
+++ b/pymodbus/constants.py
@@ -3,17 +3,7 @@
 This is the single location for storing default
 values for the servers and clients.
 """
-import logging
-
 from pymodbus.interfaces import Singleton
-
-
-# set logging format and default level for library.
-logging.basicConfig(
-    format="%(asctime)s %(levelname)-5s %(module)s:%(lineno)s %(message)s",
-    datefmt="%H:%M:%S",
-    level=logging.WARNING,
-)
 
 
 class Defaults(Singleton):  # pylint: disable=too-few-public-methods


janiversen (changes requested):
It is true it affect the whole application, but removing it, is not a good solution since the removal makes it difficult for less experienced users to add logging.

I suppose packaging it as a function and documenting that would be a compromise.

JonasNorling:
I think that users, both less and more experienced, will more likely be surprised that their debug/info log messages disappear if they import pymodbus.constants. At least I was surprised by this — it is common for libraries to use the logger but they don't typically reconfigure it.

IMHO, it's not in the scope of pymodbus to provide functions for configuring logging when the standard logging facility is used. I agree that it would be a good idea to describe in the documentation how to get log messages out of pymodbus, if it's not already mentioned.

tracernz:
This is also listed as a pattern to avoid in the logging cookbook: https://docs.python.org/3/howto/logging-cookbook.html#adding-handlers-other-than-nullhandler-to-a-logger-in-a-library

janiversen:
You might be right with your arguments,  but the solution is not to remove the formatting, the information in the formatting is needed to debug modbus problems. I am open to merge a suggestion that do not remove logging.

The logging cookbook is example patterns not a law, if it was that bad pylint/flake8 would not allow it.

tracernz:
Indeed, and I must say the logging was very useful when taking first steps with pymodbus.

mh-superbox:
Setting `logging.basicConfig()` in `pymodbus.constants import defaults` causes me to get duplicate entries:

```
Python 3.10.4 (main, Jun 18 2022, 13:11:42) [GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging
>>> logger = logging.getLogger("unipi-control")
>>> logger.setLevel(logging.DEBUG)
>>> handler = logging.StreamHandler()
>>> handler.setFormatter(logging.Formatter("%(levelname)s | %(message)s"))
>>> logger.addHandler(handler)
>>> logger.info("TEST")
INFO | TEST
>>> from pymodbus.client import AsyncModbusTcpClient
>>> logger.info("TEST")
INFO | TEST
19:44:24 INFO  <stdin>:1 TEST
```

janiversen:
See the conversation above.

mh-superbox:
> See the conversation above.

I have read the conversation. At the moment this does not change the fact that I get double entries and that this can't be r be overwritten.

JonasNorling:
Closing this PR so that someone else can suggest a solution. Opened bug #1119 instead.

PR #1113, opened 2022-10-14
Bugfix ModbusSerialServer setup so handler is called correctly.

jabjoe:
I was hit by ModbusSerialServer being broken by the previous commit to async_io.py.

There is no test covering ModbusSerialServer which is probably why this got through. One could be added using pty.openpty.

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -841,7 +841,7 @@ class ModbusSerialServer:  # pylint: disable=too-many-instance-attributes
         try:
             self.transport, self.protocol = await create_serial_connection(
                 asyncio.get_event_loop(),
-                self.handler(self),
+                lambda: self.handler(self),
                 self.device,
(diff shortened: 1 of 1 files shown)


janiversen:
Actually there are tests for ModbusSerialServer in test_examples.py

janiversen (approved):
Change looks good.

janiversen:
Thanks for the patch, looking forward to see more work from you.

jabjoe:
> Actually there are tests for ModbusSerialServer in test_examples.py

Ah, I see it now. run_async_server can take "serial" and use StartAsyncSerialServer then. Sorry. I was looking at test_server_asyncio.py which does the others but not StartAsyncSerialServer.

janiversen:
you also have run_sync_server, so we should have it all covered. Apart from that your patch was correct and are merged, thanks again.

PR #1120, opened 2022-10-17
move logging configuration behind function call

### LINKED ISSUE #1119: pymodbus.constants interferes with application's logging configuration
### Versions

* Python: 3.10
* OS: Ubuntu Linux
* Pymodbus: 3.0.0dev5
* Modbus Hardware (if used): 

### Description

pymodbus.constants calls logging.basicConfig() on import, overriding any logging that the application attempts to configure later on. This is unexpected and a documented anti-pattern for Python logging.

There is more discussion in PR #1106, which I will close so that someone else can suggest a solution that the maintainers can agree with.

untzag:
After this change users would need to call `pymodbus.apply_pymodbus_logging_config()` to enable logging as before. This way logs are not injected into my application in an unexpected way.

Happy to change if maintainers wish, or feel free to reject if another option is prefered. A change to this behavior would be really great as my application has carefully formatted logs that will not be parsed correctly due to this injection.

fixes #1119


pymodbus/__init__.py
```diff
@@ -3,17 +3,12 @@
 Released under the the BSD license
 """
 
-import logging as __logging
-from logging import NullHandler as __null
-
 import pymodbus.version as __version
+from ._logging import *
```
janiversen, line 7: hmmm a file named _logging is private and should surely not be imported in __init__.py

Just to be clear my obstruction if the file name.

I am pretty sure this will also cause a pylint problem, the asterisk.

pymodbus/_logging.py
```diff
@@ -0,0 +1,19 @@
+__all__ = ["apply_pymodbus_logging_config"]
```
janiversen, line 1: The name should start with "pymodbus".

pymodbus/_logging.py
```diff
@@ -0,0 +1,19 @@
...
+from logging import NullHandler
+
+
+# ---------------------------------------------------------------------------#
+#  Block unhandled logging
+# ---------------------------------------------------------------------------#
+logging.getLogger(__name__).addHandler(NullHandler())
+
+
+def apply_pymodbus_logging_config():
```
janiversen, line 14: please change function name as well as file name.

Why not include the function directly in __init__ ?

janiversen (changes requested):
Small but important changes. I like the principle of the PR.

untzag:
I personally prefer to keep functionality cleanly separated into different files, which is why I prefer the private file with `import *` pattern. This is just a preference however so I will defer to maintainers :smile:

janiversen:
I have no problem having it a file ! but the file contains a public function so naming it _logging is at least confusing. If you just rename the file (and the function to keep the naming standard) I am happy.

janiversen:
Please solve the pylint/flake8 problems.

janiversen:
Yeah I think that would be a good place to mention this function.

Please write something like:
- Use this function to activate debug in the library in a standard format
- Remark you can activate debug with logging.setlevel("DEBUG") without using this function, but will get another format
- Please use this function if you create an issue and supply a log, since the maintainers need the information to reproduce/verify your problem.

Do not use my words but the idea :-) I am a hard core programmer and have to torture myself to write documentation.

Thanks for taking care of this.

untzag:
OK I added a docstring which should be rendered in autodoc and put mention of the function in a few places that made sense to me. Let me know what you think and if I should add more elsewhere.

janiversen (changes requested):
All good apart from the pylint/flake8 error.

janiversen (approved):
LGTM

PR #1131, opened 2022-10-21
Obey timeout parameter on connection

### LINKED ISSUE #1130: No explicit timeout on client TCP/IP connections
### Versions

* Python: 3.10
* OS: MacOSX
* Pymodbus: https://github.com/riptideio/pymodbus/commit/b4f5a558254f8242fd65fb49fe10f4f255610b32
* Modbus Hardware (if used): any (I'm using a Productivity2000 PLC)

### Pymodbus Specific
* Client: TCP/IP async

### Description

The TCP/IP client connection code does not obey the `timeout` parameter inherited from `ModbusBaseClient` https://github.com/riptideio/pymodbus/blob/b4f5a558254f8242fd65fb49fe10f4f255610b32/pymodbus/client/base.py#L22-L28

This is because it relies on [`loop.create_connection()`](https://docs.python.org/3/library/asyncio-eventloop.html#opening-network-connections), which does not have any overall timeout parameter.  A failed network call will _eventually_ timeout in 45-120s, although I'm not sure which layer actually does it.  (There is only `ssl_handshake_timeout` but this is very specific)

https://github.com/riptideio/pymodbus/blob/b4f5a558254f8242fd65fb49fe10f4f255610b32/pymodbus/client/tcp.py#L82-L89

(PR incoming)

alexrudd2:
Closes #1130

I'm not sure if it's worth having separate timeout parameters for the initial connection and subsequent requests.  This takes the simpler approach

### DIFF
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -105,6 +105,7 @@ class AsyncModbusSerialClient(ModbusBaseClient):
                 bytesize=self.params.bytesize,
                 stopbits=self.params.stopbits,
                 parity=self.params.parity,
+                timeout=self.params.timeout,
                 **self.params.kwargs,
             )
             await self._connected_event.wait()
--- a/pymodbus/client/tcp.py
+++ b/pymodbus/client/tcp.py
@@ -83,8 +83,11 @@ class AsyncModbusTcpClient(ModbusBaseClient):
         """Connect."""
         _logger.debug("Connecting.")
         try:
-            transport, protocol = await self.loop.create_connection(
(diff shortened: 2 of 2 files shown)


janiversen (changes requested):
How about serial  ?

tls uses tcp and udp is connectionless so they are covered.

alexrudd2:
> How about serial ?
> 
> tls uses tcp and udp is connectionless so they are covered.

OK, I made an attempt but cannot test it whatsoever right now.

janiversen (approved):
LGTM. Thanks.

PR #1132, opened 2022-10-21
Add Docker and container registry support 

serroba:
# What?

This PR adds support to create docker images locally against the current working environment
Additionally, this PR is adding the required scripts to automatically create docker images and publish them to public repositories -Docker Hub and GitHub Container Registry (ghcr), as a package.

When publishing to Docker Hub, in this PR, we are publishing images built for `linux/amd64` and `linux/arm64`. Additional platform can be easily expanded if there is more need for that.

# Why?

This can greatly accelerate the exploration of this library. As an external user, the ability to interact with the library and potentially spin up servers and play with the clients can drastically reduce the need to learn the more lower level aspects of the library. This also makes very easy the ability to import the library in a larger project using `docker-compose`.

Additionally, this is a continuation of the conversation I had with @janiversen a few weeks ago: https://github.com/riptideio/pymodbus/commit/4b57def1eea90fb9575b608ac24109c72107dc0c#r86307726

# Testing/Proof

I tested this against my own fork. I tried several times until I got it working. 

The GH actions can be seen here:

https://github.com/serroba/pymodbus/actions

Of particular interest:

* Publishing to Docker Hub: https://github.com/serroba/pymodbus/actions/runs/3297517052/jobs/5438427074
* Publishing as a GitHub package or ghcr: https://github.com/serroba/pymodbus/actions/runs/3297517075/jobs/5438427120

# Container Images Registries

## Docker Hub

To test proper integration with Docker Hub, I created my own repository with the equivalent of this project. The images can be seen on: https://hub.docker.com/repository/docker/serroba/pymodbus

## GitHub Container Registry

To test this, the integration is native, so no need to do anything outside this PR here: https://github.com/serroba/pymodbus/pkgs/container/pymodbus

# Examples and usage with Docker Compose:

I have forked an existent public repository https://github.com/huntabyte/tig-stack and the now created image using this PR

Example:

```
  pymodbus-server:
    container_name: pymodbus-server
    image: serroba/pymodbus:v3.0.0dev5
    command: ["./examples/server_sync.py"]
```

Things are working as expected:
image

Also, we can just things directly in the container:

```
docker run -it serroba/pymodbus:v3.0.0dev5 examples/server_sync.py
```

image

Or enter the REPL using the images:

```
docker run -it serroba/pymodbus:v3.0.0dev5 pymodbus.server run --modbus-server tcp --framer socket --unit-id 1 --unit-id 4 --random 2
```

image

### DIFF
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -23,6 +23,14 @@ jobs:
     steps:
       - name: Check out the repo
         uses: actions/checkout@v3
+
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
       
       - name: Log in to Docker Hub
         uses: docker/login-action@v2
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -12,9 +12,6 @@ name: Publish Docker image
 on:
   release:
     types: [published]
-env:
-  DOCKER_HUB_NAMESPACE: serroba
-  DOCKER_HUB_REPOSITORY: pymodbus
 
 jobs:
   push_to_registry:
@@ -42,7 +39,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v4
         with:
-          images: ${{ env.DOCKER_HUB_NAMESPACE }}/${{ env.DOCKER_HUB_REPOSITORY }}
+          images: ${{ github.repository }}
       
       - name: Build and push Docker image
         uses: docker/build-push-action@v3
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -7,40 +7,32 @@
 # To get a newer version, you will need to update the SHA.
 # You can also reference a tag or branch, but the action may change without warning.
 
-name: Create and publish a Docker image
+name: Publish Docker image
 
 on:
-  push:
-    branches: ['release']
-
-env:
-  REGISTRY: ghcr.io
-  IMAGE_NAME: ${{ github.repository }}
+  release:
+    types: [published]
 
 jobs:
-  build-and-push-image:
+  push_to_registry:
+    name: Push Docker image to Docker Hub
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      packages: write
-
     steps:
-      - name: Checkout repository
+      - name: Check out the repo
         uses: actions/checkout@v3
-
-      - name: Log in to the Container registry
+      
+      - name: Log in to Docker Hub
         uses: docker/login-action@v2
         with:
-          registry: ${{ env.REGISTRY }}
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      
       - name: Extract metadata (tags, labels) for Docker
         id: meta
         uses: docker/metadata-action@v4
         with:
-          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-
+          images: my-docker-hub-namespace/my-docker-hub-repository
+      
       - name: Build and push Docker image
         uses: docker/build-push-action@v3
         with:
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -48,7 +48,7 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           context: .
-          platforms: linux/amd64,linux/arm64,linux/arm/v7
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -12,6 +12,9 @@ name: Publish Docker image
 on:
   release:
     types: [published]
+env:
+  DOCKER_HUB_NAMESPACE: serroba
+  DOCKER_HUB_REPOSITORY: pymodbus
 
 jobs:
   push_to_registry:
@@ -31,7 +34,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v4
         with:
-          images: my-docker-hub-namespace/my-docker-hub-repository
+          images: ${{ env.DOCKER_HUB_NAMESPACE }}/${{ env.DOCKER_HUB_REPOSITORY }}
       
       - name: Build and push Docker image
         uses: docker/build-push-action@v3
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+# syntax=docker/dockerfile:1
+
+FROM python:3.8-slim-buster
+
+WORKDIR /app
+
+COPY requirements.txt requirements.txt
+RUN pip3 install -r requirements.txt
+
+COPY . .
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,9 @@
 
 FROM python:3.8-slim-buster
 
-WORKDIR /app
-
-COPY requirements.txt requirements.txt
-RUN pip3 install -r requirements.txt
+WORKDIR /pymodbus
 
 COPY . .
+
+RUN pip3 install -r requirements.txt && pip3 install -e .
+
--- a/README.rst
+++ b/README.rst
@@ -197,6 +197,10 @@ Or to install a specific release:
 
     pip install -U pymodbus==X.Y.Z
 
+You can also use Docker to run a local image with the package installed on the image:
+
+    docker pull riptideio/pymodbus
+
 Otherwise you can pull the trunk source and install from there::
 
     git clone git://github.com/riptideio/pymodbus.git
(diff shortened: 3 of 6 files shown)


.github/workflows/docker-hub.yml
```diff
@@ -0,0 +1,51 @@
...
+        uses: docker/setup-qemu-action@v2
+
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
```
serroba, line 35: this requires project admin privileges. Is a matter of creating an account in docker hub, free project is sufficient.

We can add collaborators in Docker Hub, so a dedicated account can be created rather than the primary one.

Once the account has been created, is just a matter of copy&paste credentials as part of the secrets of the project

.github/workflows/docker-hub.yml
```diff
@@ -0,0 +1,51 @@
...
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ github.repository }}
```
serroba, line 42: This will try to publish to `docker.com` using the `owner/project-name` format.

If merged, this would take the form of: `riptideio/pymodbus`

.github/workflows/docker-hub.yml
```diff
@@ -0,0 +1,51 @@
...
+
+# GitHub recommends pinning actions to a commit SHA.
+# To get a newer version, you will need to update the SHA.
+# You can also reference a tag or branch, but the action may change without warning.
+
+name: Publish Docker image
+
+on:
+  release:
+    types: [published]
```
serroba, line 14: this only create and publish images if a release has been created.

We could change this to create images on other actions, but I thought this would be a nice minimum action to have as a starting point at least.

.github/workflows/docker-hub.yml
```diff
@@ -0,0 +1,51 @@
...
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ github.repository }}
+      
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
```
serroba, line 48: https://hub.docker.com/layers/serroba/pymodbus/v3.0.0dev10/images/sha256-9d163cd8b9fba5634d372841a67f3b5aff42806425217339017bdc8bdb29838c?context=repo

An example of multiple platforms working as expected:

image

janiversen (changes requested):
Without having actually tried it but only reviewed it, it looks good.

However you need to add documentation, so that users know it exist.

If I looked right you need me to add a user/password to secrets (if my powers allow it) anything else?

janiversen:
Added DOCKER_USERNAME / DOCKER_PASSWORD to secrets. Username is janiversen (which you need for documentation).

Can you make the action so I e.g. can create docker images with some of the examples ?

Dockerfile
```diff
@@ -0,0 +1,10 @@
+# syntax=docker/dockerfile:1
+
+FROM python:3.8-slim-buster
+
+WORKDIR /pymodbus
+
+COPY . .
+
+RUN pip3 install -r requirements.txt && pip3 install -e .
+
```
janiversen, line 10: This only installs pymodbus but does nothing.

You need to spin up the repl server, so that a user can use the docker image as a virtual device. Or have I missed something ?

janiversen (changes requested):

Dockerfile
serroba: That's correct. This isn't meant to be only a repl interface. Is up to the client to decide what to do with the image. In my case I'm happy spinning up the example server, I'm not after the repl

serroba:
@janiversen the only thing required was the creation of those credentials. Thank you for adding them to the secrets. Nothing else is required on your end.

WRT documentation, one thing is updating the readme, which I'm happy to do.

> Added DOCKER_USERNAME / DOCKER_PASSWORD to secrets. Username is janiversen (which you need for documentation).

I'm not sure what you mean here. Do you mean adding comments to the YML files where those values are used?


> Can you make the action so I e.g. can create docker images with some of the examples ?

I don't understand what you mean. Can you please clarify. Thanks

Dockerfile
janiversen: But if you only install pymodbus the image the users cannot use the image before they have added extra functionality and that seems to complicate the usage.

I like the idea of a docker image, but it needs to be something that really helps a user...If you had a complete development environment or a server running it would help the user.

janiversen:
You need to document in the README that there are a docker image available, where and how to use it.

janiversen:
I was thinking of having e.g, docker images with: a standard server, a repl server and a forwarder.

serroba:
Users can create a `docker compose` file and add a `command` instruction. That allows to use the same image with a user-selected command 

The example I showed is this:

```yml
pymodbus-server:
    container_name: pymodbus-server
    image: serroba/pymodbus:v3.0.0dev5
    command: ["./examples/server_sync.py"]
```

in this case, the container is spin up and we run the sync server. A different person could add the REPL in place instead.

.github/workflows/docker-hub.yml
serroba: @janiversen I don't think we need to document the concrete user you have created, unless you want this to be known by others in the public (I don't recommend that).
Otherwise, I would recommend avoid any specific reference to a given user. Is something the administrators of the package can decide to manage privately

serroba: replying to this https://github.com/riptideio/pymodbus/pull/1132#issuecomment-1287645811

.github/workflows/docker-hub.yml
janiversen: hmmmm the username I gave you is NOT riptideio on purpose and I would not like my user to create the riptideio organisation that is something for members of riptideio to do.

That is why I suggested to use janiversen/pymodbus

.github/workflows/docker-hub.yml
janiversen: that more than enough.

Dockerfile
serroba: The REPL requires a long running session, is not a designed as server. You can manually do it, if you run
`docker exec -it`, which opens an interactive terminal. But you can't, at least, not to my knowledge, spin up the REPL as a canonical docker command.

If you know how to spin up the REPL by default, happy to add to the Dockerfile, clients can override the default run command regardless, so I don't mind if that's your preference. But I don't think it's possible to do what you mentioned.

janiversen:
your example of compose is good, but that is what needs to be documented.

README.rst
```diff
@@ -197,6 +197,10 @@ Or to install a specific release:
 
     pip install -U pymodbus==X.Y.Z
 
+You can also use Docker to run a local image with the package installed on the image:
+
+    docker pull riptideio/pymodbus
```
janiversen, line 202: I am not sure that is correct, with the user I added as secret owner will be janiversen.

pymodbus/repl/client/README.md
```diff
@@ -32,7 +38,9 @@ Commands:
   serial
   tcp
 
+# Or using a Docker run command instead
 
+✗ docker run -it riptideio/pymodbus pymodbus.console --help
```
janiversen, line 43: And how about the compose options you wrote about earlier.

janiversen (changes requested):

README.rst
serroba: oh... ideally the docker namespace is the same as the one in github. 

If that's not possible sure, I can update to use your own account. But it creates a sense of discomfort when the namespace in docker is different to the "vendor"/package-namespace.

I can update though if you think that's the way to go. Your call.

serroba: To be clear, docker hub allows multiple collaborators for the same project/repository.

Having your user as a credential is fine and doesn't necessarily implies that the docker hub has to be under `janiversen`. You could still have `janiversen` as a username under the `riptideio` namespace

pymodbus/repl/client/README.md
serroba: The compose example does not applies to the REPL, but I can add a particular example to run the REPL using docker

README.rst
```diff
@@ -236,6 +240,23 @@ This installs pymodbus in your virtual environment with pointers directly to the
...
+
+If you would like to use this image as part of a `docker compose` project, you can provide a custom command. For example,
+you can spin the server by creating a docker compose file containing the following:
+
+```yml
+pymodbus-server:
+    container_name: pymodbus-server
+    image: riptideio/pymodbus:X.Y.Z
+    command: ["./examples/server_sync.py"]
+```
```
serroba, line 255: @janiversen what do you think about this now? is it better explained how this can customised?

janiversen:
CI does not like your text:
```
../README.rst:252: ERROR: Unexpected indentation.
[628](https://github.com/riptideio/pymodbus/actions/runs/3303097490/jobs/5450746414#step:5:629)
../README.rst:250: WARNING: Inline literal start-string without end-string.
[629](https://github.com/riptideio/pymodbus/actions/runs/3303097490/jobs/5450746414#step:5:630)
../README.rst:250: WARNING: Inline interpreted text or phrase reference start-string without end-string.
[630](https://github.com/riptideio/pymodbus/actions/runs/3303097490/jobs/5450746414#step:5:631)
../README.rst:255: WARNING: Block quote ends without a blank line; unexpected unindent.
[631](https://github.com/riptideio/pymodbus/actions/runs/3303097490/jobs/5450746414#step:5:632)
../README.rst:255: WARNING: Inline literal start-string without end-string.
[632](https://github.com/riptideio/pymodbus/actions/runs/3303097490/jobs/5450746414#step:5:633)
../README.rst:255: WARNING: Inline interpreted text or phrase reference start-string without end-string.
```

janiversen (approved):
LGTM.

PR #1134, opened 2022-10-22
removes undue reported error when forwarding

alexandrecuer:
the first commit solves a cosmetic problem mentionned in the discussion : https://github.com/riptideio/pymodbus/discussions/1110#discussioncomment-3871293

it removes the following undue error, reported when running a TCP server forwarding a serial server :
```
17:10:21 ERROR mixin:113 Please do not use unit=, convert to slave=.
````

the second commit is a suggestion of homogeneisation in register_read_message.py


janiversen (approved):
Good catch! thanks.

PR #1135, opened 2022-10-22
server/asyncio.py change in the _log_exception method to correct the logging when disconnecting the socket

alexandrecuer:
in the _log_exception method :  
- all logging should be debug and not error
- no logging should be thrown if the handler is not running, otherwise cancellation messages are duplicated, we see the first cancellation message at the closing stage and the same is shown when we enter the closed state

this answers to https://github.com/riptideio/pymodbus/discussions/1110#discussioncomment-3871519

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -100,11 +100,13 @@ class ModbusBaseRequestHandler(asyncio.BaseProtocol):
 
     def _log_exception(self):
         """Show log exception."""
+        if not self.running:
+            return
         if isinstance(self, ModbusConnectedRequestHandler):
             txt = f"Handler for stream [{self.client_address[:2]}] has been canceled"
-            _logger.error(txt)
+            _logger.debug(txt)
         elif isinstance(self, ModbusSingleRequestHandler):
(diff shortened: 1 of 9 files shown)


pymodbus/server/async_io.py
```diff
@@ -100,11 +100,13 @@ def __init__(self, owner):
 
     def _log_exception(self):
         """Show log exception."""
+        if not self.running:
+            return
```
janiversen, line 104: Log exception can occur before and especially after server is running.

janiversen (changes requested):

pymodbus/server/async_io.py
alexandrecuer: The _log_exception() is only called twice, in handle() when self.running is True and in connection_lost(), so always after server is running. But maybe i missunderstand something

janiversen: _log_exception can be called from anywhere, we add/remove _log_exception often. Apart from that it is not logical nor correct that _log_exception depends on the state of the server.

janiversen (approved):
LGTM. Thanks.

PR #1137, opened 2022-10-23
Dev multiple server

janiversen:

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -2,6 +2,7 @@
 # pylint: disable=missing-type-doc
 import asyncio
 import logging
+import platform
 import ssl
 import traceback
 from binascii import b2a_hex
@@ -956,6 +957,15 @@ class _serverList:
             await self.task
         except asyncio.CancelledError:
             pass
+        if platform.system().lower() == "windows":
+            owntask = asyncio.current_task()
+            for task in asyncio.all_tasks():
+                if task != owntask:
+                    task.cancel()
+                try:
+                    await task
+                except asyncio.CancelledError:
+                    pass
         self.i_am_stopped.set()
 
     def request_stop(self):
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -34,12 +34,6 @@ except ImportError:
 # --------------------------------------------------------------------------- #
 _logger = logging.getLogger(__name__)
 
-# --------------------------------------------------------------------------- #
-# Allow access to server object, to e.g. make a shutdown
-# --------------------------------------------------------------------------- #
-_server_stopped = None  # pylint: disable=invalid-name
-_server_stop = None  # pylint: disable=invalid-name
-
 
 def sslctx_provider(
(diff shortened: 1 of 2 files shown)


dhoomakethu:
My personal opinion, any new code should have `type hints` by default. Makes it easier to understand code. Also would be nice to have `index` instead of `inx` or `server_jobs` instead of `my_jobs` which makes it easier to read and get a context of what these variables are for.

janiversen:
I agree that all APi should have type hints, but it is pure PITA to maintain it for internal stuff. Another  reason not to do it internally is the sometimes very wicked arguments used, I presume the old support for twisted/tornado caused that.

As an example of this “strangeness” at API level, just look at client/mixin.py all those functions are not declared async although they do return a future in case of async. To be clear I find the solution rather efficient instead of demanding type hints.

Variable names are something that can always be better…You have maintained rr and rq for a very long time. I will change a bit in the naming.

janiversen:
speaking of un-maintained stuff, you created http://riptideio.github.io/pymodbus/ in 2018, but it seems very un-maintained !

janiversen:
hmmm cannot reopen this one.

dhoomakethu:
There are many things that I could not maintain or keep up to do date , the github page you shared  should have an active branch and the idea was to keep updating it on new changes automatically with the ci pipeline. Since we got the read the docs in place it was no longer maintained.

janiversen:
Which is a reason to unpublish it (just a hit of one button), and not just leave it outdated. I have now updated it.

dhoomakethu:
I don't know what are you getting at. I did what I could do with the time and opportunity I had. In any case thanks for taking care of it.

janiversen:
Nothing special.  I am just a bit tired of this constant cleanup and at the same time feeling that the unspoken aim of this project is to minimize maintenance and changes.

Maybe it is better for the project if I stop trying, and instead make a fork and continue the renewal process there….at least that way you will have peace.

janiversen:
"I did what I could do with the time and opportunity I had", that is a fair statement.

I just unpublished the site and removed the branch, total time usage less than 30 seconds.

dhoomakethu:
@janiversen seriously you are drifting this conversation in to something else which is not at all related to the original comment I made on this PR. Please look in to the recent commits I have made both for REPL and you will find that I have made enough efforts to have type-hints in place. There are many things that could still be better. But your attitude does not go right with the decorum with which one has to communicate in an online forum with a person who is not personally known to you.

dhoomakethu:
If you are having so much problem with this library, I honestly request you please fork and maintain your own version and do what so ever you want to do with it. I will retire this library for good.

janiversen:
If you seriously think about retiring the library, then please do a transfer of the repo instead, that way all forks link etc are automatically moved.

Should you decide to do so, you will of course get “maintainer “ status immediately  together with any from riptideio who wants it.

But please consider it carefully, to me the difference between a fork and the master is just a matter of the responsibility.

and please accept my apologies if I somehow offended you with my words that was not my intention, being Danish I am used to speak my heart.

PR #1147, opened 2022-10-27
Fix type hints for port and timeout

### LINKED ISSUE #1142: Minimum client timeout 1s?
### Versions

* Python: 3.11
* OS: macOS
* Pymodbus: 3.0
* Modbus Hardware (if used):

### Pymodbus Specific
* Server: tcp - sync
* Client: tcp - sync

### Description

The minimum timeout that can be set on a client is one seconds as the timeout parameter is an int and at 0 it does not connect. Wouldn't it be better to have timeout as a float to be able to specify tighter timeouts especially for TCP clients?

### Code and Logs

None

alexrudd2:
8b24d397427079afb1eeb7d902b20f9754d83d56 had an error.  `port` should have remained `int`

  Closes #1142

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -25,7 +25,7 @@ class ModbusBaseClient(ModbusClientMixin):
     **Parameters common to all clients**:
 
     :param framer: (optional) Modbus Framer class.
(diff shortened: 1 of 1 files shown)


janiversen (approved):
LGTM.

PR #1151, opened 2022-10-30
Fix unreachable code in AsyncModbusTcpClient

### LINKED ISSUE #1150: Unreachable code in AsyncModbusTcpClient
### Versions

* Pymodbus: 3.0.0

### Pymodbus Specific
* Client: tcp - async

### Description

Unreachable code in else statement due to the early return in the try statement.
This causes that there is no logging message being reported and to never reset the delay if there would be due to a temporary disconnect.

See also `_connect` function of the async TCP client 
https://github.com/riptideio/pymodbus/blob/61d9e9ac20954634a453befae50576da2333e40b/pymodbus/client/tcp.py#L82-L100

WouterTuinstra:
Fixes #1150 Unreachable code in else statement due to the early return in the try statement.

### DIFF
--- a/pymodbus/client/tcp.py
+++ b/pymodbus/client/tcp.py
@@ -89,7 +89,6 @@ class AsyncModbusTcpClient(ModbusBaseClient):
                 ),
                 timeout=self.params.timeout,
             )
-            return transport, protocol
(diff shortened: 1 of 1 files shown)


janiversen (approved):
Good catch

PR #1154, opened 2022-10-30
Avoid unwanted reconnects

wlcrs:
Fixes a possible race condition resulting in an unwanted reconnect due to `protocol_lost_connection` that would do a reconnect if `params.host` is still filled in right after doing the disconnect of the protocol-class.

The order of these statements is already ok in `udp.py`.


janiversen (approved):
LGTM.

janiversen:
Please solve the flake8 and black problem.

janiversen:
Solved black for you.

wlcrs:
I guess this solved the flake8 issues too? I don't see any of them when I check locally.

janiversen (approved):
LGTM. Thanks.

PR #1157, opened 2022-10-31
Server simulator with datastore with json data

janiversen:


janiversen:
Just a test to see if it works on all platforms.

PR #1162, opened 2022-11-04
Fix typo.

printfinn:
Not sure if it's a typo, is it supposed to be Redis? I'm not aware of Redit in Python, but if we do have Redit, please ignore this PR.


janiversen (approved):
LgTM. Thanks

PR #1170, opened 2022-11-05
Add pygments as requirement for repl.

janiversen:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,6 @@
 version 3.0.2
 ----------------------------------------------------------
+* Add pygments as requirement for repl
 * Update datastore remote to handle write requests (#1166)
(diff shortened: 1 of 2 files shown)


janiversen:
@dhoomakethu I added this, hope it is ok.

dhoomakethu:
Earlier `repl` and it's dependencies are installed with `pip install pymodbus[repl]` which should have installed all the requirements. These optional requirements for repl were not installed as part of the normal pymodbus install. Just want to make sure if the same is still being followed or is it going to be installed with normal installation (`pip install pymodbus`)

janiversen:
you can still install with ```pip install pymodbus[repl]``` however even though repl/client/main.py imports pygments it was not in requirements.txt, which was not ok.

dhoomakethu:
Thank you

PR #1180, opened 2022-11-09
Do not install run dependencies for linters

alexrudd2:
The current CI takes a long time (~40 mins of GH Actions runtime).  Although it's free and parallelized I think this could still be improved.

`tox` is a tool for code testing, but it's overkill for linting source code with e.g. black, codespell...  I've simplified the install for each linter to just its own package.  (It may be best to use a simpler action to just do `git checkout` + run the linter, but I don't want to make too major changes. )

This significantly improves the linter runtimes:
- pylint broke so I left it alone :P
- codespell -> 60% faster
- bandit -> 50% faster
- flake8 -> 60% faster
- isort -> 40% faster
- black -> 20% faster

### DIFF
--- a/bandit.yaml
+++ b/bandit.yaml
@@ -1,5 +1,9 @@
 # https://bandit.readthedocs.io/en/latest/config.html
-
+exclude_dirs:
+  - doc
+  - .venv
+  - .git
+  - .tox
 tests:
   - B103
   - B108
--- a/setup.cfg
+++ b/setup.cfg
@@ -476,8 +476,7 @@ check-str-concat-over-line-jumps=no
 #max-line-length-suggestions=
 
 [flake8]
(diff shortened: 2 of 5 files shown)


janiversen:
pylint probably broke because you did not set the command line arguments used in tox.ini

janiversen:
Now that I got a chance to look at it a bit deeper, I can see you state one thing in the title "do not use tox for....", but you change tox.ini so actually we continue to use tox.

pylint needs the dependencies because it controls the source and ensures e.g. that calls to external library functions can be resolved.

This PR is quite good, but it would be a lot nicer to avoid tox in CI, that is surely a couple of minutes saved.

I agree that your consumption in CI is important, but the worst task in the job is pypy, it is build every time, that should be either cached or we should use a container with pypy prebuilt.

PR #1185, opened 2022-11-11
Serial receive incomplete bytes.issue#1183

### LINKED ISSUE #1183: Function code 0x15 fails because of incomplete receival of bytes
### Versions

* Python:3.10.7
* OS:Win 10
* Pymodbus:3.0.2
* Modbus Hardware (if used):USB to TTL(CH340)

### Pymodbus Specific
* Server: rtu - sync
* Client: rtu - sync

### Description

We were trying to send many bytes data to slave by function code 0x15. Finally, we found that serial received data is incomplete, especially when the baud rate is low. In function recv of pymodbus/client/serial.py, when size if not None, serial does not wait data.

### Code and Logs

```python
# code and logs here.

from pymodbus.client import ModbusSerialClient as ModbusClient
from pymodbus.file_message import FileRecord, WriteFileRecordRequest

import os
import argparse

parser = argparse.ArgumentParser(description='Firmware update on MODBus RTU.')

parser.add_argument('-f',
                    action='store',
                    type=str,
                    metavar='',
                    dest='file',
                    help='Full path and filename of the file to be downloaded. '
                         'Example: -f C:\git\Firmware_Download\my_file.bin')

args = parser.parse_args()

client = ModbusClient(method='rtu',port ='COM8', parity = 'E', baudrate=9600, timeout=5)
client.connect()

file_len = os.path.getsize(args.file)
record_len = file_len // 2
record_len = record_len if record_len < 121 else 121

file = open(args.file, 'rb')
file_data = file.read(record_len * 2)

frame = [FileRecord(reference_type=0x06, file_number=0x01, record_number=0, 
                 record_data=file_data, record_length=record_len)]
rq =  WriteFileRecordRequest(records=frame, unit=1)
print(rq.encode())
rr = client.execute(rq)

client.close()
```
When I modify the recv in this way, there is no problem. I don't know if there will be any other problems with this change. I hope you can help me review. Thank you.
```python
    def recv(self, size):
        """Read data from the underlying descriptor."""
        super().recv(size)
        if not self.socket:
            raise ConnectionException(
                self.__str__()  # pylint: disable=unnecessary-dunder-call
            )
        if size is None:
            size = self._wait_for_data()
        elif size > self._in_waiting():
            self._wait_for_data()
        result = self.socket.read(size)
        return result
```

banana-sun:
In function recv of pymodbus/client/serial.py, when size if not None, serial does not wait data.

fixes #1183

### DIFF
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -304,6 +304,8 @@ class ModbusSerialClient(ModbusBaseClient):
             )
         if size is None:
             size = self._wait_for_data()
+        elif size > self._in_waiting():
+            self._wait_for_data()
         result = self.socket.read(size)
         return result
 
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -83,10 +83,12 @@ class mockSocket:  # pylint: disable=invalid-name
     def __init__(self):
         """Initialize."""
         self.data = None
+        self.in_waiting = 0
 
     def mock_store(self, msg):
         """Store message."""
         self.data = msg
+        self.in_waiting = len(self.data)
 
     def mock_retrieve(self, size):
         """Get message."""
@@ -97,6 +99,7 @@ class mockSocket:  # pylint: disable=invalid-name
         else:
             retval = self.data[0:size]
         self.data = None
+        self.in_waiting = 0
         return retval
 
     def close(self):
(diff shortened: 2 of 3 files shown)


janiversen (changes requested):
Seems the change breaks most of the tests.

janiversen (approved):
Added commit to solve problems.

Issue #1203, opened 2022-11-28
AsyncModbusTcpClient ignores reconnect_delay == 0 


Cavaler:
pymodbus 3.0.2

Documentation says:
> reconnect_delay doubles automatically with each unsuccessful connect. Set reconnect_delay=0 to avoid automatic reconnection.

However, creating AsyncModbusTcpClient(reconnect_delay=0) just causes it to reconnect immediately.

I would suggest checking for delay_ms in protocol_lost_connection() instead of host. Same in connect() on error.
And also set delay_ms to 0 in close(), this would allow calling connect() again on previously close()'s client.

Cavaler:
--- tcp.old	2022-11-28 12:22:03.442917900 +0200
+++ tcp.py	2022-11-28 18:58:22.041898200 +0200
@@ -69,7 +69,7 @@
         """Stop client."""
 
         # prevent reconnect.
-        self.params.host = None
+        self.params.delay_ms = 0
 
         if self.connected and self.protocol and self.protocol.transport:
             self.protocol.transport.close()
@@ -104,7 +104,8 @@
         except Exception as exc:  # pylint: disable=broad-except
             txt = f"Failed to connect: {exc}"
             _logger.warning(txt)
-            asyncio.ensure_future(self._reconnect())
+            if self.params.delay_ms > 0:
+                asyncio.ensure_future(self._reconnect())
         else:
             txt = f"Connected to {self.params.host}:{self.params.port}."
             _logger.info(txt)
@@ -130,7 +131,7 @@
 
         self.connected = False
         self.protocol = None
-        if self.params.host:
+        if self.params.delay_ms > 0:
             asyncio.ensure_future(self._reconnect())
 
     async def _reconnect(self):

janiversen:
Nice enhancement, would you mind submitting a pull request ?

janiversen:
Thanks again for the PR, its merged to dev, and will be available in version 3.1

PR #1205, opened 2022-11-28
AsyncModbusTcp/UdpClient honors delay_ms == 0 (#1203)

Cavaler:

### DIFF
--- a/pymodbus/client/udp.py
+++ b/pymodbus/client/udp.py
@@ -88,7 +88,7 @@ class AsyncModbusUdpClient(ModbusBaseClient):
         :meta private:
         """
         # prevent reconnect:
-        self.params.host = None
+        self.delay_ms = 0
 
         if self.connected and self.protocol and self.protocol.transport:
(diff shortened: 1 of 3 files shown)


pymodbus/client/udp.py
```diff
@@ -158,7 +158,7 @@ def protocol_lost_connection(self, protocol):
 
             self.connected = False
             self.protocol = None
-            if self.params.host:
```
Cavaler, line 161: I don't have any UDP server, so haven't tried it

janiversen: I think it is covered by our test suite.

janiversen: and just for information, pymodbus have an excellent server, which are very good for testing cases like this.

janiversen (approved):
LGTM, lets see what CI says.

janiversen:
Thanks for the patch, would be nice if you find other things to submit.

PR #1222, opened 2022-12-12
Clean client class structure.

janiversen:
Avoid making a second object in create_protocol, and move common code into the base classes.

RIght now modbusXclient contains a lot of code that are identical across protocols…move to base classes

Crreate_protocol creates new object modbusxclient protocol…reuse self.


janiversen:
Current status:
Still in the early stages.

PR #1221, opened 2022-12-12
Solve stop/start problems of some client/server combinations.

janiversen:
as #1207 points out, there are a problem when stopping a client/server in some combinations.

add a new set of tests (test_server_task.py) that test 3 different scenarios:
- Start server then close (server is in listen mode)
- Start server then Start/read/stop client then stop server
- Start server then Start/read client then stop server and finally stop client
This is done for all protocols serial/tcp/tls/udp
and for sync as well as async.

Since this seems to be a timing problem all tests will be run 10 times to provoke intermittent issues.

fixes #1207
fixes #1202
fixes #1198

PR #1225, opened 2022-12-12
add check_ci.sh

janiversen:


janiversen:
Disable test_examples.py until PR #1221 is merged.

PR #1229, opened 2022-12-13
add 'change_rate' randomization option

hahattan:

### DIFF
--- a/pymodbus/repl/server/main.py
+++ b/pymodbus/repl/server/main.py
@@ -170,7 +170,7 @@ def run(
         0,
         "--change-rate",
         "-c",
-        help="Chance of randomizing reads. 0=never, 100=always, 1=1%"
+        help="Rate in % registers to change. 0=none, 100=all, 12=12% of registers"
         ", and so on. Applicable IR and DI.",
     ),
 ):
--- a/pymodbus/repl/server/main.py
+++ b/pymodbus/repl/server/main.py
@@ -166,6 +166,13 @@ def run(
         help="Randomize every `r` reads. 0=never, 1=always,2=every-second-read"
         ", and so on. Applicable IR and DI.",
     ),
+    change_rate: int = typer.Option(
+        0,
+        "--change-rate",
+        "-c",
+        help="Chance of randomizing reads. 0=never, 100=always, 1=1%"
+        ", and so on. Applicable IR and DI.",
+    ),
 ):
     """Run Reactive Modbus server.
 
@@ -195,6 +202,7 @@ def run(
 
     modbus_config["handler"] = handler
     modbus_config["randomize"] = randomize
+    modbus_config["change_rate"] = change_rate
     app = ReactiveServer.factory(
         modbus_server,
         framer,
--- a/pymodbus/server/reactive/main.py
+++ b/pymodbus/server/reactive/main.py
@@ -113,6 +113,7 @@ class ReactiveModbusSlaveContext(ModbusSlaveContext):
         holding_registers: BaseModbusDataBlock = None,
         zero_mode: bool = False,
         randomize: int = 0,
+        change_rate: int = 0,
         **kwargs,
     ):
         """Reactive Modbus slave context supporting simulating data.
@@ -123,6 +124,8 @@ class ReactiveModbusSlaveContext(ModbusSlaveContext):
         :param holding_registers: Holding registers data block
         :param zero_mode: Enable zero mode for data blocks
         :param randomize: Randomize reads every <n> reads for DI and IR,
+                          default is disabled (0)
+        :param change_rate: Chance of randomizing reads for DI and IR,
                           default is disabled (0)
         :param min_binary_value: Minimum value for coils and discrete inputs
         :param max_binary_value: Max value for discrete inputs
@@ -142,6 +145,9 @@ class ReactiveModbusSlaveContext(ModbusSlaveContext):
         min_register_value = kwargs.get("min_register_value", 0)
         max_register_value = kwargs.get("max_register_value", 65535)
         self._randomize = randomize
+        self._change_rate = change_rate
+        if self._randomize > 0:
+            self._change_rate = 0
         self._lock = threading.Lock()
         self._read_counter = {"d": 0, "i": 0}
         self._min_binary_value = min_binary_value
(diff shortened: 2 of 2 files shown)


janiversen:
Let’s see what CI thinks

hahattan:
Sry just fixed on of the CI error, can you trigger again ?

janiversen:
if you have the newest dev, you can run the checks locally with ./check_ci.sh

I am off for the next 8-10 hours.

hahattan:
any idea why the CI / black is failing ?

pymodbus/server/reactive/main.py
```diff
@@ -177,6 +183,20 @@ def getValues(self, fc_as_hex, address, count=1):
                     #              "'%s'", _block_type, values, address)
                     self.store[_block_type].setValues(address, values)
                 self._read_counter[_block_type] += 1
+        elif self._change_rate > 0 and _block_type in {"d", "i"}:
+            for offset in range(count):
+                if random.randint(1, 100) <= self._change_rate:
```
dhoomakethu, line 188: @hahattan Thanks for this PR, would you mind giving a short description with examples (if possible) how does this work?

hahattan: Hi, 

So I have a use case that an client is reading datas from repl-server and sending them to somewhere else if the value has changed.
I was using the `-r/--random` to simulate this scenario but the result is not what I want:
For example if I'm reading 1000 registers and expecting their values have 20% chance being updated. I used `--random 5`, the expected data sent by client would be `0,0,0,0,1000,0,0,0,0,1000 ....`. (what I want is something like `200,200,200,200,200...`)

pymodbus/repl/server/main.py
```diff
@@ -166,6 +166,13 @@ def run(
         help="Randomize every `r` reads. 0=never, 1=always,2=every-second-read"
         ", and so on. Applicable IR and DI.",
     ),
+    change_rate: int = typer.Option(
+        0,
+        "--change-rate",
+        "-c",
+        help="Chance of randomizing reads. 0=never, 100=always, 1=1%"
```
janiversen, line 173:
```suggestion
        help="Rate in % of registers to change. 0=none, 100=all, 12=12% of registers"
```
I believe this is a better help.

pymodbus/server/reactive/main.py
dhoomakethu: So if I understand this feature correctly, you are trying to return random values for a percentage of input registers/ discrete inputs being read. If I am reading 10 registers and have set `-c=20` then only 2 out of the 10 registers are expected to change the value, even though I have passed `--random=1` which means randomise values for every read. Is my understanding correct?

pymodbus/repl/server/main.py
dhoomakethu: IMHO, `--change-rate` ,`--probability` or simply `--likely` are all OK.

pymodbus/server/reactive/main.py
hahattan: Not exactly, I made `--random` overwrite `--change-rate` because `--random` IMO is more predictable:
```python
        if self._randomize > 0:
            self._change_rate = 0
```

So if user provide both `-r` and `-c` only `-r` will work

dhoomakethu: Got it, so if `-r` is not passed , then my understanding of how `-c` should works is good?

hahattan: Yes

dhoomakethu: In any case how does `if random.randint(1, 100) <= self._change_rate:` ensure that we are going to change only the required percentage of registers/discrete inputs being changed. `random.randint` is not deterministic and there might be cases where the number of registers being changed are above and below the required count. Is this acceptable?

How about calculating and rounding off the number of registers to be changed and randomly update values?

```
elif self._change_rate > 0 and _block_type in {"d", "i"}:
    changed_indicies = []
    regs_to_changes = round(count * self._change_rate/100)
    random_indices = random.sample(count, regs_to_changes)
    for offset in random_indices:
        with self._lock:
            # Update values
            if _block_type == "d":
                min_val = self._min_binary_value
                max_val = self._max_binary_value
            else:
                min_val = self._min_register_value
                max_val = self._max_register_value
            self.store[_block_type].setValues(
                address + offset, random.randint(min_val, max_val)
```

hahattan: > and there might be cases where the number of registers being changed are above and below the required count. Is this acceptable?

Yes, this is acceptable for me. Sorry for misunderstanding what you meant in previous comment.

> 
> How about calculating and rounding off the number of registers to be changed and randomly update values?
> 
> ```
> elif self._change_rate > 0 and _block_type in {"d", "i"}:
>     changed_indicies = []
>     regs_to_changes = round(count * self._change_rate/100)
>     random_indices = random.sample(count, regs_to_changes)
>     for offset in random_indices:
>         with self._lock:
>             # Update values
>             if _block_type == "d":
>                 min_val = self._min_binary_value
>                 max_val = self._max_binary_value
>             else:
>                 min_val = self._min_register_value
>                 max_val = self._max_register_value
>             self.store[_block_type].setValues(
>                 address + offset, random.randint(min_val, max_val)
> ```

I'm okay with this solution as well, if there's no other comment I can push the commit with suggested changes.

dhoomakethu: Yes, if you are OK please use the suggested changes. Thanks

janiversen (approved):
Nice change !

pymodbus/server/reactive/main.py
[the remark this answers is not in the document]
hahattan: I push a new commit per @dhoomakethu's suggestion and I think that solved the problem

pymodbus/server/reactive/main.py
```diff
@@ -385,13 +405,15 @@ def create_context(
...
         randomize: int = 0,
+        change_rate: int = 0,
     ):  # pylint: disable=dangerous-default-value
         """Create Modbus context.
 
         :param data_block_settings: Datablock (dict) Refer DEFAULT_DATA_BLOCK
         :param unit: Unit id for the slave
         :param single: To run as a single slave
         :param randomize: Randomize every <n> reads for DI and IR.
+        :param change_rate: Rate in % of registers to change for DI and IR.
```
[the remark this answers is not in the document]
hahattan: I push a new commit per @dhoomakethu's suggestion and I think that solved the problem

pymodbus/server/reactive/main.py
```diff
@@ -142,6 +145,9 @@ def __init__(
         min_register_value = kwargs.get("min_register_value", 0)
         max_register_value = kwargs.get("max_register_value", 65535)
         self._randomize = randomize
+        self._change_rate = change_rate
+        if self._randomize > 0:
```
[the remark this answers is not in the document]
hahattan: After the new commit, do you have any suggestion on how to handle the case that user provide both flags ?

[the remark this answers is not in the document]
janiversen: My personal preference is to terminate with an error message, but at the very least the user needs to be aware that the -c is voided.

janiversen:
The failing test is not due to your code, it is a CI problem.

I merge this now, please make a new PR for what you decide to do with overwriting -r

PR #1231, opened 2022-12-14
terminate when using 'randomize' and 'change_rate' at the same time

hahattan:


hahattan:
@janiversen would you mind releasing v3.0.3 with the new `--change-rate` function ?

janiversen:
the current plan is to make 3.1 before the end of the year. We already have a number of commits done for that so I would really prefer not to revert back and make 3.0.3.

I suppose you know that you can patch your site-packages directory in the meantime.

PR #1244, opened 2023-01-09
Fix reconnectDelay to be within (100ms, 5min)

### LINKED ISSUE #1243: TCP reconnect_delay changed from (100ms, 5min) to (5 min, infinite)
### Versions

* Python: 3.11
* OS: macOS
* Pymodbus: newer than 3ce0dbd8f1b77cf4278fb024500ba64bd2b4ee08

### Pymodbus Specific
* async TCP client (probably affects all)

### Description

As of https://github.com/riptideio/pymodbus/pull/266, pymodbus had a minimum reconnection delay of 100ms, which doubled on every reconnection failure up to a maximum of 5min.  (This is a simple binary exponential backoff)
```python
class ReconnectingAsyncioModbusTcpClient(object)
    #: Minimum delay in milli seconds before reconnect is attempted.
    DELAY_MIN_MS = 100
    #: Maximum delay in milli seconds before reconnect is attempted.
    DELAY_MAX_MS = 1000 * 60 * 5

    def __init(self):
           #: Reconnect delay in milli seconds.
            self.delay_ms = self.DELAY_MIN_MS
    def _reconnect(self):
        _logger.debug('Waiting %d ms before next '
                      'connection attempt.' % self.delay_ms)
        yield from asyncio.sleep(self.delay_ms / 1000)
        self.delay_ms = min(2 * self.delay_ms, self.DELAY_MAX_MS)
        yield from self._connect()
```
This worked fine, as far as I can tell.

In 3ce0dbd8f1b77cf4278fb024500ba64bd2b4ee08 this logic was changed.  The minimum reconnection delay is now 5 min (30000 ms) and there is no maximum.

```python
class Defaults(Singleton):
        ReconnectDelay = 1000 * 60 * 5

class AsyncModbusTcpClient(ModbusBaseClient):
    def __init__(self, 
                 reconnect_delay: int = Defaults.ReconnectDelay):
        self.delay_ms = self.params.reconnect_delay
    async def _reconnect(self):
        txt = f"Waiting {self.delay_ms} ms before next connection attempt."
        _logger.debug(txt)
        await asyncio.sleep(self.delay_ms / 1000)
        self.delay_ms = 2 * self.delay_ms
```

I do not see the purpose of such a change, as it makes reconnections take very very long.  Was it perhaps unintentional?  I believe the logic change should be reverted (while preserving the refactor).

alexrudd2:
Closes #1243

I'm not sure if this breaks the serial code, but can't get pytest setup locally so I need to run the CI.

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -31,7 +31,8 @@ class ModbusBaseClient(ModbusClientMixin):
     :param close_comm_on_error: (optional) Close connection on error.
     :param strict: (optional) Strict timing, 1.5 character between requests.
     :param broadcast_enable: (optional) True to treat id 0 as broadcast address.
-    :param reconnect_delay: (optional) Delay in milliseconds before reconnecting.
+    :param reconnect_delay_min: (optional) Minimum delay in milliseconds before reconnecting.
+    :param reconnect_delay_max: (optional) Maximum delay in milliseconds before reconnecting.
     :param kwargs: (optional) Experimental parameters.
 
     .. tip::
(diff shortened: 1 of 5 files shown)


janiversen (approved):
Let's keep reconnect_delay, for 2. reasons:
- avoid a breaking API change
- when not using max, it is more logical to set reconnect_delay.

janiversen:
Thanks for the PR.

PR #1247, opened 2023-01-09
Parallelize pytest with pytest-xdist

alexrudd2:
Consider adding this requirement for `pytest-xdist` to speed up pytest runs by running them [in 
parallel](https://pytest-xdist.readthedocs.io/en/latest/distribution.html).

Pros:
+50% speedup on my laptop

Cons:
- Extra dependency
- Pytest output does not list individual tests any more

```
collected 417 items

test/test_all_messages.py ...                                                                              [  0%]
test/test_bit_read_messages.py .........                                                                   [  2%]
test/test_bit_write_messages.py .........                                                                  [  5%]
test/test_client.py ............................................s.........                                 [ 17%]
test/test_client_sync.py .....
```
```
gw0 [417] / gw1 [417] / gw2 [417] / gw3 [417] / gw4 [417] / gw5 [417] / gw6 [417] / gw7 [417]
.......................................s.................................................................. [ 25%]
.......................................................................................................... [ 50%]
.......................................................................................................... [ 76%]
...................................................................................................        [100%]
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
             cmd: make -C doc/ html
             type: lint
           - name: pytest
-            cmd: pytest --cov=pymodbus --cov=test --cov-report=term-missing --cov-report=xml -v --full-trace --timeout=20
+            cmd: pytest --cov=pymodbus --cov=test --cov-report=term-missing --cov-report=xml -v --full-trace --timeout=20 --numprocesses auto
             type: test
         os:
           - name: Linux
--- a/check_ci.sh
+++ b/check_ci.sh
@@ -10,5 +10,5 @@ black --safe --quiet examples/ pymodbus/ test/
 isort .
 pylint --recursive=y examples pymodbus test
 flake8
-pytest
+pytest --numprocesses auto
 echo "Ready to push"
--- a/requirements.txt
+++ b/requirements.txt
@@ -70,3 +70,4 @@ pytest==7.1.3
 pytest-asyncio==0.19.0
 pytest-cov==4.0.0
 pytest-timeout==2.1.0
+pytest-xdist==3.1.0


.github/workflows/ci.yml
```diff
@@ -47,7 +47,7 @@ jobs:
             cmd: make -C doc/ html
             type: lint
           - name: pytest
-            cmd: pytest --cov=pymodbus --cov=test --cov-report=term-missing --cov-report=xml -v --full-trace --timeout=20
+            cmd: pytest --cov=pymodbus --cov=test --cov-report=term-missing --cov-report=xml -v --full-trace --timeout=20 --numprocesses auto
```
janiversen, line 50: If this works it is pr accident. Multiple tests open a listener on port 5020, if they run in parallel they will fail.

janiversen (changes requested):
I am not sure this is a good idea, at least you either:
- need to exclude all tests that starts the server
- assign different ports to different tests that open a listener.

alexrudd2:
https://github.com/riptideio/pymodbus/blob/035ffc2991cceb109fc0243709e765a0b98a19a4/test/test_server_asyncio.py#L189-L199

Sorry for not understanding the full test suite.  However, is the port not already randomized?  If I change `random_port` to 5020 the test suite fails.  As-is, it passes.

janiversen:
random_port is a special test used only in test_server_asyncio.py, but that does not solve problem.

There are not much to understand about the test suite, a lot of the tests spin up a server which listens on a given port (typically 5020), then in the same test the client connects to port 5020 and do whatever the purpose of the test is.

Running several of these tests in parallel is impossible, because on 1 system only 1 process can listen to 1 port. Look at test_examples.py as an example.

janiversen:
Merged, due to my lack of understanding why this works.

Thanks for the PRs, looking forward to see more PRs from your side.

alexrudd2:
> Merged, due to my lack of understanding why this works.
> 
> Thanks for the PRs, looking forward to see more PRs from your side.

I do not either 😂.  My understanding of the expected problem matches yours.  I was expecting to have to use `@pytest.mark.paramaterize` but it worked somehow...


Glad to be able to contribute in a small way.  My employer maintains several FOSS packages dependant on pymodbus, and I ported them from 2.5.3 to 3.x. I'm still trying to understand a problem that port introduced regarding reconnecting, which means I need to understand pymodbus better.

janiversen:
Let me know if I can help with your understanding….open a discussion if there are things you do not understand.

I am not a big expert, I have only been working directly on pymodbus for about a year, but I think I have been in nearly all parts of the code.

PR #1249, opened 2023-01-12
make x-dist default for purest (no need for extra parameters).

janiversen:


janiversen:
@alexrudd2 a little extra, with this patch you just need to write "pytest". 

And I finally think I know why x-dist works, it is made for non-threadsafe tests, I am happy you did not choose pytest-parallel that fails.

janiversen:
Wrong repo.

PR #1250, opened 2023-01-13
Create docker-publish.yml

dhoomakethu:
Trying to use https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions to build docker images and push to the github container registry


janiversen (approved):
Perfect idea, but I cannot really tell you if the file is ok :-)

PR #1262, opened 2023-01-14
Do not build docker on pull requests.

janiversen:
If needed use Actions/CI to build manually.


janiversen:
@dhoomakethu I blocked docker generation for pull requests, because 99% do not use docker.

PR #1263, opened 2023-01-14
Solve xdist problem with sql datastore.

janiversen:

### DIFF
--- a/API_changes.rst
+++ b/API_changes.rst
@@ -5,10 +5,11 @@ PyModbus - API changes.
 -------------
 Version 3.1.0
 -------------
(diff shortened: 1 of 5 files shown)


janiversen:
@alexrudd2 There was problems with xdist !

I added a @pytest so make it run as a single process, good for you to know in your tests.

alexrudd2:
> @alexrudd2 There was problems with xdist !
> 
> I added a @pytest so make it run as a single process, good for you to know in your tests.

You now know more about `xdist` than I do :)

I'm glad that the CI now runs so much faster than last year.

PR #1271, opened 2023-01-16
Added return in AsyncModbusSerialClient.connect

dries007:
AsyncModbusSerialClient.connect is missing a return, which causes it to be unusable for `async with` blocks, since the `__aenter__` always assumes it failed now.


janiversen (approved):
Good catch, thanks.

janiversen:
I expect to release v3.1.1 at the end of the month, if I can get 2 pull requests done.

PR #1281, opened 2023-01-24
Fix Unix sockets parsing

alexrudd2:
@peufeu2 found a typo:

> tcp.py line 213:
> `            if self.params.host.startswith("unit:"):`
> 
> "unit" should be "unix", since it's about unix sockets...
> 

_Originally posted by @peufeu2 in https://github.com/pymodbus-dev/pymodbus/issues/1279#issuecomment-1400424302_

### DIFF
--- a/pymodbus/client/tcp.py
+++ b/pymodbus/client/tcp.py
@@ -210,7 +210,7 @@ class ModbusTcpClient(ModbusBaseClient):
         if self.socket:
             return True
         try:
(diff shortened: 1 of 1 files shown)


janiversen:
Now how did that make it onto dev, my fault, I thought I had corrected it.

janiversen (approved):
good catch, and thanks for the PR.

PR #1296, opened 2023-01-25
Document simulator.

janiversen:


alexrudd2:
Screenshot 2023-01-27 at 10 33 05 AM

This has created two duplicate files with the same name but different cases.

janiversen:
That is clearly an error, which I did not see. I do not have access to a command line at the moment so if you have time please make a PR. 

Let’s try only to use lower case file names (I had a problem with Visual Studio Code, that used class names as file names).

janiversen:
Trying to see if I can do it from github.

PR #1298, opened 2023-01-26
fix no auto reconnect after close/connect in TCPclient

### LINKED ISSUE #1198: calling client.close() does not terminate the client process but causes a reconnect.
### Versions

* Python:
* OS:
* Pymodbus:
* Modbus Hardware (if used):

### Pymodbus Specific
* Server: tcp/rtu/ascii - sync/async
* Client: tcp/rtu/ascii - sync/async

### Description

### Code and Logs

```python
# code and logs here.

# please use the following to format logs when posting them here
import pymodbus
pymodbus.pymodbus_apply_logging_config()
```

peufeu2:
Greetings,

AsyncModbusTcpClient.close() followed by connect() to reuse the object left it unable to auto-reconnect because delay_ms was set to 0 in close() but not reset to default setting in connect().

Added canceling of auto-reconnect background task in close().

Added check to avoid launching more than one auto-reconnect task in case user calls connect() too many times.

fixes #1198


pymodbus/client/tcp.py
```diff
@@ -97,7 +108,7 @@ def _create_protocol(self):
         protocol.factory = self
         return protocol
 
-    async def _connect(self):
+    async def _connect( self ):
```
janiversen, line 111:
```suggestion
    async def _connect(self):
```
Please keep our standard

janiversen (changes requested):

janiversen:
Please also solve the CI errors.

janiversen:
delay_ms is not part of the API, but in many of the tests are allowed to use internal values.

Please correct the tests, maybe the tests should simply use the timeout parameter.

peufeu2:
Done, looks good!

janiversen (approved):
LGTM. Thanks

janiversen:
Seems there are a problem with python 3.11, I have restarted the job to see if it is a real error.

PR #1309, opened 2023-01-27
Fix register type check for size bigger than 3

duc996:
This patch fixes the issue where larger string requests (len > 3) trigger an error.


janiversen (changes requested):
This looks like a wrong check to me. If you look at our test cases, we are testing a couple of strings that are larger than 3.

As goes for reading the strings, it is totally legal (unless type_checking is active) to read the middle or only part of a string.

The suggested fix will also influence the second register of UINT32 and float32, which I assume is not your idea.

janiversen:
You need to supply a test case (see test directory) so we can see the error, apply your patch and see it is solved.

janiversen:
As I suspected this PR breaks our tests….of course it could be that the test cases are wrong, but that calls for an explanation.

duc996:
> This looks like a wrong check to me. If you look at our test cases, we are testing a couple of strings that are larger than 3.
I can't find any test case with strings larger than 3. Perhaps I'm not looking on the right place.
> 
> As goes for reading the strings, it is totally legal (unless type_checking is active) to read the middle or only part of a string.
I believed the validate_type is only called when type_checking is active.
> 
> The suggested fix will also influence the second register of UINT32 and float32, which I assume is not your idea.
Yes, I thought an 32bit values must have on the fist 2 bytes the type and the following 2 the NEXT type.

Anyway, you are completely right, I have missed a test case and all the unit tests. A bit field test seems to be broken, I'll try to find out way (I haven't expected that bit field tests can be for more than 1 byte).
I'll try to understand the sources a little bit more and give you some feedback as soon as possible.

FYI: I have tried the modbus simulator in the server/simulator directory and while testing I had issues to read strings longer than 3 bytes. That's why I have started to dig into this issue.

janiversen:
OK, I look forward to hear from you.

I hope you are aware that string size 3, 5, 7…..do not exist !! strings consist of bytes and each register contains 2 bytes, thus a string size is always a multiple of 2.

janiversen:
If you look at the latest dev in readthedocs, there are some (hopeful very helpful) documentation of the simulator datastore.

duc996:
I have just created a new branch on my forked project and committed the correction and updated unit test. It seems that I can't update the pull request without created a branch. Take a look an let me know if I should create a new pull request.

>All bit request (coils, discrete) return a minimum of 16 bits….if you do read_coils(addr, count=7) you will get 1 register with >2 bytes returned where the 7 msb contains valid bit.
>if you do read_coils(addr, count=21) you will get 2 register with 4 bytes returned where the 21 msb contains valid bits.
>Maybe you should read the modbus standard, pymodbus follows that standard 100%.
I'm aware of that, but when I started testing I started with request on address 1, which don't work on the simulator. I found out that the first valid address for coils seems to be 16. Additionally it's really hard to find out what value must be configured to get the expected response on coil requests.
Additionally the NEXT cell type is intriguing me. I thought large bit fields would have also a next type.
And why should it not be possible to read a string partially? The current type check will refuse these requests.

And for the string size it concerns the count parameter. In my comments I have at some point changed the count to bytes which is not correct. So it should be strings longer than 6 bytes.

janiversen:
String is NOT == count but string size == count * 2, because each register contains 2 bytes.

janiversen:
I cannot look at your forked project. You can always use “git rebase” to update the branch which are base for this PR, or if you prefer, close this PR and make a new one.

duc996:
I haven't created a branch before the PR and I haven't found a way to update the PR without branch. I'll create a new PR.

PR #1311, opened 2023-01-27
Delete ModbusCommonBlock.png

janiversen:
Let see if this works


janiversen:
@alexrudd2 thanks for catching this !

alexrudd2:
> @alexrudd2 thanks for catching this !

Actually it was `git` that warned me :)

PR #1314, opened 2023-01-27
More direct readme links for REPL

alexrudd2:
Skip an intermediate page and go directly to the relevant README

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -140,7 +140,7 @@ Pymodbus REPL comes with many handy features such as payload decoder
 to directly retrieve the values in desired format and supports all
 the diagnostic function codes directly .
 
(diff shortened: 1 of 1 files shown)


janiversen (approved):
OK for me, even though I do not understand the difference. When I open the readme in my browser it uses “tree” not “blob”.

alexrudd2:
hmm, `tree` versus `blob` was unexpected, but doesn't matter.

The link for "Pymodbus REPL client" should go directly to `[...]dev/pymodbus/repl/client/README.md` instead of `[...]dev/pymodbus/repl/`

janiversen:
Thanks.

janiversen:
You are now among the 5 most active developers https://github.com/pymodbus-dev/pymodbus/graphs/contributors thanks for your contributions, it helps a lot !

PR #1321, opened 2023-01-30
Set socket keepalive on client and server.

janiversen:
This detects a broken connection, even though there are no modbus calls.


janiversen:
This will never work, because UDP do not have connections, and windows is always connected (to the serial port).

PR #1323, opened 2023-01-31
Fix register type check for size bigger than 3 registers (6 bytes)

duc996:
This commit fixes the register type check for configuration where more than 3 registers are configured as one area (I think for the moment this happens only for string types for the moment).
I updated unit tests for this case.

### DIFF
--- a/pymodbus/datastore/simulator.py
+++ b/pymodbus/datastore/simulator.py
@@ -517,22 +517,12 @@ class ModbusSimulatorContext:
     _write_func_code = (5, 6, 15, 16, 22, 23)
     _bits_func_code = (1, 2, 5, 15)
 
-    def validate(self, func_code, address, count=1):
-        """Check to see if the request is in range.
+    def loop_validate(self, address, end_address, fx_write):
+        """Validate entry in loop.
 
         :meta private:
         """
-        if func_code in self._bits_func_code:
-            # Bit count, correct to register count
-            count = int((count + WORD_SIZE - 1) / WORD_SIZE)
-            address = int(address / 16)
-        real_address = self.fc_offset[func_code] + address
-        if real_address <= 0 or real_address > self.register_count:
-            return False
-
-        fx_write = func_code in self._write_func_code
-        i = real_address
-        end_address = real_address + count
+        i = address
         while i < end_address:
             reg = self.registers[i]
             if fx_write and not reg.access or reg.type == CellType.INVALID:
@@ -548,12 +538,28 @@ class ModbusSimulatorContext:
                 if i + 1 >= end_address:
                     return False
                 i += 2
-            else:  #  CellType.STRING
+            else:
                 i += 1
-                while i  < end_address:
+                while i < end_address:
                     if self.registers[i].type == CellType.NEXT:
                         i += 1
         return True
+
+    def validate(self, func_code, address, count=1):
+        """Check to see if the request is in range.
+
+        :meta private:
+        """
+        if func_code in self._bits_func_code:
+            # Bit count, correct to register count
+            count = int((count + WORD_SIZE - 1) / WORD_SIZE)
+            address = int(address / 16)
+        real_address = self.fc_offset[func_code] + address
+        if real_address <= 0 or real_address > self.register_count:
+            return False
+
+        fx_write = func_code in self._write_func_code
+        return self.loop_validate(real_address, real_address + count, fx_write)
 
     def getValues(self, func_code, address, count=1):  # pylint: disable=invalid-name
         """Return the requested values of the datastore.
@@ -731,9 +737,7 @@ class ModbusSimulatorContext:
             check = (CellType.UINT32, CellType.FLOAT32, CellType.STRING)
             reg_step = 2
 
-        for i in range(  # pylint: disable=consider-using-any-or-all
-            real_address, real_address + count, reg_step
-        ):
+        for i in range(real_address, real_address + count, reg_step):
             if self.registers[i].type in check:
                 continue
             if self.registers[i].type is CellType.NEXT:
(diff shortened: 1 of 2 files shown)


janiversen:
Did you try the test ? 

please remember what I wrote make a test case thst shows the problem, add your fix  and check that the test works.

your test does not work with your fix.

duc996:
> Did you try the test ?
> 
> please remember what I wrote make a test case thst shows the problem, add your fix and check that the test works.
> 
> your test does not work with your fix.

Yes, I run the test, but only the simulator test but other tests are broken.
I missed the point that the commit order is important. I have updated the pull request with 1. the updated test case and 2. the correction.

A little side note: who should I start my tests? I used the commend from the CI, but there is also the pylint. Is there any command to start all tests from CI? Also serial-ModbusRtuFramer and serial-ModbusAsciiFramer are failing most of the time on my local PC event without my patch so I think this is not related to it.

janiversen:
Read the documentation that is a good start.

duc996:
> Read the documentation that is a good start.
Thanks, I found the check_ci.sh script.

janiversen:
Let me take your questions one by one.
a) commit order is not important ! CI checks the Pull Request as a merged commit.
b) of course other tests break, it is one library, so changes in one corner can easily affect another part.
c) I have not idea, what you mean by  serial-ModbusRtuFramer and serial-ModbusAsciiFramer fails, the test cases must work also on your PC. If it is your usage, well...

I will take a look first at your test case, then at the production code.

pymodbus/datastore/simulator.py
```diff
@@ -724,8 +724,11 @@ def validate_type(self, func_code, real_address, count):
         for i in range(  # pylint: disable=consider-using-any-or-all
             real_address, real_address + count, reg_step
         ):
-            if self.registers[i].type not in check:
-                return False
+            if self.registers[i].type in check:
+                continue
+            if self.registers[i].type is CellType.NEXT:
```
janiversen, line 729: This is not right: real_address + 0 may not be NEXT, because it means you read off base.

For BITS, UINT16 next is never allowed (they occupy exact 1 register).
For UINT32, FLOAT32, first address must never be NEXT, second address must be NEXT, 3 address never etc....
For STRING first address must be STRING, next addresses must be NEXT / STRING.

So it is a bit more complex.

janiversen (changes requested):
Nice catch in the tests !!! Seems I forgot to test that one.

Actually I think you are the first to try strong typing.

janiversen:
When changing validate_type, we should also check that count for UINT32, FLOAT32 is a multiple of 2.

duc996:
Thanks for your reposes.
for a): I thought it's for you to see the testcase first and can judge if it's a valid usecase.
b) I was to optimistic :)
and c) if I run the check_ci script twice, I get e.g. the following error:
 
`OSError: [Errno 98] error while attempting to bind on address ('127.0.0.1', 5030): address already in use
--------------------------------------------------------------- Captured log call ---------------------------------------------------------------
INFO     root:client_async.py:44 ### Create client object
INFO     root:client_async.py:120 ### Client starting
DEBUG    pymodbus.client.serial:serial.py:107 Starting serial connection
WARNING  pymodbus.client.serial:serial.py:125 Failed to connect: Could not open port socket://127.0.0.1:5030: [Errno 111] Connection refused
------------------------------------------------------------- Captured log teardown -------------------------------------------------------------
DEBUG    asyncio:selector_events.py:54 Using selector: EpollSelector
============================================================ short test summary info ============================================================
FAILED test/test_examples.py::test_exp_async_server_client[serial-ModbusRtuFramer-5020-10] - AssertionError
=================================================== 1 failed, 459 passed, 1 skipped in 6.98s ====================================================
"pytest --numprocesses auto" command filed with exit code 1.`

pymodbus/datastore/simulator.py
duc996: As I said before, I don't really understand the NEXT celltype.
Why should it not be authorized to read the UINT32 in two requests or to read two UINT32 values in one request, but it's authorized to read only one register of the UINT32.
For strings it's similar, if there is a longer string and I would like to get only a part of it, why this wouldn't be allowed?
Initially I thought the type check is more of a function code check.

janiversen:
that is because you have something running that blocks port 5030.

pymodbus/datastore/simulator.py
janiversen: NEXT is quite easy, it is the second register in UINT32, FLOAT32, and remaining registers in STRING.

Consider this sequence of registers:
12: UINT32,  13: NEXT,  14: UINT32, 15: NEXT, 16:UINT32, 17:NEXT

The following are valid reads:
Read_holding(12, 2)
Read_holding(12, 6)
Read_holding(14,2)

The following are invalid (Because they do not read whole UINT32)
Read_holding(12, 1)
Read_holding(13, 1)
Read_holding(12, 3)

FLOAT32 are exactly the same.

STRINGS are not so different:
12: STRING, 13: NEXT, 14: NEXT, 15: NEXT, 16: STRING, 17: NEXT

The following are valid:
read_holding(12, 4)
read_holding(12, 6) # read 2 complete strings (allowed just as it is allowed to read multiple UINT32)
read_holding(16, 2)

Type check is there to ensure you read the strings correctly, so the following is not allowed:
read_holding(12, 1)
read_holding(12, 1)
read_holding(13, 1)

Does that help ?

janiversen: The type check is there to ensure you read the registers as the designer of the device intended. It is a "verification", and you can find X reasons why it should be different, but then the answer is configure it to False!

duc996: yes, much better :)

duc996:
> that is because you have something running that blocks port 5030.
No, I don't have anything running on port 5030. The first start is running ok, but the following tests don't work. I had to run my tests several times :)
It's on my linux pc and with netstat I can see that the sockets are in TIME_WAIT. I have to wait for this state to finish and than the test is reworking. It seems on my system this timeout is 60 seconds.

janiversen:
Look at CI, we use Linux and it works.

janiversen:
Thanks.

PR #1344, opened 2023-02-16
Remove unused linters

alexrudd2:
It appears to me that neither `pep8` nor `lib2to3` are used anymore, and can be removed.


janiversen:
Good catch.

PR #1351, opened 2023-02-17
Fix more typing issues

alexrudd2:
I'm trying out `mypy`.  This is still a WIP, although you can merge early if you want.  

What should we do with implicit optional types? https://adamj.eu/tech/2022/10/18/python-type-hints-implicit-optional-types/


janiversen:
I look forward to having mypy activity….at the very least for the api. Thanks a lot for taking this on.

I prefer to allow “implicit option types” that is writing “text: str = None” is allowed. Forcing “text: Optional[str, None] = None” is just a real PITA.

I am happy to review your stuff whenever you feel you have reached a level, where you want it merged. Just put it in review, and I will review/approve fast.

jamesbraza:
@janiversen fwiw, even though `Optional` is a bit ugly and expands signatures, it's recommended in PEP 484.  The authors actually changed their minds on the implicit typing.

https://peps.python.org/pep-0484/#union-types

> A past version of this PEP allowed type checkers to assume an optional type when the default value is `None`

> This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

janiversen:
I know that it is recommended, but it not demanded. We support 3.8+ so for us it would be a lot of useless typing, which do not make the code better or more readable.

pymodbus/repl/server/main.py
```diff
@@ -138,7 +138,7 @@ def run(
         autocompletion=framers,
         help="Modbus framer to use",
     ),
-    modbus_port: str = typer.Option("5020", "--modbus-port", "-p", help="Modbus port"),
+    modbus_port: int = typer.Option(5020, "--modbus-port", "-p", help="Modbus port"),
```
janiversen, line 141: This needs a conversion to int, otherwise we receive a str.

pymodbus/repl/server/main.py
```diff
@@ -180,7 +180,6 @@ def run(
     modbus_config = modbus_config.get(modbus_server, {})
     modbus_config = process_extra_args(extra_args, modbus_config)
     if modbus_server != "serial":
-        modbus_port = int(modbus_port)
```
janiversen, line 183: Removing this causes problems, because the parameter is read as a str and the lower layers expect an int.

janiversen (changes requested):
I think we should merge this soon to avoid having clashes with other pull requests. Making a merge after every 10-20 files sounds good to me.

It is a big job you have taken on, but it looks very very nice.

pymodbus/repl/server/main.py
alexrudd2: `typer` does the conversion already. https://typer.tiangolo.com/tutorial/parameter-types/#data-conversion

Also, since `typer` was new to me I tested it (by adding `print(modbus_port, type(modbus_port)`)

```
❯  pymodbus.server run --modbus-server tcp --modbus-port 1234

1234 <class 'int'>
```

alexrudd2:
> I think we should merge this soon to avoid having clashes with other pull requests. Making a merge after every 10-20 files sounds good to me.
> 
> It is a big job you have taken on, but it looks very very nice.

OK, it should be ready to merge.  `mypy` errors go from 111 -> 94 with this PR.

janiversen (approved):
A very good start. I did not know that typer() converted, thanks for testing it.

janiversen:
I merged the first part….looks very good to me. 

Looking forward to see next part.

PR #1356, opened 2023-02-18
Even more type fixes

alexrudd2:
Down to 32 `mypy` errors.

I don't quite understand how the mixin works, but using `@abstractmethod` seemed appropriate.  Perhaps it should also be used in 
`base.py`?
https://github.com/pymodbus-dev/pymodbus/blob/39f1c8c9ba07cd5c974e5367c895fcbd39ecf858/pymodbus/client/base.py#L153-L160

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -1,5 +1,4 @@
 """Modbus Client Common."""
-from abc import abstractmethod
 from typing import Any, List, Tuple, Union
 
 import pymodbus.bit_read_message as pdu_bit_read
@@ -10,7 +9,7 @@ import pymodbus.mei_message as pdu_mei
 import pymodbus.other_message as pdu_other_msg
 import pymodbus.register_read_message as pdu_reg_read
 import pymodbus.register_write_message as pdu_req_write
-from pymodbus.pdu import ModbusRequest
+from pymodbus.pdu import ModbusRequest, ModbusResponse
 
 
 class ModbusClientMixin:  # pylint: disable=too-many-public-methods
@@ -40,8 +39,7 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
     def __init__(self):
         """Initialize."""
 
-    @abstractmethod
-    def execute(self, request: ModbusRequest):
+    def execute(self, request: ModbusRequest) -> ModbusResponse:
         """Execute request (code ???).
 
         :param request: Request to send
@@ -53,7 +51,7 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
         .. tip::
             Response is not interpreted.
         """
-        return request  # (for pytest)  Concrete methods return a response
+        return request
 
     def read_coils(
         self, address: int, count: int = 1, slave: int = 0, **kwargs: Any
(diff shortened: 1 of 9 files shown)


janiversen (changes requested):
These methods are not abstract, that is the implemtataion. The methods are called from apps that use the library.

janiversen:
Base client have a couple of abstract methods, very easy to see by the assert (only statement in the method).

Methods in mixin are final and not inherited.

alexrudd2:
I only changed `ModbusClientMixin.execute`.   It only echoes `request` and does not return a `ModbusResponse`.  All the subclasses which override `execute` return a subclass of `ModbusResponse`.  

Therefore, it looked like an abstract method, but I wasn't sure because of its use in `pytest`
Perhaps I misunderstood?

pymodbus/client/mixin.py
```diff
@@ -39,7 +40,8 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
     def __init__(self):
         """Initialize."""
 
-    def execute(self, request: ModbusRequest) -> ModbusResponse:
+    @abstractmethod
```
janiversen, line 43: it returns a ModbusResponse when overwritten, so it should be defined as such. Remember this is an API spec for the library.

This is an. abstract method, but I strongly prefer not to define it as such, because it can be called from an app.

we have a lot of methods that are overwritten but that does not make them abstract. only those that must be overwritten should be declared as abstract.

The method can return a modbusResponse()

requirements.txt
```diff
@@ -68,3 +68,4 @@ pytest-asyncio==0.20.3
 pytest-cov==4.0.0
 pytest-timeout==2.1.0
 pytest-xdist==3.1.0
+sqlalchemy-stubs==0.4
```
janiversen, line 71: I cannot see you use this ?

We should not use a sub, unless it is absolutely needed.

janiversen (changes requested):
A couple of smaller changes.

requirements.txt
alexrudd2:
```
pymodbus/datastore/database/sql_datastore.py:4: error: Skipping analyzing "sqlalchemy": module is installed, but missing library stubs or py.typed marker  [import]
pymodbus/datastore/database/sql_datastore.py:5: error: Skipping analyzing "sqlalchemy.types": module is installed, but missing library stubs or py.typed marker  [import]
pymodbus/datastore/database/sql_datastore.py:6: error: Skipping analyzing "sqlalchemy.pool": module is installed, but missing library stubs or py.typed marker  [import]
pymodbus/datastore/database/sql_datastore.py:7: error: Skipping analyzing "sqlalchemy.schema": module is installed, but missing library stubs or py.typed marker  [import]
pymodbus/datastore/database/sql_datastore.py:8: error: Skipping analyzing "sqlalchemy.sql": module is installed, but missing library stubs or py.typed marker  [import]
pymodbus/datastore/database/sql_datastore.py:9: error: Skipping analyzing "sqlalchemy.sql.expression": module is installed, but missing library stubs or py.typed marker  [import]
```

There are missing type stubs for `pyserial` also, available via `pip install types-pyserial` but I'm not sure if they're correct right now: https://github.com/python/typeshed/pull/9722

janiversen (approved):
Execute should really return a MobusResponse, but it seems you have put that in another PR.

Issue #1357, opened 2023-02-18
mypy reports invalid return value for execute() on ModbusClientMixin and its subclasses


alexrudd2:
```python
class ModbusClientMixin:  # pylint: disable=too-many-public-methods
    def execute(self, request: ModbusRequest) -> ModbusResponse:
        return request 

    def read_coils(
        self, address: int, count: int = 1, slave: int = 0, **kwargs: Any
    ) -> pdu_bit_read.ReadCoilsResponse:
        return self.execute(
            pdu_bit_read.ReadCoilsRequest(address, count, slave, **kwargs)
        )
```
```
pymodbus/client/mixin.py:54: error: Incompatible return value type (got "ModbusRequest", expected "ModbusResponse")  [return-value]
pymodbus/client/mixin.py:68: error: Incompatible return value type (got "ModbusResponse", expected "ReadCoilsResponse")  [return-value]
x20
```

@alexrudd2 proposed removing the type hint for the return value (`-> ModbusResponse`) from `ModbusClientMixin.execute()` and using `@abstractmethod`.  

> It only echoes request and does not return a ModbusResponse. All the subclasses which override execute return a subclass of ModbusResponse. Therefore, it looked like an abstract method, but I wasn't sure because of its use in pytest
_Originally posted by @alexrudd2 in https://github.com/pymodbus-dev/pymodbus/issues/1356#issuecomment-1435769793_

@janiversen disagreed:
> it returns a ModbusResponse when overwritten, so it should be defined as such. Remember this is an API spec for the library.
> This is an abstract method, but I strongly prefer not to define it as such, because it can be called from an app.
> we have a lot of methods that are overwritten but that does not make them abstract. only those that must be overwritten should be declared as abstract.
> The method can return a modbusResponse()
_Originally posted by @janiversen in https://github.com/pymodbus-dev/pymodbus/pull/1356#discussion_r1111114890_

alexrudd2:
`mypy` reports ~20 errors, and arguably they are valid.  
`ModbusClientMixin.execute()` returns `ModbusRequest`, which does not match the type hint of `ModbusResponse` (1 error)

A subclass such as `ModbusBaseClient.execute()` actually returns different class instances depending on the method. 
 Example: `ReadCoilsResponse` or `ReadDiscreteInputsResponse`.  The type hints for various methods that call `execute()` are these _specific classes_, which does not match the part class `ModbusResponse`. (~20 errors)

What I attempted:
[...]

PR #1359, opened 2023-02-19
repl config path

alexrudd2:
`mypy` is not happy with the option `modbus_config` being used for both a Path and a dictionary (possibly) loaded from that path.  This fixes another 9 errors.

Since this is technically an interface change, I've kept it separate from my other type fixes.  Another option to to use e.g. `modbus_config_dict` but that seemed verbose.
```python
if modbus_config:
    with open(modbus_config) as my_file:  # pylint: disable=unspecified-encoding
        modbus_config_dict = json.load(my_file)
```
@dhoomakethu

### DIFF
--- a/API_changes.rst
+++ b/API_changes.rst
@@ -9,6 +9,7 @@ Version 3.2.0
   instead of using defer_start instantiate the Modbus<type>Server directly.
 - `ReturnSlaveNoReponseCountResponse` has been corrected to
   `ReturnSlaveNoResponseCountResponse`
+- Option `--modbus-config` for REPL server renamed to `--modbus-config-path`
 
 -------------
 Version 3.1.0
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -1,5 +1,4 @@
 """Modbus Client Common."""
-from abc import abstractmethod
 from typing import Any, List, Tuple, Union
 
 import pymodbus.bit_read_message as pdu_bit_read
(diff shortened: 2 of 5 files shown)


janiversen (approved):
LGTM, but please fix “ # pylint: disable=unspecified-encoding” while you are at in (copied from the description of this PR.

I am also very unhappy about the SQL-stubs, its an extra requirement that really are not needed.

dhoomakethu:
LGTM. Thanks

dhoomakethu (approved):

janiversen:
Added commit to:
- add mypy==1.0.1 (but currently do not activate it in CI or check_ci)
- Remove SQLAlchemyStubs.

If a library do not contain type hints, we should not add extra requirements just to cover that. We use our own version of async-pyserial that too comes with a level of type hints.

The sqlalchemystub slipped through my review, hence correcting it.

janiversen:
Added commit to solve black/pylint.

janiversen:
Added commit with more comments in mixin, to explain the use of execute().

Mypy needs to be silenced for the execute function (not all the others).

@alexrudd2 I will leave it up to you to correct or merge this, since I have corrected it (You have 2 approvals).

alexrudd2:
Screenshot 2023-02-19 at 12 02 06 PM

It seems there is an automatic rule to change PR targets which caused some confusion.  I intended this to be merged into #1356 Sorry

Anyways, I rebased it to squash the lint commits.

alexrudd2:
This is totally unrelated to https://github.com/pymodbus-dev/pymodbus/issues/1357 since I do not yet have a solution to it, so I removed that closing keyword.

alexrudd2:
> @alexrudd2 I will leave it up to you to correct or merge this, since I have corrected it (You have 2 approvals).

I think this PR is now ready.  However, `dev` is a protected branch, and I'm not authorized to merge into it.

janiversen:
I thought you were, I will take a look on that later. The protection should only require an approval, which you have.

Issue #1319, opened 2023-01-30
pymodbus.server requires redis in 3.1.2


beenje:
### Versions

* Python: 3.11
* OS: Linux
* Pymodbus: 3.1.2
* Modbus Hardware (if used):

### Pymodbus Specific
* Server: tcp/rtu/ascii - sync/async
* Client: tcp/rtu/ascii - sync/async

### Description

Running `pymodbus.server --help` raises an exception when `redis` isn't installed.

```
pymodbus.server --help
Traceback (most recent call last):
  File "/home/conda/feedstock_root/build_artifacts/pymodbus_1675064514324/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/bin/pymodbus.server", line 7, in <module>
    from pymodbus.repl.server.main import app
  File "/home/conda/feedstock_root/build_artifacts/pymodbus_1675064514324/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.11/site-packages/pymodbus/repl/server/main.py", line 16, in <module>
    from pymodbus.server.reactive.default_config import DEFAULT_CONFIG
  File "/home/conda/feedstock_root/build_artifacts/pymodbus_1675064514324/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.11/site-packages/pymodbus/server/__init__.py", line 5, in <module>
    from pymodbus.server.async_io import (
  File "/home/conda/feedstock_root/build_artifacts/pymodbus_1675064514324/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.11/site-packages/pymodbus/server/async_io.py", line 12, in <module>
    from pymodbus.datastore import ModbusServerContext
[...]

PR #1114, opened 2022-10-14
Make serial import dynamic

janiversen:
fixes #1112

PR #1364, opened 2023-02-19
Lint and type hints for REPL

alexrudd2:
This removes all the `mypy` warnings for REPL, and eliminates a couple other lints also.


pymodbus/repl/client/mclient.py
```diff
@@ -321,13 +321,14 @@ def report_slave_id(self, slave=Defaults.Slave, **kwargs):
             }
         return ExtendedRequestSupport._process_exception(resp)
 
-    def read_exception_status(self, **kwargs):
```
alexrudd2, line 324: I think this was previously broken

pymodbus/repl/client/mclient.py
```diff
@@ -244,7 +244,7 @@ def readwrite_registers(
             read_address=read_address,
             read_count=read_count,
             write_address=write_address,
-            write_registers=write_registers,
+            values=values,
```
alexrudd2, line 247: I think this was also broken, as the mixin doesn't have `write_registers`.  Does anyone use `readwrite_registers`?

pymodbus/repl/client/main.py
```diff
@@ -1,25 +1,9 @@
...
-try:
-    from prompt_toolkit import PromptSession, print_formatted_text
-except ImportError:
-    print(
-        "prompt toolkit is not installed!! "
-        'Install with "pip install prompt_toolkit --upgrade"'
-    )
-    sys.exit(1)
 
+import click
```
janiversen, line 5: This will not work !!! the import needs to be behind a try/except, otherwise you require click to be installed always (you will see the same in  client/serial and other places).

pymodbus/repl/client/main.py
```diff
@@ -1,25 +1,9 @@
...
+from prompt_toolkit import PromptSession, print_formatted_text
```
janiversen, line 6: This should also be in try/except.

pymodbus/repl/client/main.py
```diff
@@ -46,9 +30,7 @@
 
 _logger = logging.getLogger()
 
-click.disable_unicode_literals_warning = True
-
-TITLE = f"""
+TITLE = rf"""
```
janiversen, line 33: You remove a unicode warning ???

pymodbus/repl/client/mclient.py
```diff
@@ -220,10 +220,10 @@ def read_input_registers(self, address, count=1, slave=Defaults.Slave, **kwargs)
...
     def readwrite_registers(
         self,
-        read_address,
-        read_count,
-        write_address,
-        write_registers,
+        read_address=0,
+        read_count=0,
+        write_address=0,
+        values=0,
```
janiversen, line 226: You change optional parameters to positional parameters, are you really sure that is a good idea ?

pymodbus/repl/client/mclient.py
```diff
@@ -321,15 +321,14 @@ def report_slave_id(self, slave=Defaults.Slave, **kwargs):
...
-        In a remote device.
+    def read_exception_status(self, slave=Defaults.Slave, **kwargs):
+        """Read contents of eight Exception Status output in a remote device.
 
+        :param slave: Modbus slave unit ID
         :param kwargs:
         :return:
         """
-        request = ReadExceptionStatusRequest(**kwargs)
+        request = ReadExceptionStatusRequest(slave, **kwargs)
```
janiversen, line 331: These changes needs to be confirmed by @dhoomakethu he is the expert on this.

janiversen (changes requested):
@dhoomakethu can you please review the REPL parts carefully, I am not sure if there are side effects of these changes.

The SImulator changes are OK.

pymodbus/repl/client/mclient.py
janiversen: Could very well be, but then lets fix it.

pymodbus/repl/client/mclient.py
janiversen: Mixin should have all standard function code, independent of who uses them.

janiversen: and actually it is often used for climate controls. You read the old value  e.g. temperature and set a new one (e.g. target temperature).

I am not sure how this is broken, but if so, our tests are not correct.

alexrudd2: > Mixin should have all standard function code, independent of who uses them.
https://github.com/pymodbus-dev/pymodbus/blob/d2c7c778707d8e0cc1ea9ea30ffccf945ea56b89/pymodbus/client/mixin.py#L422-L428

The `write_registers` _method_ does not have a `write_registers` _argument_.  Instead it is `values`

> and actually it is often used for climate controls. You read the old value e.g. temperature and set a new one (e.g. target temperature).
> I am not sure how this is broken, but if so, our tests are not correct.

The broken functionality appears to be only in the REPL.  Users such as your mentioned climate controls don't use REPL.

The REPL tests are very limited, and don't actually check functionality.  If it was indeed broken, that's the first success for `mypy`. :)

pymodbus/repl/client/main.py
alexrudd2: Yes, read the commit message.  https://github.com/pymodbus-dev/pymodbus/pull/1364/commits/00a58ca1bee36b3057c91e7b4fd40a0000570269

pymodbus/repl/client/main.py
alexrudd2: The import is only needed for REPL, which can't be used anyway without `click`.  the existing code which I removed calls `sys.exit(1)`

https://github.com/pymodbus-dev/pymodbus/blob/d2c7c778707d8e0cc1ea9ea30ffccf945ea56b89/pymodbus/repl/client/main.py#L9-L13

pymodbus/repl/client/mclient.py
alexrudd2: The method in the base class already uses default values, meaning this is no functional change.
https://github.com/pymodbus-dev/pymodbus/blob/d2c7c778707d8e0cc1ea9ea30ffccf945ea56b89/pymodbus/client/mixin.py#L500-L508

pymodbus/repl/client/mclient.py
alexrudd2: This commit does exactly that.

pymodbus/repl/client/main.py
janiversen: Yes but we have had problems, latest with sqlalchemy because python loads the files and thus tries to do the import. The problem arises is the package is referenced (__jnit__) so for optional libraries we prefer to use try/except.

Having the try/except allows python to pass the module without having click installed (which is optional).

janiversen: This is the "modern" way we handle optional packages:
```
try:
    import serial
except ImportError:
    pass
```

pymodbus/repl/client/mclient.py
janiversen: is logical.

janiversen:
Once the try/except is in place (in the modern way) I will approve/merge. Of course if dhoomakethu have comments before that they should be addressed, if not they can be addressed later.

pymodbus/repl/client/mclient.py
janiversen: Very nice. The REPL server will soon be replaced by simulator, and in mid-term the REPL client will need a severe update.

pymodbus/repl/client/main.py
alexrudd2: https://github.com/pymodbus-dev/pymodbus/pull/1114 was the import error with `pyserial`, which was caused because it was imported in the generic server class.  This is called by the normal server/client code paths.
https://github.com/pymodbus-dev/pymodbus/issues/1319 was the import error with `sqlalchemy` and `redis`, which was caused because they were imported in the datastore.  This is called by the normal server/client code paths.

I don't think this is true here, since no other code in the normal server/client code paths calls the REPL.  If something did, it would already trigger the (existing) `sys.exit(1)`.  I have tested using the pymodbus client without `click` installed, and it's fine.  So the Try/Except is not protecting anything.

As far as I can tell, if `click` is not installed the REPL is useless.  So the Try/Except only changes the `ImportError` to a `NameError`.

I can add the Try/Except, of course, but don't think it has any use.

janiversen: Your analysis is correct, I was merely preparing for the future where __init__ contain all external classes, and deep links are not allowed.

Lets leave it for now.

janiversen:
As you might have noticed I work heavily in pymodbus/client (mainly base.py), so please do not make a lot of changes in that directory.

alexrudd2:
> As you might have noticed I work heavily in pymodbus/client (mainly base.py), so please do not make a lot of changes in that directory.

These are the current mypy problems there.  I will let your work settle down before touching it.
```
pymodbus/client/base.py:131: error: Too many arguments for "ModbusFramer"  [call-arg]
pymodbus/client/base.py:188: error: "ModbusBaseClient" has no attribute "protocol"; maybe "use_protocol"?  [attr-defined]
pymodbus/client/base.py:190: error: "ModbusBaseClient" has no attribute "protocol"; maybe "use_protocol"?  [attr-defined]
pymodbus/client/base.py:191: error: "connect" of "ModbusBaseClient" does not return a value  [func-returns-value]
```

PR #1366, opened 2023-02-20
Clean mixin execute

janiversen:

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -53,11 +53,7 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
         .. tip::
             Response is not interpreted.
         """
-
-        # The implementation of this method is only used in test, to secure that
-        # the methods uses the correct PDU.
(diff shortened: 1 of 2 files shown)


janiversen:
@alexrudd2 this should solve the mypy problem with mixin.py

alexrudd2:
> @alexrudd2 this should solve the mypy problem with mixin.py

Yes, it solves the problem with `execute()`.  Thanks!

I believe it should now be possible to solve the others, which are all similar.   mypy does not know that `ReadCoilsResponse` is a subclass of `ModbusResponse`.  Probably it should be e.g. `Type[ModbusResponse]` which is a type, instead of  `ModbusResponse`, an _an instance of a class_.  I will try to do something like this.
`pymodbus/client/mixin.py:69: error: Incompatible return value type (got "ModbusResponse", expected "ReadCoilsResponse")  [return-value]`

janiversen:
Sounds correct, but to be honest for me mypy is "try and fail" 👎 but it is nice once it is done.

I will not be touching mixin.py for a while.

PR #1367, opened 2023-02-20
remove second client instance in async mode.

### LINKED ISSUE #1199: Client class makes a new object in create_protocol, this is not needed and complicate debugging.
### Versions

* Python:
* OS:
* Pymodbus:
* Modbus Hardware (if used):

### Pymodbus Specific
* Server: tcp/rtu/ascii - sync/async
* Client: tcp/rtu/ascii - sync/async

### Description

### Code and Logs

```python
# code and logs here.

# please use the following to format logs when posting them here
import pymodbus
pymodbus.pymodbus_apply_logging_config()
```

janiversen:
fixes #1199

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -63,10 +63,6 @@ class ModbusBaseClient(ModbusClientMixin):
     **Application methods, common to all clients**:
     """
 
-    state = ModbusTransactionState.IDLE
-    last_frame_end: float = 0
-    silent_interval: float = 0
-
     @dataclass
     class _params:  # pylint: disable=too-many-instance-attributes
         """Parameter class."""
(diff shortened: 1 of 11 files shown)


pymodbus/client/base.py
```diff
@@ -185,9 +184,9 @@ def execute(self, request: ModbusRequest = None) -> ModbusResponse:
         :raises ConnectionException: Check exception text.
         """
         if self.use_protocol:
-            if not self.protocol:
+            if not self._connected:
                 raise ConnectionException(f"Not connected[{str(self)}]")
-            return self.protocol.execute(request)
+            return self.async_execute(request)
```
alexrudd2, line 189: I do not think this works.  `async_execute` is defined as a coroutine (`async def async_execute(self, request=None):`) so calling it does nothing.

janiversen: ???? did you actually test this ?

async_execute returns a future, which allows the app to do 
```
    rr = await read_coils()
```
just like the old self.protocol.execute did.

PR #1368, opened 2023-02-20
Use type narrowing and runtime asserts to satisfy mypy

### LINKED ISSUE #1357: mypy reports invalid return value for execute() on ModbusClientMixin and its subclasses

alexrudd2:
Closes #1357, but I don't like it.

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -66,9 +66,11 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
         :param kwargs: (optional) Experimental parameters.
         :raises ModbusException:
         """
-        return self.execute(
+        response = self.execute(
             pdu_bit_read.ReadCoilsRequest(address, count, slave, **kwargs)
         )
+        assert isinstance(response, pdu_bit_read.ReadCoilsResponse)
+        return response
 
     def read_discrete_inputs(
         self, address: int, count: int = 1, slave: int = 0, **kwargs: Any
@@ -81,9 +83,11 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
         :param kwargs: (optional) Experimental parameters.
         :raises ModbusException:
         """
-        return self.execute(
+        response = self.execute(
             pdu_bit_read.ReadDiscreteInputsRequest(address, count, slave, **kwargs)
         )
+        assert isinstance(response, pdu_bit_read.ReadDiscreteInputsResponse)
+        return response
 
     def read_holding_registers(
         self, address: int, count: int = 1, slave: int = 0, **kwargs: Any
@@ -96,9 +100,11 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
         :param kwargs: (optional) Experimental parameters.
         :raises ModbusException:
         """
-        return self.execute(
+        response = self.execute(
             pdu_reg_read.ReadHoldingRegistersRequest(address, count, slave, **kwargs)
         )
+        assert isinstance(response, pdu_reg_read.ReadHoldingRegistersResponse)
+        return response
 
     def read_input_registers(
         self, address: int, count: int = 1, slave: int = 0, **kwargs: Any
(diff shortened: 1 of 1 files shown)


pymodbus/client/mixin.py
```diff
@@ -66,9 +66,11 @@ def read_coils(
         :param kwargs: (optional) Experimental parameters.
         :raises ModbusException:
         """
-        return self.execute(
+        response = self.execute(
             pdu_bit_read.ReadCoilsRequest(address, count, slave, **kwargs)
         )
+        assert isinstance(response, pdu_bit_read.ReadCoilsResponse)
```
janiversen, line 72: This will not work in practical life !! we only get a ReadCoilsResponse if the device responded positively...otherwise we get typically ModBusIOException (or something similar) this is NOT an exception but a message from the device.

Apart from that it is hard to see how that helps on mypy. I can accept it if you remove the assert, but it really needs a comments as to why this is nessecary.

janiversen (changes requested):

pymodbus/client/mixin.py
alexrudd2: > This will not work in practical life !! we only get a ReadCoilsResponse if the device responded positively...otherwise we get typically ModBusIOException (or something similar) this is NOT an exception but a message from the device.

Good point, and it means all the existing type signatures are too narrow.
They should become, for example, `Union[pdu_bit_read.ReadCoilsResponse, exceptions.ModbusException]`



> Apart from that it is hard to see how that helps on mypy. I can accept it if you remove the assert, but it really needs a comments as to why this is nessecary.

The problem is the return type for `execute()` is too broad, and the other methods want _more specific_ types.

```python

def execute(self, request: ModbusRequest) -> ModbusResponse:  # cannot be more specific with static analysis, only at runtime
    raise ModbusException("Pymodbus internal ERROR")

def read_coils(
    self, address: int, count: int = 1, slave: int = 0, **kwargs: Any
) -> pdu_bit_read.ReadCoilsResponse:  # more specific than ModbusResponse
```

I think the best solution is to remove the response type hint for `execute()`.  I did this in https://github.com/pymodbus-dev/pymodbus/pull/1356/commits/636e9a779774302bfcbf0d82a4542c11965f56a4 but you disagreed  :).  I am trying alternative solutions, but don't like any of them (hence 'Draft' status)

janiversen: exceptions.ModbusException is a real exception and not a modbus Message. Have a look at client_calls.py in examples to see what can be returned (or raised).

Something is not right, in home assistant we often use broad classes (like e.g. sensor) that are then narrowed down when being used (like e.g. a binary_sensor). However if the problems goes away by removing the return type from execute() then by all means. 

The alternative is to add a cast to the execute() in each method.

PR #1369, opened 2023-02-20
Fix type hints for http_server

alexrudd2:
Eliminates 3 more errors.

```
pymodbus/server/simulator/http_server.py:141: error: Argument 2 to "ModbusSimulatorContext" has incompatible type "str"; expected "Dict[str, Callable[..., Any]]"  [arg-type]
pymodbus/server/simulator/http_server.py:178: error: No overload variant of "__setitem__" of "list" matches argument types "int", "str"  [call-overload]
pymodbus/server/simulator/http_server.py:178: note: Possible overload variants:
pymodbus/server/simulator/http_server.py:178: note:     def __setitem__(self, SupportsIndex, Callable[[Any, Any], Any], /) -> None
pymodbus/server/simulator/http_server.py:178: note:     def __setitem__(self, slice, Iterable[Callable[[Any, Any], Any]], /) -> None
pymodbus/server/simulator/http_server.py:181: error: Need type annotation for "call_list" (hint: "call_list: List[<type>] = ...")  [var-annotated]
```

### DIFF
--- a/pymodbus/server/simulator/http_server.py
+++ b/pymodbus/server/simulator/http_server.py
@@ -130,7 +130,7 @@ class ModbusSimulatorServer:
         }
         if custom_actions_module:
             actions_module = importlib.import_module(custom_actions_module)
-            custom_actions_module = actions_module.custom_actions_dict
+            custom_actions_dict = actions_module.custom_actions_dict
         server = setup["server_list"][modbus_server]
         server["loop"] = asyncio.get_running_loop()
         if server["comm"] != "serial":
@@ -138,7 +138,9 @@ class ModbusSimulatorServer:
             del server["host"]
             del server["port"]
         device = setup["device_list"][modbus_device]
-        self.datastore_context = ModbusSimulatorContext(device, custom_actions_module)
+        self.datastore_context = ModbusSimulatorContext(
+            device, custom_actions_dict or None
+        )
         datastore = ModbusServerContext(slaves=self.datastore_context, single=True)
         comm = comm_class[server.pop("comm")]
         framer = framer_class[server.pop("framer")]
--- a/pymodbus/server/simulator/http_server.py
+++ b/pymodbus/server/simulator/http_server.py
@@ -163,10 +163,10 @@ class ModbusSimulatorServer:
         self.web_app.on_startup.append(self.start_modbus_server)
         self.web_app.on_shutdown.append(self.stop_modbus_server)
         self.generator_html = {
-            "log": [None, self.build_html_log],
-            "registers": [None, self.build_html_registers],
-            "calls": [None, self.build_html_calls],
-            "server": [None, self.build_html_server],
+            "log": ["", self.build_html_log],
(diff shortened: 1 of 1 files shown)


pymodbus/server/simulator/http_server.py
```diff
@@ -374,12 +373,12 @@ def helper_build_html_calls_submit_set(self, params):
...
         """Build html calls submit."""
         call_foot = ""
-        if params["submit"] == "Clear":
-            self.call_list = []
-            call_foot = "Cleared list, monitoring active"
-        elif params["submit"] == "Stop":
+        if params["submit"] == "Stop":
             self.call_monitor = CallTypeMonitor()
             call_foot = "Stopped monitoring"
+        # elif params["submit"] == "Clear":
```
alexrudd2, line 379: I preserved this logic since there seems to be an intended use of `call_list` but I don't know what it is yet.  It should either be finished or removed entirely.

pymodbus/server/simulator/http_server.py
```diff
@@ -130,15 +130,15 @@ def __init__(
         }
         if custom_actions_module:
             actions_module = importlib.import_module(custom_actions_module)
-            custom_actions_module = actions_module.custom_actions_dict
+            custom_actions_dict = actions_module.custom_actions_dict
```
janiversen, line 133: Changing the name is ok, but then you need to make sure the custom_actions_dict is set to None when custom_actions_module is None otherwise line 141 is wrong.

pymodbus/server/simulator/http_server.py
janiversen: Why do you remove the submit function "Clear" that is a response to a http request, that is needed. It clears a listbox in the html.

janiversen (changes requested):
Your changes to get mypy working are VERY useful, but please try not to remove/change code without understanding it.

Your work is really appreciated and I try my best to help you understand the code, but it seems this batch of review involved a lot more than just satisfying mypy (I think about the comments I left in the datastore as well).

pymodbus/server/simulator/http_server.py
alexrudd2: `call_list` is never used anywhere in the code, besides being set to `[]`.  I think some logic might be missing.

janiversen: Yes simulator is work in progress as noted in TODO.md, but that does not mean it is correct to remove it.

alexrudd2: In that case, when you fully implement `call_list` you'll have to type it as `List[<Something>]`.  Neither mypy nor I can guess the future.

janiversen: You are right, and that can be done now as list[str], but removing the code is not a particular friendly approach.

alexrudd2: Resolved in rebase

janiversen (approved):
LGTM, thanks.

PR #1370, opened 2023-02-20
expand datastores to handle arbitrary iterables and mappings

### LINKED ISSUE #1276: Allow setting values from arbitrary iterables and mappings
### Versions

* Python: 3.10
* OS: Linux
* Pymodbus: 3.0.2
* Modbus Hardware (if used): n/w

### Description

`ModbusSparseDataBlock.setValues()` using a one-dimensional `numpy.ndarray` with registers values. This does not work. Instead of using individual values to set register values, what I got was the whole array assigned to that value. This requires a workaround of calling `.tolist()` on the array every time I'm setting values, or using a list` in the first place.

Looking at the code of `ModbusSequentialDataBlock`, we see (docstrings omitted):

```python
    def setValues(self, address, values):
        if not isinstance(values, list):
            values = [values]
        start = address - self.address
        self.values[start : start + len(values)] = values
```

I believe the check should actually be `if not isinstance(values, collections.abc.Iterable)`. This would treat any iterable the way lists are now, allowing for a wider range of use cases.

For sparse data block, the case is similar, although there is an additional test `isinstance(values, dict)`, which should probably be `isinstance(values, collections.abc.Mappin)`.

alexrudd2:
Closes #1276.

I did not test; @jaskij do you know how to test from this branch?
`python3 -m pip install --upgrade https://github.com/pymodbus-dev/pymodbus/archive/issue-1276.zip`


pymodbus/version.py
```diff
@@ -37,7 +37,7 @@ def __str__(self) -> str:
         return f"[{self.package}, version {self.short()}]"
 
 
-version = Version("pymodbus", 3, 1, "x", "")
+version = Version("pymodbus", 3, 1, 4, "alpha")
```
alexrudd2, line 40: @janiversen I think this is better than using `3, 1, "x",` which does not work with `pip`
```
     raise InvalidVersion(f"Invalid version: '{version}'")
pkg_resources.extern.packaging.version.InvalidVersion: Invalid version: '3.1.x'
```

jaskij:
A good question. You'd have to pass non-dict mappings to sparse data blocks and non-list iterables to sequential datablocks. Also, run it through mypy when using appropriate arguments.

For non-list iterables being passed to sequential datablock functions you can use numpy arrays (that's how I came across this bug).

alexrudd2:
> A good question. You'd have to pass non-dict mappings to sparse data blocks and non-list iterables to sequential datablocks. Also, run it through mypy when using appropriate arguments.
> 
> For non-list iterables being passed to sequential datablock functions you can use numpy arrays (that's how I came across this bug).

Sorry, what I meant was can you install pymodbus from this branch?  Then you can use your numpy array code and see if it's fixed.

I don't use the datastore at all, and therefore can't readily test.

janiversen:
Install pymodbus from a branch (just read the README) “pip install -e .”

pymodbus/version.py
janiversen: This seems very unrelated. I have no problem does this, but it will be after next release.

pymodbus/datastore/store.py
```diff
@@ -174,13 +177,13 @@ def getValues(self, address, count=1):
         start = address - self.address
         return self.values[start : start + count]
 
-    def setValues(self, address, values):
+    def setValues(self, address: int, values: Collection):
```
janiversen, line 180: I think list[int] is more correct.

This call and the corresponding getValues are only used internally, and do not change independent of how data is stored in the datastore.

pymodbus/datastore/store.py
```diff
@@ -174,13 +177,13 @@ def getValues(self, address, count=1):
...
         """Set the requested values of the datastore.
 
         :param address: The starting address
         :param values: The new values to be set
         """
-        if not isinstance(values, list):
+        if not isinstance(values, abc.Collection):
```
janiversen, line 186: I think collection is far too broad, you need to add test case to prove that it works.

pymodbus/datastore/store.py
```diff
@@ -263,45 +266,50 @@ def getValues(self, address, count=1):
...
             )
-        _process_as_dict(values)
-
-    def setValues(self, address, values, use_as_default=False):
+        _process_as_mapping(values)
+
+    def setValues(
+        self,
+        address: int,
+        values: Union[Mapping, Collection, float],
```
janiversen, line 296: setValues only uses an int or list[int].

Please when you make such rather deep changes, have a look at where it is used.

pymodbus/datastore/store.py
```diff
@@ -263,45 +266,50 @@ def getValues(self, address, count=1):
...
+    ):
         """Set the requested values of the datastore.
 
         :param address: The starting address
         :param values: The new values to be set
         :param use_as_default: Use the values as default
         :raises ParameterException:
         """
-        if isinstance(values, dict):
+        if isinstance(values, abc.Mapping):
```
janiversen, line 306: Same here.

janiversen (commented):
I think there are some misunderstanding in how a datastore works. You fill it with data in the init call….I think that is what is requested to be mapping etc…

Once running validate/get/set is called from other parts of pymodbus to ensure/get/set data, but those calls are always with value type: int or list[int].

pymodbus/version.py
janiversen: and it does install without problems, at least on my local machine.

janiversen (changes requested):
I wonder why GitHub decided that my request for changes was not a request. Lets try again.

pymodbus/version.py
alexrudd2: > This seems very unrelated. I have no problem does this, but it will be after next release.

The only reason for this now was so @jaskij has a simpler way to install and test.

> and it does install without problems, at least on my local machine.

with which command?  `python3 -m pip install -e .` fails for me.  So does `python3 -m pip install --upgrade https://github.com/pymodbus-dev/pymodbus/archive/issue-1276.zip` (until I changed `Version`).

alexrudd2:
> Install pymodbus from a branch (just read the README) “pip install -e .”

```
   raise InvalidVersion(f"Invalid version: '{version}'")
pkg_resources.extern.packaging.version.InvalidVersion: Invalid version: '3.1.x'
```

pymodbus/datastore/store.py
alexrudd2: Of course I look at usage.  That's why I had to expand the typing of `ModbusSparseDataBlock._process_values()` when I changed `setValues()`

alexrudd2:
> I think there are some misunderstanding in how a datastore works. You fill it with data in the init call….I think that is what is requested to be mapping etc…
> 
> Once running validate/get/set is called from other parts of pymodbus to ensure/get/set data, but those calls are always with value type: int or list[int].


In #1370 @jaskij referred to using `setValues()`, not `__init__()`.  You replied:
> Your analysis sounds correct, Pull requests are welcome, and will be reviewed positively.

_Originally posted by @janiversen in https://github.com/pymodbus-dev/pymodbus/issues/1276#issuecomment-1385986615_

Perhaps you have changed your mind after seeing the change?  Should the answer be to only change `__init__`, and we tell him not to access `setValues()` directly?

I do not use this code, which is why I asked him to test it for his use-case.

alexrudd2:
> I think collection is far too broad, you need to add test case to prove that it works.

OK, I can try to add a test case.  However, we can't use the `numpy.ndarry` and add a dependency just for testing.  So I requested functional testing from the end-user to understand if my changes work at all.

pymodbus/version.py
janiversen: I just ran it on my local machine:
```
(pymodbus) pymodbus % pip install -e .
Obtaining file:///Users/jan/repos/pymodbus
  Preparing metadata (setup.py) ... done
Installing collected packages: pymodbus
  Attempting uninstall: pymodbus
    Found existing installation: pymodbus 3.1.3
    Uninstalling pymodbus-3.1.3:
      Successfully uninstalled pymodbus-3.1.3
  Running setup.py develop for pymodbus
Successfully installed pymodbus-3.1.x

[notice] A new release of pip is available: 23.0 -> 23.0.1
[notice] To update, run: pip install --upgrade pip
(pymodbus) pymodbus % 
```
You probably did not run the pip command in your clone.

Your zip file is not a correct installation, it misses important things like wheel, and we should not recommend such installations, people who want to install in between releases need to clone the repository.

pymodbus/datastore/store.py
janiversen: yes he did but not inside the datastore, he wanted it to define the datastore, which is quite different (and a very legal request).

pymodbus/datastore/store.py
janiversen: Well then you would have seen that get/set/validate operate on registers which are lists of UINT16 and nothing else.

You have 3 angels on datastore:
- the definition or setup which is from outside and need to flexible.
- the usage which is from pymodbus internal only, and is only using list[int]
- the internal storage, which are different in the different types of datastore

Each view have its own set of rules, please do not try to make those 3 identical, that is wrong.

janiversen:
"he" does not use "getValues", you are mixing issues:
- "he" is trying to define a datastore in a more flexible way (a perfectly legal request) this affects ONLY __init__
- get/set/validate is ONLY used internally by the Pymodbus server, and should not be changed.
- datastore itself stores its data in a way that is best for that particular type of datastore (see e.g. sequential and sparce, which are very different), and we should make assumption in the base how that happens.

pymodbus/version.py
alexrudd2: You must be running `setuptools <= 65.0.0`.  I think this will break for you in `66.0.0`. 
https://github.com/pypa/setuptools/issues/3772

> Your zip file is not a correct installation, it misses important things like wheel, and we should not recommend such installations, people who want to install in between releases need to clone the repository.

I would not recommend for normal use, but it works perfectly fine here, including creating the wheel:
```
python3 -m pip install --upgrade https://github.com/pymodbus-dev/pymodbus/archive/issue-1276.zip
Collecting https://github.com/pymodbus-dev/pymodbus/archive/issue-1276.zip
  Downloading https://github.com/pymodbus-dev/pymodbus/archive/issue-1276.zip
     | 1.1 MB 5.1 MB/s 0:00:00
  Preparing metadata (setup.py) ... done
Building wheels for collected packages: pymodbus
  Building wheel for pymodbus (setup.py) ... done
  Created wheel for pymodbus: filename=pymodbus-3.1.4a0-py2.py3-none-any.whl size=161228 sha256=b29508c6c29a81f43c57d69aede4fdf181f8fb4706123920f7bc378ce76585ac
  Stored in directory: /Users/a.ruddick/Library/Caches/pip/wheels/8e/f6/cf/a802e8a32b1bb7f9fbd8b57bd82eb4cefce5f90d773751cc75
Successfully built pymodbus
Installing collected packages: pymodbus
  Attempting uninstall: pymodbus
    Found existing installation: pymodbus 3.1.x
    Uninstalling pymodbus-3.1.x:
      Successfully uninstalled pymodbus-3.1.x
Successfully installed pymodbus-3.1.4a0
```

pymodbus/datastore/store.py
alexrudd2: I re-read the original issue description twice.  It clearly discusses `setValues()`, and never mentions `__init__`.

Let's wait for the original submitter to clear this confusion.

alexrudd2:
I re-read the original issue description twice.  It clearly discusses `setValues()`, and never mentions `__init__`.

Let's wait for the original submitter to clear this confusion.

pymodbus/version.py
janiversen: I am running setuptools 65.4.1, but only for pymodbus. I have venv for each project, and setuptools got installed, when I installed pymodbus. 

I actually wonder how setuptools got installed but I expect it is part of the python installation (which again is in the venv for pymodbus).

I did not know that it creates wheel, but I still think it is dangerous to suggest such installations to users. For a couple of reasons:
- your zip is a snapshot, not frozen in time, so we cannot checkout the version to reproduce problems (install, make a commit, install, and both have the same version).
- users that want to be on the bleeding edge should clone the repository, what allows them to get the latest commits easy.

apart from that, I noticed that "pip install -e ." also works with a standard python 3.11 installation.

I have however no problem in you recommending what you think is right,, but I do have a problem changing version on dev without a release. After a release we can (as I wrote earlier) decide if we want to change the scheme.

janiversen:
No problem in waiting, but I am 100% sure he is not using get/set/validate...and if he is, it is of no interest to use, since it is outside the scope of pymodbus. Datastore is only intended to be used with the Pymodbus server/simulator.

But instead of rereading the issue, try to read e.g. in our examples how datastore is used, that will help you a lot more.

pymodbus/version.py
alexrudd2: https://github.com/pymodbus-dev/pymodbus/issues/1372

pymodbus/datastore/store.py
janiversen: Yes I know that the documentation suggest you can use setValues() in the sparseDatablock, this is actually a mistake, because it makes people assume they can change the values. setValues() is used internally by the server, and changing it might have side effects on the server.

Changing the values while the server is running will sooner or later cause raising problems, because there are no protection in setValues() to secure access between threads (or tasks). This is more likely with the sparseDataBlock than with the others because you actually replace an array.

The right way is to provide a new method, which should be valid for all datastores, that allows apps to change value/values. Such a method can provide a very generic interface allowing the flexibility needed and of course a lock() to secure the integrity of the data.

dhoomakethu:
`setValues` is supposed to be used by all datablock implementations along with `getValues` for retreiving the value. If you see the base `DataBloc` class https://github.com/pymodbus-dev/pymodbus/blob/cc4c06b4cfea72e6c17676b227be39397f592954/pymodbus/datastore/store.py#L104 , the method is supposed to be implemented in all the derived classes.  The update needs to happen with in a server context. Refer https://github.com/pymodbus-dev/pymodbus/blob/cc4c06b4cfea72e6c17676b227be39397f592954/examples/server_updating.py#L45
 
 Apologies if this discussion is not what I though is about as I am quiet behind the latest changes in the library.

janiversen:
You are spot on the issue, but are you also aware that the example you point does not provide any locking at all, so if you add an array (like wanted here), the chances grow that the following sequence will happen:
````
     block.validateAddress()  —> called async. server, returns OK.
     block.setValues(<array>) —> called from the app
     block.getValues() -> called async. from server
```
For sequential datastores https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/datastore/store.py#L186
the array is modified, and if shorter could make a previously valid address invalid.

for sparsedatablock https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/datastore/store.py#L275 the problem might arise already when assigning the array.

Normally the server uses validate/set/get to serve read/write requests, but there are no array overwrite in that code. Making the same function public, means at the very least we need to add checks to secure the server do not break down.

It might  be true, that this is how it has been for years, but that is no reason to expand on it. It is so easy to make a public facing method with the safeguard and leave the others untouched.

dhoomakethu:
@janiversen Yes thats a simple example to show the functionality, A much broader requirements needs to handle the business logic either at the `processingIncomingPacket` layer or at `setValues` with sufficient read/write locks. My suggestion would be to keep this outside the scope of the pymodbus and let anybody who wants this functionality implement it the way required. Just an honest opinion.

janiversen:
and a good opinion.

PR #1371, opened 2023-02-20
Fix typing for reactive server

alexrudd2:
Solves the following mypy errors:
- [ ] `pymodbus/server/reactive/main.py:441: error: Argument 1 to "ReactiveModbusSlaveContext" has incompatible type "**Dict[Any, Dict[Any, Any]]"; expected "BaseModbusDataBlock"  [arg-type]`
- [x] `pymodbus/server/reactive/main.py:450: error: Incompatible types in assignment (expression has type "ReactiveModbusSlaveContext", variable has type "Dict[int, ReactiveModbusSlaveContext]")  
[assignment]`

### DIFF
--- a/pymodbus/datastore/store.py
+++ b/pymodbus/datastore/store.py
@@ -51,7 +51,7 @@ from pymodbus.exceptions import NotImplementedException, ParameterException
 # ---------------------------------------------------------------------------#
 #  Datablock Storage
 # ---------------------------------------------------------------------------#
-class BaseModbusDataBlock(dict):
+class BaseModbusDataBlock:
     """Base class for a modbus datastore
 
     Derived classes must create the following fields:
@@ -72,12 +72,16 @@ class BaseModbusDataBlock(dict):
         :param value: The default value to set to the fields
         """
         self.default_value = value  # pylint: disable=attribute-defined-outside-init
-        self.values = [self.default_value] * count
+        self.values = [  # pylint: disable=attribute-defined-outside-init
+            self.default_value
+        ] * count
         self.address = 0x00  # pylint: disable=attribute-defined-outside-init
 
     def reset(self):
         """Reset the datastore to the initialized default value."""
-        self.values = [self.default_value] * len(self.values)
+        self.values = [  # pylint: disable=attribute-defined-outside-init
+            self.default_value
+        ] * len(self.values)
 
     def validate(self, address, count=1):
         """Check to see if the request is in range.
--- a/pymodbus/datastore/store.py
+++ b/pymodbus/datastore/store.py
@@ -51,7 +51,7 @@ from pymodbus.exceptions import NotImplementedException, ParameterException
 # ---------------------------------------------------------------------------#
 #  Datablock Storage
 # ---------------------------------------------------------------------------#
-class BaseModbusDataBlock:
+class BaseModbusDataBlock(dict):
     """Base class for a modbus datastore
 
     Derived classes must create the following fields:
@@ -72,16 +72,12 @@ class BaseModbusDataBlock:
         :param value: The default value to set to the fields
         """
         self.default_value = value  # pylint: disable=attribute-defined-outside-init
-        self.values = [  # pylint: disable=attribute-defined-outside-init
-            self.default_value
-        ] * count
+        self.values = [self.default_value] * count
         self.address = 0x00  # pylint: disable=attribute-defined-outside-init
 
     def reset(self):
         """Reset the datastore to the initialized default value."""
-        self.values = [  # pylint: disable=attribute-defined-outside-init
-            self.default_value
-        ] * len(self.values)
+        self.values = [self.default_value] * len(self.values)
 
     def validate(self, address, count=1):
         """Check to see if the request is in range.
--- a/pymodbus/server/reactive/main.py
+++ b/pymodbus/server/reactive/main.py
@@ -415,7 +415,7 @@ class ReactiveServer:
             unit = [unit]
         slaves = {}
         for i in unit:
-            block = {}
+            block = BaseModbusDataBlock()
             for modbus_entity, block_desc in data_block.items():
                 start_address = block_desc.get("block_start", 0)
                 default_count = block_desc.get("block_size", 0)
--- a/pymodbus/server/reactive/main.py
+++ b/pymodbus/server/reactive/main.py
@@ -415,7 +415,7 @@ class ReactiveServer:
             unit = [unit]
         slaves = {}
         for i in unit:
-            block = BaseModbusDataBlock()
+            block = {}
             for modbus_entity, block_desc in data_block.items():
                 start_address = block_desc.get("block_start", 0)
                 default_count = block_desc.get("block_size", 0)
--- a/pymodbus/server/reactive/main.py
+++ b/pymodbus/server/reactive/main.py
@@ -447,7 +447,7 @@ class ReactiveServer:
             if not single:
                 slaves[i] = slave_context
             else:
-                slaves = slave_context
+                slaves[0] = slave_context
         server_context = ModbusServerContext(slaves, single=single)
         return server_context
 


pymodbus/datastore/store.py
```diff
@@ -51,7 +51,7 @@
 # ---------------------------------------------------------------------------#
 #  Datablock Storage
 # ---------------------------------------------------------------------------#
-class BaseModbusDataBlock:
+class BaseModbusDataBlock(dict):
```
janiversen, line 54: Why does it inherit from a dict, it really have nothing to do with a dict.

janiversen: And in most datastores the values are not stored in a dict.

pymodbus/datastore/store.py
```diff
@@ -72,16 +72,12 @@ def default(self, count, value=False):
         :param value: The default value to set to the fields
         """
         self.default_value = value  # pylint: disable=attribute-defined-outside-init
-        self.values = [  # pylint: disable=attribute-defined-outside-init
-            self.default_value
-        ] * count
+        self.values = [self.default_value] * count
```
janiversen, line 75: This does not seem right, and you missed the previous disable !

The problem is that the base class misses an __init__ where the variables should be defined.

pymodbus/server/reactive/main.py
```diff
@@ -415,7 +415,7 @@ def create_context(
             unit = [unit]
         slaves = {}
         for i in unit:
-            block = {}
+            block = BaseModbusDataBlock()
```
janiversen, line 418: I think this is wrong, but I have to look deeper into it.

janiversen (changes requested):
Some comments.

pymodbus/datastore/store.py
janiversen: Inhering from dict would mean the following is a legal use:
```
   block = BaseModbusDataBlock()
   response = block["address"]
```
But that is not the use case, and never will be, since we never just get address, we always request address, count (using get/set/validate).

alexrudd2: The `block` dict is passed to the constructor for `ReactiveModbusslaveContext` using `**` unpacking.
https://github.com/pymodbus-dev/pymodbus/blob/031c4574eaf2c82995baab23e17ddaf9664057f2/pymodbus/server/reactive/main.py#L440-L446
That constructor takes several `BaseModbusDataBlock`s as keyword arguments.
https://github.com/pymodbus-dev/pymodbus/blob/031c4574eaf2c82995baab23e17ddaf9664057f2/pymodbus/server/reactive/main.py#L103-L116

This means that `block` must be something like `Dict[str, BaseModbusDataBlock]`

So any `block[item]` must also be `BaseModbusDataBlock`s, which means the two ways of creating an item must follow this:
https://github.com/pymodbus-dev/pymodbus/blob/031c4574eaf2c82995baab23e17ddaf9664057f2/pymodbus/server/reactive/main.py#L432-L438
(1) `block[modbus_entity]` = {... } fails because it's too broad (a dict).
(2)`block[modbus_entity] = db(...)` fails for because `db` is a too broad (also a dict).  If I type `db` to `BaseModbusDataBlock`, it cannot take the two arguments, anyways.

....somewhere around here I gave up, and made `BaseModbusDataBlock(dict)`. :)
Typing is very difficult; I will try again.

pymodbus/server/reactive/main.py
alexrudd2: On further analysis (see above), yes it's wrong.  It should be `Dict[str, BaseModbusDataBlock]`

pymodbus/datastore/store.py
alexrudd2: In this case I followed `pylint` blindly.  It warned me the disable was unnecessary.  I will revisit after rethinking `BaseModbusDataBlock(dict)`

pymodbus/datastore/store.py
janiversen: You are again mixing issues....you talk about how the datastore is defined (__init__) and then by adding dict, you redefine the usage to be as a dict, which is wrong.

Please accept the 3 views of the datastore, that might help you quite a bit.

pymodbus/server/reactive/main.py
janiversen: No it should be "{}" or "dict", it is NOT a BaseModbusDataBlock, it is the setup to define a BaseModbusDataBlock, unless I have misunderstood the code.

alexrudd2: > You are again mixing issues....you talk about how the datastore is defined (init) and then by adding dict, you redefine the usage to be as a dict, which is wrong.
> No it should be "{}" or "dict", it is NOT a BaseModbusDataBlock, it is the setup to define a BaseModbusDataBlock, unless I have misunderstood the code.

`pymodbus/server/reactive/main.py:441: error: Argument 1 to "ReactiveModbusSlaveContext" has incompatible type "**Dict[Any, Dict[Any, Any]]"; expected "BaseModbusDataBlock"  [arg-type]`

`mypy` is very strict, please try running it and you will see why I think in these ways.  It does not care about the intended use of the code, only about how broad/narrow types are.  If a variable is defined narrowly in some downstream code, then _all previous users_ of that variable must be as narrow.   They cannot be broader (which is what the error is here).  This includes the initialization; there is nothing special about it.

`block: Dict[str, BaseModbusDataBlock]` != `block: BaseModbusDataBlock.  It means a dict with keys that are `str` and values that are `BaseModbusDataBlock`.

If `block` is permitted to be any arbitrary dict, then (I think) `mypy` is concerned that some code may do something stupid.  For instance, `block['coils'] = 3`.

`BaseModbusDataBlock(dict)` was a mistake because it was too broad.  Instead I need to find a way to make the other definitions more narrow.

===

> Please accept the 3 views of the datastore, that might help you quite a bit.
Yes, it's true.  Thank you for the explanations, I am slowly learning the intent of the code. It's difficult to serve two masters, you and `mypy`.  :)

janiversen: [Argument 1 to React](error: Argument 1 to "ReactiveModbusSlaveContext") does not mean that ModbusSlaveContext should be a dict !

I am aware that mypy likes strict definition and I have NOTHING against that, but block is NOT a BaseModbusDataBlock, it is used to define one.

BaseModbusDataBlock do not inherit from anything ! that is a wrong idea. You need to add more strict definitions where data are defined for the DataBlock, which is in Init.

Be aware ReactiveModbusSlaveContext is yet another context version, look in directory datastore and you will find more. Being inherited means that it:
-  can expand the base __init__ to allow new type of definitions
-  has its own internal way of storing things.
BUT get/set/validate must remain unchanged, or it would break the server.

alexrudd2: > [Argument 1 to React](error: Argument 1 to "ReactiveModbusSlaveContext") does not mean that ModbusSlaveContext should be a dict !

When did I say this?  I explained how `block`, which is a dict, is unpacked using `**` into (a subclass of) `ModbusSlaveContext`.


> block is NOT a BaseModbusDataBlock, it is used to define one.

I said that `block` _is currently_ a dict, and the members of `block` (such as `coils`) must be `BaseModbusDataBlock` by the time they are used to initialize `ReactiveModbusSlaveContext`, which is typed to require e.g. `coils: BaseModbusDataBlock` (https://github.com/pymodbus-dev/pymodbus/commit/30f71669353114b8a00e5bf60be2e9ebccc65515).

> BaseModbusDataBlock do not inherit from anything ! that is a wrong idea. 

I am not arguing for `BaseModbusDataBlock(dict)`, and have agreed many times it's wrong. It's why I changed the PR to draft status.  I'm also not arguing for `ModbusSlaveContext(dict)`. I will revert the commit so you can stop focusing on telling me this, and perhaps try to work on a proper solution.

https://github.com/pymodbus-dev/pymodbus/blob/031c4574eaf2c82995baab23e17ddaf9664057f2/pymodbus/server/reactive/main.py#L440-L446
I'm arguing that the _unpacking of `block`_ into the the arguments for `[Reactive]ModbusSlaveContext` means that block should probably be `Dict[str, BaseModbusDataBlock]`.  

```
block: Dict[str, BaseModbusDataBlock] = {
  "coils": BaseModbusDataBlock(),
  "discrete_inputs" : BaseModbusDataBlock(),
 [...]
}
slave_context = ReactiveModbusSlaveContext( **block,  [...]) 

i.e.

slave_context = ReactiveModbusSlaveContext( 
  coils = BaseModbusDataBlock(), discrete_inputs=BaseModbusDataBlock()
) 

```

janiversen: I think it is better I let you work, and will review when it is ready. 

May I polity recommend you look at our examples to see how we use Context. BaseModbusDataBlock is a Base not a type.

alexrudd2: `/s/BaseModbusDataBlock/Type[BaseModbusDatablock]`

Perhaps that was our confusion.  It gets annoying when sometimes you need `Type[]` (for arguments), and sometimes you need class instances (in return values).
https://adamj.eu/tech/2021/05/16/python-type-hints-return-class-not-instance/.

Thank you for your patience with me.  Typing is difficult to learn in complex cases.

janiversen: No problem, we both pursue the same goal, a better library.

Thinking about Type[], but without knowing it, could this be a solution to mixin.py ?

alexrudd2: > Thinking about Type[], but without knowing it, could this be a solution to mixin.py ?

Sadly, no.  The "abstract" `execute` has a broader type, and the individual callers expect a narrower one.

alexrudd2:
closing in favor of https://github.com/pymodbus-dev/pymodbus/pull/1381

PR #1373, opened 2023-02-21
Pin setuptools to prevent breakage with Version including "X"

### LINKED ISSUE #1372: pip install does not work with setuptools >= 66.0.0 on dev branch, since the version is non-standard
https://github.com/pypa/setuptools/issues/3772

`setuptools` `66.0.0` removed support for `LegacyVersion`, which includes the current usage of `"X"` as a patch version on `dev`
https://github.com/pymodbus-dev/pymodbus/blob/031c4574eaf2c82995baab23e17ddaf9664057f2/pymodbus/version.py#L40

I originally discovered this with `pip install -e .` but it will also break the various forms of `pip install <URL of dev branch>`.
```
raise InvalidVersion(f"Invalid version: '{version}'")
pkg_resources.extern.packaging.version.InvalidVersion: Invalid version: '3.1.x'
```

_Originally posted by @alexrudd2 in https://github.com/pymodbus-dev/pymodbus/issues/1370#issuecomment-1438606512_

> I am running setuptools 65.4.1, but only for pymodbus. I have venv for each project, and setuptools got installed, when I installed pymodbus. 
> I actually wonder how setuptools got installed but I expect it is part of the python installation (which again is in the venv for pymodbus).
> apart from that, I noticed that "pip install -e ." also works with a standard python 3.11 installation.
> I have however no problem in you recommending what you think is right,, but I do have a problem changing version on dev without a release. After a release we can (as I wrote earlier) decide if we want to change the scheme.

_Originally posted by @janiversen in https://github.com/pymodbus-dev/pymodbus/pull/1370#discussion_r1113333142_
 
@janiversen, setuptools 66.0.0 was released in January, and the latest Python3.11 in January so it does not include it yet.

I propose switching to using alpha/beta release Versions on `dev`.
Current release is `Version("pymodbus", 3, 1, 3, "")`
Next release will be `Version("pymodbus", 3, 2, 0, "")`
`dev` should be `Version("pymodbus", 3, 2, 0, "alpha")`

alexrudd2:
Closes #1372


janiversen (approved):
A good compromise !

Issue #1363, opened 2023-02-19
Should non-list values be allowed for multiple writes?


alexrudd2:
The type hints and tests for `write_registers()` in `ModbusClientMixin` indicate this method only accepts lists of values.

https://github.com/pymodbus-dev/pymodbus/blob/d2c7c778707d8e0cc1ea9ea30ffccf945ea56b89/pymodbus/client/mixin.py#L422-L442

However, for at least a decade (https://github.com/pymodbus-dev/pymodbus/commit/66bd765b9de81f181ba90d817a028928d6d3edb5) the underlying `WriteMultipleRegistersRequest` will actually accept single values and convert to a list.
https://github.com/pymodbus-dev/pymodbus/blob/d2c7c778707d8e0cc1ea9ea30ffccf945ea56b89/pymodbus/register_write_message.py#L149-L160

(1) Should this functionality be retained for `write_registers()`?  
If no, we're technically breaking backwards compatibility although it's unclear if anyone uses it.  `pymodbus` itself does not appear to.
If yes, we should 
- [x] update docstrings
- [x] Type hints become (~`Union[int, float, str, List[Union[int, float, str]]]`~ `Union[List[int], int`)
- [ ] Add tests for this functionality

(2) Same for `readwrite_registers()`
(3) Same for ~`write_multiple_coils()`~ `write_coils(values: `Union[List[bool], bool])`

alexrudd2:
Does Homeassistant use `write_registers(values=123)` or `write_registers(values=[123])`?

janiversen:
Just checked the code it actually uses  [123] so demanding a list() is ok.

alexrudd2:
Do you have access to GitHub advanced code search?  I have requested access to the beta but don't have it yet.  It would be best to search for other users of these functions.

Otherwise, I could `raise DeprecationWarning` for a while?

alexrudd2:
( I think) any other users which use `mypy` will already throw errors because the type signature is `List[Union[int, float, str]]`

Also, I recently learned that `Union[float, int]` is redundant: https://stackoverflow.com/questions/50928592/mypy-type-hint-unionfloat-int-is-there-a-number-type

janiversen:
No I do not have access, I prefer using google, since it gives a much broader perspective.

I am not sure where you would raise the warning, but it should not be in production code. You have lost me why that is needed. Just define it as list[int]

janiversen:
[...]

PR #1377, opened 2023-02-21
Allow List[int] or int but not float or str

alexrudd2:
See #1363.

This does not fully close that issue, but at least corrects the type hints.

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -466,7 +466,7 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
         read_address: int = 0,
         read_count: int = 0,
         write_address: int = 0,
-        values: Union[List[int], int] = [0],
+        values: Union[List[int], int] = 0,
         slave: int = 0,
         **kwargs
     ) -> pdu_reg_read.ReadWriteMultipleRegistersResponse:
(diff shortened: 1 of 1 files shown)


alexrudd2:
@janiversen 
I'm going to leave this here to not interrupt your further work with the mixin/base.

janiversen:
@alexrudd2 this one can be merged, when you want, just change it to "ready for review", I solved a couple of issues.

janiversen:
While at it, I corrected mixin.py, so it gets a green light from mypy.

alexrudd2:
> While at it, I corrected mixin.py, so it gets a green light from mypy.

Ok, good.  I tried other options, but the only two that worked were 
(1) removing the return type from `execute()` => leaves the API untyped
(2) broadening the return types to `ModbusResponse` (which you've done)

alexrudd2:
Things brings it down from 150 to 5 `mypy` errors, all of them in `base.py` i think.  🎉

janiversen:
client/base.py is up for a major change, so it is not really worth while investing too much time there.

I know mypy allows a list of files not to consider in the configuration, using that would allow us to add mypy to CI right now, and then you can reduce the list as you go along.

janiversen:
And just for the record, the old return scope was far too limited, because there are something like 5 classes that can be returned all included in ModbusResponse.

Actually using ModbusResponse, might have a positive effect, because users see that all calls return the same struct (with a field not being filled depending on the request).

alexrudd2:
> client/base.py is up for a major change, so it is not really worth while investing too much time there.
> 
> I know mypy allows a list of files not to consider in the configuration, using that would allow us to add mypy to CI right now, and then you can reduce the list as you go along.

Yes, it's why I've not bothered.  

I will see if I can get things ready for CI (and enable `py.typed`).  In fact this week, I already tried running `mypy` on some of my applications which use `mypy` as a library and it appeared to work.

PR #1379, opened 2023-02-21
remove nosec

janiversen:
nosec is referring to a linter we no longer use.


alexrudd2:
Ahaha, I was preparing to introduce `ruff` which would include `bandit` again...

janiversen:
I am not removing bandit (which also uses nosec) we removed another tool a long time ago, because it had too many false positives.

Adding ruff is a nice plan !

janiversen:
btw there are a few places where bandit complain and I intent to correct those with this PR.

Correcting is a lot better than disabling checks.

PR #1381, opened 2023-02-21
Fix mypy errors in reactive server

alexrudd2:
Supercedes https://github.com/pymodbus-dev/pymodbus/pull/1371

It turns out the correct answer was, in fact, cleaner code instead of trying to abuse `mypy`. :)


janiversen (approved):
LGTM. Simplifying code is a special art 🚀

PR #1383, opened 2023-02-21
Remove unneeded attribute from version

alexrudd2:
Fixes `pymodbus/version.py:41: error: "Version" has no attribute "__name__"  [attr-defined]`

the comment refers to "epydoc", which has not been used for a long time.  I could find no other use.


janiversen (approved):
lgtm.

Issue #1360, opened 2023-02-19
Set level for mypy.


janiversen:
### Versions

* Python:
* OS:
* Pymodbus:
* Modbus Hardware (if used):

### Pymodbus Specific
* Server: tcp/rtu/ascii - sync/async
* Client: tcp/rtu/ascii - sync/async

### Description

What were you trying, what has happened, what went wrong, and what did you expect?

### Code and Logs

```python
# code and logs here.

# please use the following to format logs when posting them here
import logging
import pymodbus
pymodbus.pymodbus_apply_logging_config(logging.DEBUG)
```

Once mypy is properly implemented (thanks to @alexrudd2), we should get to a level similar to:
https://github.com/home-assistant/core/blob/dev/mypy.ini

alexrudd2:
> Then I suggest (but you decide):
> - activate the options that are (relative) easy to solve
> - keep the rest deactivated
> - Group the ones you think we should activate for in a interface module
> - Group the rest as unwanted
> I have done a similar thing in the pylint setup, and it helps me remember that an option was tried and discarded etc.
> 
> You do not need to do it all in this PR, just put it up for review when you want.

_Originally posted by @janiversen in https://github.com/pymodbus-dev/pymodbus/issues/1392#issuecomment-1446612777_

janiversen:
I think we can safely close this now.

PR #1388, opened 2023-02-24
Enable mypy in CI

alexrudd2:
Three typing errors remain in with the default `mypy` config.  They are all in `base.py` and can be ignored since it will be refactored soon.

By setting `py.typed` in the module folder, other programs built on top of `pymodbus` can use its typing.  This works locally on my system when installed with `python3 -m pip install -e .`. I'm not sure if this will happen automatically when installed from PyPI; we may need to look at the `typeshed` project.

Current config:
```
[mypy]
strict_optional = False
````

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,9 @@ jobs:
           - name: pytest
             cmd: pytest --cov=pymodbus --cov=test --cov-report=term-missing --cov-report=xml -v --full-trace --timeout=20
             type: test
+          - name: mypy
+            cmd: mypy pymodbus
+            type: lint
         os:
           - name: Linux
             on: ubuntu-latest
--- a/check_ci.sh
+++ b/check_ci.sh
@@ -10,5 +10,6 @@ black --safe --quiet examples/ pymodbus/ test/
 isort .
 pylint --recursive=y examples pymodbus test
 flake8
+mypy pymodbus
 pytest --numprocesses auto
 echo "Ready to push"
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -121,7 +121,7 @@ class ModbusBaseClient(ModbusClientMixin):
         self.params.kwargs = kwargs
 
         # Common variables.
-        self.framer = self.params.framer(ClientDecoder(), self)
+        self.framer = self.params.framer(ClientDecoder(), self)  # type: ignore
         self.transaction = DictTransactionManager(self, **kwargs)
         self.delay_ms = self.params.reconnect_delay
         self.use_protocol = False
@@ -147,7 +147,7 @@ class ModbusBaseClient(ModbusClientMixin):
         Use register() to add non-standard responses (like e.g. a login prompt) and
         have them interpreted automatically.
         """
-        self.framer.decoder.register(custom_response_class)
+        self.framer.decoder.register(custom_response_class)  # type: ignore
 
     def connect(self) -> None:
         """Connect to the modbus remote host (call **sync/async**).
@@ -187,7 +187,7 @@ class ModbusBaseClient(ModbusClientMixin):
             if not self._connected:
                 raise ConnectionException(f"Not connected[{str(self)}]")
             return self.async_execute(request)
-        if not self.connect():
+        if not self.connect():  # type: ignore
             raise ConnectionException(f"Failed to connect[{str(self)}]")
         return self.transaction.execute(request)
 

### CODE REVIEW

**Review by alexrudd2 — COMMENTED** [review]
@janiversen  I know you did not like adding adding type stubs as development dependencies, but I don't see another way to make it work.

**Comment by janiversen** [comment]
Please have a look at the config I sent you https://github.com/pymodbus-dev/pymodbus/issues/1360

Home Assistant uses mypy in a way where whole files are ignored, we should copy that way, otherwise e.g. base.py will continue to give problems (I will be making a lot of changes there, and do not want to fight with mypy every time).

As far as I can see, we simply need to add the files to mypy.ini (keeping it for now at the level you want, by commenting out the parts not used now). By adding the files we remove the need for stubs, something I feel rather strongly about.

The reason I am against stubs, if due to maintenance, the stub have its own version, meaning when we upgrade de main e.g. pyserial, we also need to secure that the stub is upgrade...a real PITA !!!

Mypy is something (very) nice to have but is not mandatory, so it should not cause more maintenance (as with stubs).

File: pymodbus/client/base.py [hunk 1]
```diff
@@ -121,7 +121,7 @@ def __init__(
         self.params.kwargs = kwargs
 
         # Common variables.
-        self.framer = self.params.framer(ClientDecoder(), self)
+        self.framer = self.params.framer(ClientDecoder(), self)  # type: ignore
```
janiversen, line 124: Please have the whole file ignored, since there will be a lot of changes over the next short time.

check_ci.sh
```diff
@@ -10,5 +10,6 @@ black --safe --quiet examples/ pymodbus/ test/
 isort .
 pylint --recursive=y examples pymodbus test
 flake8
+mypy pymodbus
```
janiversen, line 13: Please add a mypy.ini file, so we can exclude files and also track how far we are in respect of changing the source to be typed.

requirements.txt
```diff
@@ -70,3 +70,6 @@ pytest-asyncio==0.20.3
 pytest-cov==4.0.0
 pytest-timeout==2.1.0
 pytest-xdist==3.1.0
+types-redis
```
janiversen, line 73: We only add these to please mypy, that is not the way to go (see other comments).

janiversen (changes requested):
what does the file py.typed do ? I cannot see references and it is empty.

.github/workflows/ci.yml
```diff
@@ -49,6 +49,9 @@ jobs:
           - name: pytest
             cmd: pytest --cov=pymodbus --cov=test --cov-report=term-missing --cov-report=xml -v --full-trace --timeout=20
             type: test
+          - name: mypy
```
janiversen, line 52: Please move this up to black etc, so we group all linters together, and tests together.

janiversen (changes requested):

check_ci.sh
alexrudd2: https://github.com/pymodbus-dev/pymodbus/blob/dev/setup.cfg#L504

It already exists as a section in `setup.cfg`, no need for a separate file.

alexrudd2:
> what does the file py.typed do ? I cannot see references and it is empty.

https://peps.python.org/pep-0561/

It's supposed to be empty, indicating that the module is typed, allowing other users/applications of the module to so static analysis across modules works.

alexrudd2:
> Please have a look at the config I sent you #1360
> 
> Home Assistant uses mypy in a way where whole files are ignored, we should copy that way, otherwise e.g. base.py will continue to give problems (I will be making a lot of changes there, and do not want to fight with mypy every time).

I looked at that config.  There is currently no need to ignore files since `mypy` reports no errors (ignoring the 3 in `base.py`).  The HomeAssistant config is more strict.  Enabling the stricter config would create hundreds of errors ( I checked).  It can be done later, one rule at a time.

> As far as I can see, we simply need to add the files to mypy.ini (keeping it for now at the level you want, by commenting out the parts not used now). By adding the files we remove the need for stubs, something I feel rather strongly about.

With this PR `mypy` runs with 0 errors against all pymodbus files.  The stubs are for other libraries we use.

> The reason I am against stubs, if due to maintenance, the stub have its own version, meaning when we upgrade de main e.g. pyserial, we also need to secure that the stub is upgrade...a real PITA !!!
> Mypy is something (very) nice to have but is not mandatory, so it should not cause more maintenance (as with stubs).

The stub packages are automatically generated by a bot in [`typeshed`](https://github.com/python/typeshed).

_These PyPI packages follow [PEP 561](http://www.python.org/dev/peps/pep-0561/) and are automatically released (multiple times a day, when needed) by [typeshed internal machinery](https://github.com/typeshed-internal/stub_uploader)._

`pyserial` releases roughly yearly, so it's not like we need to update the stubs package very frequently.  Besides, if there is an update to the stubs, _it will also be an update to the `pyserial` function types/signatures_.  This means that any code using those signatures would break anyway, no?  The alternative is to wait for the bug to occur at runtime!

pymodbus/client/base.py
alexrudd2: I think it makes sense for whomever is working on a file to
(1) don't make type errors OR
(2) disable the type checking temporarily during the refactor.

This can be done by adding `# type: ignore` of the file, I believe.

alexrudd2:
> https://peps.python.org/pep-0561/

For instance, here is the blank file in `click` that allows mypy to check pymodbus' usage of that library without installing stubs.

https://github.com/pallets/click/blob/main/src/click/py.typed

pymodbus/client/base.py
janiversen: Then please do that.

janiversen:
I did not know about py.typed, that is a good idea to add. 

I disagree with the stubs, but will not block the PR on that (my opinion is no better than your opinion).

So once base.py is clear I will approve the PR, The best option is to ignore the file in the configuration, because that way we can easily see what is being ignored.

Ignoring files in the configuration, also make it more transparent to activate new checks, when activating the check, the number of ignored files increases for a short while.

janiversen:
Let me just explain in detail why base.py as file should be ignored.

Whenever an ignore (that be flake/pylint or mypy) is used a line level, it signal that the code is correct and cannot be changed to avoid the ignore (like e.g. pylint “too many methods”). In case we do not want to refactor the code then it is better to ignore the whole file as a clear statement. Before I add an ignore I try to see if the code can be refactored.

We do not want to refactor base.py right now (because that is being done in another PR), hence ignore the file.

If we start adding ignore just to get green light from mypy, then the whole exercise is pointless, because the purpose is to get better and more stable code.

Adding an ignore in the top of the file, which should only be used when the tool do not allow an ignore list in the setup. The reason is that an ignore list is very visible, whereas an ignore in a file is just one more.

alexrudd2:
> If we start adding ignore just to get green light from mypy, then the whole exercise is pointless, because the purpose is to get better and more stable code.

After the refactor, someone will have to go through and fix the mypy errors anyways.  Actually, it was just easier to fix them now.   (If you refactor again, OK, it didn't take long).

I've left this below in the config so you can see how to ignore files.  It's not needed currently.
```
[mypy]
# exclude = pymodbus/client/base.py
```

Does this work for you?

pymodbus/client/base.py
alexrudd2: I've left this below in the config so you can see how to ignore files.  It's not needed currently.
```
[mypy]
# exclude = pymodbus/client/base.py
```

Does this work for you?

pymodbus/client/base.py
```diff
@@ -149,7 +149,7 @@ def register(self, custom_response_class: ModbusResponse) -> None:
         """
         self.framer.decoder.register(custom_response_class)
 
-    def connect(self) -> None:
+    def connect(self) -> bool:
```
janiversen, line 152: This is not correct, at the moment it depends on the implementation, some return e.g. transport, protocol, sometimes bool and sometimes none.

The correct one is None, because the user should not check it.

This btw is one of the reasons why you should not bother checking base.py, it is full of such errors.

setup.cfg
```diff
@@ -503,6 +503,7 @@ noqa-require-code = True
 
 [mypy]
 strict_optional = False
+# exclude = pymodbus/client/base.py
```
janiversen, line 506:
```suggestion
exclude = pymodbus/client/base.py  
```

janiversen (changes requested):

setup.cfg
alexrudd2: Hahah, you are very insistent.

pymodbus/client/base.py
janiversen: There are no correct answer at the moment, base.py is pure spaghetti and a mix of old/new methods which on top sometimes are async and sometimes not.

Trying to add types in this setup is a sure way to disaster ! Saying connect is a bool, is just as false as None, just leave it untyped, and exclude the whole file.

janiversen: and None is correct !! but of course the code needs to be adapted for that, which is exactly what I am doing.

janiversen (approved):
LGTM.

PR #1392, opened 2023-02-27
Expand mypy config, following HomeAssistant

alexrudd2:
Builds on #1388

See #1360

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -4,6 +4,7 @@ import asyncio
 import ssl
 import time
 import traceback
+from typing import Union
 
 from pymodbus.client.serial_asyncio import create_serial_connection
 from pymodbus.constants import Defaults
@@ -873,7 +874,7 @@ class ModbusSerialServer:  # pylint: disable=too-many-instance-attributes
     server context instance.
     """
 
-    handler = None
+    handler: ModbusSingleRequestHandler = None
 
     def __init__(
         self, context, framer=ModbusRtuFramer, identity=None, **kwargs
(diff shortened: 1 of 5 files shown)


janiversen:
Looks good. Hope the most parts can be done without tooo much editing.

alexrudd2:
> Looks good. Hope the most parts can be done without tooo much editing.

Enabling all the rules gives ~1000 errors, mostly `disallow_untyped_defs`.  I think we should more cleanly define internal versus external methods, since external ones are more important to be typed correctly.

janiversen:
That is a plan I highly agree with. As I work through the code, all real externals will be or are defined in __init__ as external symbols, that can at least to identify the external interfaces.

Can e.g. "disallow_untyped_defs" be restricted to a module ? if so, we can consider making a API module, and have all external classes (pure interface classes there). That does not sound difficult to do, and will just affect the import statement in the apps.

janiversen:
Once you update your branch (also in the other PRs) you will possibly see more problems.

alexrudd2:
> Can e.g. "disallow_untyped_defs" be restricted to a module ? if so, we can consider making a API module, and have all external classes (pure interface classes there). That does not sound difficult to do, and will just affect the import statement in the apps.

I believe so, yes.

janiversen:
and you have a couple of conflicts to solve...

alexrudd2:
> and you have a couple of conflicts to solve...

Yes, I am rebasing now locally

pymodbus/server/async_io.py
```diff
@@ -873,7 +874,7 @@ class ModbusSerialServer:  # pylint: disable=too-many-instance-attributes
     server context instance.
     """
 
-    handler = None
+    handler: ModbusSingleRequestHandler = None
```
janiversen, line 877: Handler can be all 3 handlers, but I am ok with only typing one of them.

janiversen (approved):
LGTM.

PR #1399, opened 2023-03-01
Narrow types for execute using @overload

alexrudd2:
Using `@overload` to define all the possible Request->Response combinations can successfully narrow e.g. `ModbusRequest` to the actual types.

However, it's very verbose and probably not worthwhile...


janiversen:
At a first glance, this is something I am quite unhappy about, because it is a lot of extra to maintain, and error prone, because it is different lists.

alexrudd2:
Yes, I agree.  I don't think it's worth it, but wanted to see if it can be done.  Perhaps the technique can be used elsewhere where it makes more sense.

Issue #1331, opened 2023-02-04
Parameter retry_on_empty and retries does not have any effect for serial client


jara1:
### Versions

* Python: 3.9.2
* OS: Debian
* Pymodbus: 3.1.2
* Modbus Hardware (if used): FTDI FT2232H

### Pymodbus Specific
 Client: rtu - sync

### Description
Parameter ```retry_on_empty``` and ```retries``` does not have any effect. Based on my debugging, it's not used from client context in transaction.py

As a result, there is no retry attempt.

### Code and Logs

```python
from pymodbus.client import ModbusSerialClient
import time
import logging
import pymodbus
pymodbus.pymodbus_apply_logging_config(logging.DEBUG)
client = ModbusSerialClient(method='rtu', port='/dev/ttyUSB1', baudrate=115200, timeout= 5, retry_on_empty=True, retries=10, debug=True)
client.connect()
while(1):
    rr = client.read_input_registers(address=0x3100, slave=1, count=1)
    print(rr)
    time.sleep(0.5)

client.close()


23:02:24 DEBUG logging:80 Running transaction 7
23:02:24 DEBUG logging:80 SEND: 0x1 0x4 0x31 0x0 0x0 0x1 0x3f 0x36
23:02:24 DEBUG logging:80 Changing state to IDLE - Last Frame End - 1675548144.052484 Current Time stamp - 1675548144.554936
23:02:24 DEBUG logging:80 New Transaction state "SENDING"
23:02:24 DEBUG logging:80 Changing transaction state from "SENDING" to "WAITING FOR REPLY"
23:02:34 DEBUG logging:80 Transaction failed. (Modbus Error: [Invalid Message] No response received, expected at least 4 bytes (0 received)) 
23:02:34 DEBUG logging:80 Frame - [b''] not ready
23:02:34 DEBUG logging:80 Getting transaction 1
23:02:34 DEBUG logging:80 Changing transaction state from "PROCESSING REPLY" to "TRANSACTION_COMPLETE"
Modbus Error: [Input/Output] Modbus Error: [Invalid Message] No response received, expected at least 4 bytes (0 received)
23:02:35 DEBUG logging:80 Current transaction state - TRANSACTION_COMPLETE
```

janiversen:
hmm I thought that was fixed, let me take a look.

janiversen:
Corrected on dev.

PR #1400, opened 2023-03-02
Forward retry_on_empty and retries to serial

cgernert:
Fixes https://github.com/pymodbus-dev/pymodbus/issues/1331

### DIFF
--- a/pymodbus/transaction.py
+++ b/pymodbus/transaction.py
@@ -50,9 +50,9 @@ class ModbusTransactionManager:
         self.tid = Defaults.TransactionId
         self.client = client
         self.backoff = kwargs.get("backoff", Defaults.Backoff) or 0.3
-        self.retry_on_empty = kwargs.get("retry_on_empty", Defaults.RetryOnEmpty)
+        self.retry_on_empty = client.params.retry_on_empty
         self.retry_on_invalid = kwargs.get("retry_on_invalid", Defaults.RetryOnInvalid)
-        self.retries = kwargs.get("retries", Defaults.Retries) or 1
+        self.retries = client.params.retries
         self.reset_socket = kwargs.get("reset_socket", True)
         self._transaction_lock = RLock()
         self._no_response_devices = []


janiversen:
You change from reading the external parameters to instead use the stored params, but the stored parameters are not ready, those are assigned in the base class.

cgernert:
> You change from reading the external parameters to instead use the stored params, but the stored parameters are not ready, those are assigned in the base class.

Yeah, I debugged it and don't get the Python code. 
The constructor in base.py seems to remove the parameters from kwargs list, but I don't know why:

      def __init__(
        self,
        framer: str = None,
        timeout: str | float = Defaults.Timeout,
        retries: str | int = Defaults.Retries,
        retry_on_empty: bool = Defaults.RetryOnEmpty,
        close_comm_on_error: bool = Defaults.CloseCommOnError,
        strict: bool = Defaults.Strict,
        broadcast_enable: bool = Defaults.BroadcastEnable,
        reconnect_delay: int = Defaults.ReconnectDelay,
        reconnect_delay_max: int = Defaults.ReconnectDelayMax,
        **kwargs: any,
      )...

janiversen:
????
it removes it, because it uses it.
´´´
        self.params.retry_on_empty = bool(retry_on_empty)
´´´
all parameters used are normally declared, while the rest are kwargs.

janiversen:
Anyhow you must have debugged it very lightly, since:
```
>       self.retry_on_empty = client.params.retry_on_empty
[6042](https://github.com/pymodbus-dev/pymodbus/actions/runs/4314030700/jobs/7527285279#step:6:6043)
E       AttributeError: 'NoneType' object has no attribute 'params'
```
this is a pretty clear signal, that your change was not correct.

cgernert:
> ???? it removes it, because it uses it. ´´´ self.params.retry_on_empty = bool(retry_on_empty) ´´´ all parameters used are normally declared, while the rest are kwargs.

Ah, OK. I don't write Python often. Anyway, that's the source of the problem. I will create another pull request.

PR #1402, opened 2023-03-03
Add tests for writing multiple writes with a single value

### LINKED ISSUE #1363: Should non-list values be allowed for multiple writes?

alexrudd2:
Closes #1363

Currently there are only tests for e.g. `write_coils([list])`. This adds them for `write_coils(bool)`.

See also #1377

### DIFF
--- a/test/test_bit_write_messages.py
+++ b/test/test_bit_write_messages.py
@@ -43,6 +43,8 @@ class ModbusBitMessageTests(unittest.TestCase):
             WriteSingleCoilResponse(1, 0xABCD): b"\x00\x01\xff\x00",
             WriteMultipleCoilsRequest(1, [True] * 5): b"\x00\x01\x00\x05\x01\x1f",
             WriteMultipleCoilsResponse(1, 5): b"\x00\x01\x00\x05",
+            WriteMultipleCoilsRequest(1, True): b"\x00\x01\x00\x01\x01\x01",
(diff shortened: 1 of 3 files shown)


janiversen (approved):
LGTM, thanks.

PR #1406, opened 2023-03-03
Add pre-commit config

alexrudd2:
Works for me, adds about 0.3s to commit times.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,21 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
+    # -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: check-added-large-files
+-   repo: https://github.com/pycqa/isort
+    rev: 5.12.0
+    hooks:
+    -   id: isort
+        exclude: ^(doc/_build|venv|.venv|.git|pymodbus/client/serial_asyncio)
+-   repo: https://github.com/psf/black
+    rev: 22.12.0
+    hooks:
+    -   id: black
+        args: [--safe,--quiet]
+        files: (examples|pymodbus|test)/
--- a/examples/v2.5.3/rx_messages
+++ b/examples/v2.5.3/rx_messages
@@ -88,7 +88,7 @@
 # Modbus RTU Messages
 # ------------------------------------------------------------
 # [Address ][ Function Code] [ Data ][ CRC ]
-#  1b         1b               Nb      2b   
+#  1b         1b               Nb      2b
 #
 # ./message-parser -b -p rtu -f messages
 # ------------------------------------------------------------
--- a/requirements.txt
+++ b/requirements.txt
@@ -61,6 +61,7 @@ flake8-noqa==1.3.0
 flake8-comprehensions==3.10.1
 isort==5.11.4
 mypy==1.0.1
+pre-commit==3.1.1
 pyflakes==3.0.1
 pydocstyle==6.3.0
 pycodestyle==2.10.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -52,14 +52,12 @@ sphinx-rtd-theme==1.1.1
 # -------------------------------------------------------------------
 # install:development
 bandit==1.7.4
-black==22.12.0
 codespell==2.2.2
 coverage==7.1.0
 flake8==6.0.0
 flake8-docstrings==1.7.0
 flake8-noqa==1.3.0
 flake8-comprehensions==3.10.1
-isort==5.11.4
 mypy==1.0.1
 pre-commit==3.1.1
 pyflakes==3.0.1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,11 +37,8 @@ jobs:
           - name: flake8
             cmd: flake8
             type: lint
-          - name: isort
-            cmd: isort --check .
-            type: lint
-          - name: black
-            cmd: black --check --safe --quiet examples/ pymodbus/ test/
+          - name: precommit (isort and black)
+            cmd: pre-commit run --all-files
             type: lint
           - name: docs
             cmd: make -C doc/ html
--- a/README.rst
+++ b/README.rst
@@ -224,7 +224,9 @@ Then:
 
    pip install -e .
 
-This installs pymodbus in your virtual environment with pointers directly to the pymodbus directory, so any change you make is immediately available as if installed.
+   pre-commit --install
+
+This installs pymodbus in your virtual environment with pointers directly to the pymodbus directory, so any change you make is immediately available as if installed.  It will also install `pre-commit` git hooks.
 
 Either method will install all the required dependencies
 (at their appropriate versions) for your current python distribution.
--- a/TODO_add_checks
+++ b/TODO_add_checks
@@ -1,6 +1,2 @@
-1) add .pre-commit and commit-hook
-
-2) new checks
+New checks
 - prettier
-- mypy
-
--- a/check_ci.sh
+++ b/check_ci.sh
@@ -6,8 +6,7 @@ trap 'echo "\"${last_command}\" command filed with exit code $?."' EXIT
 
 
 codespell
-black --safe --quiet examples/ pymodbus/ test/
-isort .
+pre-commit run --all-files
 pylint --recursive=y examples pymodbus test
 flake8
 mypy pymodbus
--- a/examples/v2.5.3/tx_messages
+++ b/examples/v2.5.3/tx_messages
@@ -88,7 +88,7 @@
 # Modbus RTU Messages
 # ------------------------------------------------------------
 # [Address ][ Function Code] [ Data ][ CRC ]
-#  1b         1b               Nb      2b   
+#  1b         1b               Nb      2b
 #
 # ./message-parser -b -p rtu -f messages
 # ------------------------------------------------------------
--- a/pymodbus/repl/client/main.py
+++ b/pymodbus/repl/client/main.py
@@ -231,7 +231,7 @@ def cli(client):  # pylint: disable=too-complex
             continue  # Control-C pressed. Try again.
         except EOFError:
             break  # Control-D pressed.
-        except Exception as exc:  # Handle all other exceptions pylint: disable=broad-except
+        except Exception as exc:  # pylint: disable=broad-except
             click.secho(str(exc), fg="red")
 
     click.secho("GoodBye!", fg="blue")


janiversen (changes requested):
It seems you activated some new checks, please make sure that CI have the same checks. .precommit should be a subset of CI, not the other way round.

janiversen:
Seems CI do not like the pre commit version

janiversen:
Let me just explain why it is important that pre-commit is a subset and not a superset.

I often make small changes to a PR directly in github, and some users make pull requests without having the full development environment installed.

If precommit makes checks that CI does not, these would not be done in the above cases, and we would then run the risk of seeing the effect on the next commit we make.

I do believe running black and isort at least in a precommit is a very good thing!

You have removed empty lines at the end in some files, I hope that is not a demand because I prefer to end the file with an empty line (and some of my editors actually do it automatically).

You need to update dev.

alexrudd2:
> Seems CI do not like the pre commit version

Fixed the typo in `requirements.txt`

> You have removed empty lines at the end in some files, I hope that is not a demand because I prefer to end the file with an empty line (and some of my editors actually do it automatically).

This was the pre-commit default (`end-of-file-fixer`).  I'm surprised your editor adds an entire empty line (i.e. two newlines after the last character; mine only adds single newline.). Nevertheless, I disabled that hook.

> You need to update dev.

I use rebase + `--force-with-lease`, you merge `dev` into the branch.  They have the same effect :). There was no merge conflict anyways.

examples/v2.5.3/rx_messages
```diff
@@ -88,7 +88,7 @@
 # Modbus RTU Messages
 # ------------------------------------------------------------
 # [Address ][ Function Code] [ Data ][ CRC ]
-#  1b         1b               Nb      2b   
+#  1b         1b               Nb      2b
```
janiversen, line 91: these changes worries me, I cannot get pylint/black/flake8 to report them (with our CI setup).

Have you managed to get an error (not in the pre-commit hook) ?

pymodbus/client/serial_asyncio/__init__.py
```diff
@@ -18,6 +18,7 @@
 import asyncio
 import os
 
+
```
janiversen, line 21: This is an imported package, we do not makes changes in here, unless there is a corresponding pull request in their repo.

The only reason we have it, is because they take a very long time in responding to pull requests (I have filed 2), so it was imported to solve the problem at hand.

We do not do "cosmetic" changes in their source.

This passes without problems in our current CI, so this makes me even more afraid.

requirements.txt
```diff
@@ -61,6 +61,7 @@ flake8-noqa==1.3.0
 flake8-comprehensions==3.10.1
 isort==5.11.4
 mypy==1.0.1
+pre-commit==3.1.1
```
janiversen, line 64: You are adding this, but I do not see that it is installed as a hook ??

janiversen (changes requested):
The biggest problem is that precommit catches things that our CI do not, that is the wrong way round. A precommit must be a subset (or equal) to CI, because it is not guaranteed that a pull request have passed the pre-commit.

janiversen:
Please do not misunderstand, I am all for a precommit hook, I have had a hook installed that does pylint/black/flake8 and sort for a long time.

But if we add it to the repo, we need to ensure there are no unexpected effects.

janiversen:
> > Seems CI do not like the pre commit version
> 
> Fixed the typo in `requirements.txt`
> 
> > You have removed empty lines at the end in some files, I hope that is not a demand because I prefer to end the file with an empty line (and some of my editors actually do it automatically).
> 
> This was the pre-commit default (`end-of-file-fixer`). I'm surprised your editor adds an entire empty line (i.e. two newlines after the last character; mine only adds single newline.). Nevertheless, I disabled that hook.
> 
> > You need to update dev.
> 
> I use rebase + `--force-with-lease`, you merge `dev` into the branch. They have the same effect :). There was no merge conflict anyways.

The difference is that when GitHub claim "This branch is out-of-date with the base branch", the PR cannot be merged.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
...
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
+    # -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: check-added-large-files
+-   repo: https://github.com/pycqa/isort
+    rev: 5.12.0
```
janiversen, line 12: Do we really need to specify the version again, this is very prune to errors !

Have you considered not using pre-commit, but simply use .git/hooks/pre-commit.sample to run isort etc (which are already installed).

janiversen (changes requested):
Just a thought.

examples/v2.5.3/rx_messages
alexrudd2: Strange.  `black` does remove the trailing whitespace, but only if you feed it a file directly.
```
❯ black examples/
All done! ✨ 🍰 ✨
44 files left unchanged.

❯ black examples/v2.5.3/rx_messages
reformatted examples/v2.5.3/rx_messages

All done! ✨ 🍰 ✨
1 file reformatted.
```

requirements.txt
alexrudd2: There is no way for a git server to distribute/enforce client-side hooks. (It would be a security issue).  Therefore, they must be [installed](https://pre-commit.com/#install) manually.  

In this case, with `pre-commit install`.

This PR does nothing to the CI, nor to developers that don't install pre-commit.

alexrudd2:
> The difference is that when GitHub claim "This branch is out-of-date with the base branch", the PR cannot be merged.

This is an adjustable [Branch protection rule](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches), and not a GitHub default.

I neither have access to this setting, nor care what its value is for this repo.

requirements.txt
janiversen: it should and can be done in setup.py

pymodbus/client/serial_asyncio/__init__.py
janiversen: Yes of course it is disabled, because we do not want that directory touched. precommit must follow what we have already configured.

.pre-commit-config.yaml
alexrudd2: > Do we really need to specify the version again, this is very prune to errors !

I followed the [official directions](https://pycqa.github.io/isort/docs/configuration/pre-commit.html) for `isort`, but updated the version to avoid https://github.com/PyCQA/isort/issues/2077.

We could consider using `renovatebot`, which can [supposedly](https://docs.renovatebot.com/modules/manager/pre-commit/) update `pre-commit` automatically.  (I find it much smarter than Dependabot, generally)

> Have you considered not using pre-commit, but simply use .git/hooks/pre-commit.sample to run isort etc (which are already installed).

Manually installing hooks defeats the purpose of `pre-commit`.  Ultimately it's a matter of preference, I guess.  If you don't want a shared config we can close this and add it to `.gitignore`.

janiversen: I have no problem using pre-commit, but I do have a problem with duplicated information, especially versions, that is bound to give problems. I installed the hook manually because that was the fastest for me (a simple copy) not because it is the right thing to do.

If we use pre-commit, then why have flake8 etc in requirements.txt, that makes little sense to me and it would solve the potential version conflict.

But we still need to make sure that if CI-Black is giving green light then so must pre-commit. I had for a long time the problem, that my local pylint reported a problem, and when I corrected that CI-pylint reported another problem (in the same lines), I do not want to experience that again.

examples/v2.5.3/rx_messages
janiversen: you can configure black in the yaml file, like e.g.:
```
- repo: https://github.com/psf/black
    rev: 23.1.0
    hooks:
      - id: black
        args:
          - --quiet
        files: ^((homeassistant|pylint|script|tests)/.+)?[^/]+\.py$
```
and use the same arguments as CI uses.

.pre-commit-config.yaml
jamesbraza: The solution here is to move CI to use `pre-commit`, and remove `flake8`/`isort`/`black`/etc. from `requirements.txt`.

That way:
- `pre-commit` solves the toolchain problem: how to (1) co-version tools, (2) standard method of kicking them all off simultaneously
- Configs still live in `setup.cfg`, as earlier
- An identical workflow used in local dev is used in CI (really beneficial)

requirements.txt
jamesbraza: Package installation should not have any idea about the existence of repo-level QA tooling.  QA tooling should be installed externally from `setup.py`.

Imo the CI workflow should invoke `pre-commit install-hooks` and then cache `~/.cache/pre-commit`.  Or you can use the `pre-commit` GitHub Action to automate this

jamesbraza:
Happy to see this PR!

requirements.txt
janiversen: That is a fair argument, but since it needs to be installed manually, the very least to do is to update README.rst to reflect that.

.pre-commit-config.yaml
janiversen: That is a good alternative, I like that. README.rst should still be updated to inform new developers how to setup their environment.

janiversen:
> > The difference is that when GitHub claim "This branch is out-of-date with the base branch", the PR cannot be merged.
> 
> This is an adjustable [Branch protection rule](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches), and not a GitHub default.
> 
> I neither have access to this setting, nor care what its value is for this repo.

You should care, not about the setting, but about the effect. This setting is enabled to ensure every PR is tested against the current dev, which is a simply way to avoid conflicts on dev. Locally I do not merge but run “git rebase dev”, but github only offers the merge option, which is why you sometimes see a merge commit.

janiversen:
I am starting to think about releasing 3.2, it would be nice to have this merged before.

.pre-commit-config.yaml
alexrudd2: > The solution here is to move CI to use `pre-commit`, and remove `flake8`/`isort`/`black`/etc. from `requirements.txt`.

Thank you, it really solved our problems here!

alexrudd2:
> > > The difference is that when GitHub claim "This branch is out-of-date with the base branch", the PR cannot be merged.
> > 
> > 
> > This is an adjustable [Branch protection rule](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches), and not a GitHub default.
> > I neither have access to this setting, nor care what its value is for this repo.
> 
> You should care, not about the setting, but about the effect. This setting is enabled to ensure every PR is tested against the current dev, which is a simply way to avoid conflicts on dev. Locally I do not merge but run “git rebase dev”, but github only offers the merge option, which is why you sometimes see a merge commit.

Yes, I understand the effect.  As I said, I use `git rebase dev && git push --force-with-lease` which is the same as you merging via GitHub.

janiversen (approved):
LGTM. Thanks

PR #1409, opened 2023-03-06
Logging configurability

jamesbraza:
- Adds the ability to override `logging.basicConfig` parameters
- Adds type hints to the signature
- Fixes minor wordings in docstring

### DIFF
--- a/pymodbus/__init__.py
+++ b/pymodbus/__init__.py
@@ -1,6 +1,6 @@
 """Pymodbus: Modbus Protocol Implementation.
 
-Released under the the BSD license
+Released under the BSD license.
 """
 
 from logging import WARNING
@@ -14,9 +14,10 @@ __author__ = "Galen Collins, Jan Iversen"
 __maintainer__ = "dhoomakethu, janiversen"
 
 
-def pymodbus_apply_logging_config(level=WARNING):
-    """Apply basic logging configuration used by default by Pymodbus maintainers.
+def pymodbus_apply_logging_config(level: int = WARNING, **config_overrides) -> None:
+    """Apply basic logging configuration.
 
-    Please call this function to format logging appropriately when opening issues.
+    Please call this function to format logging appropriately when opening
+    issues.  Default formatting is specified by the Pymodbus maintainers.
     """
-    Log.apply_logging_config(level)
+    Log.apply_logging_config(level, **config_overrides)
--- a/pymodbus/__init__.py
+++ b/pymodbus/__init__.py
@@ -14,7 +14,9 @@ __author__ = "Galen Collins, Jan Iversen"
 __maintainer__ = "dhoomakethu, janiversen"
 
 
-def pymodbus_apply_logging_config(level: int = WARNING, **config_overrides) -> None:
+def pymodbus_apply_logging_config(
+    level: int = WARNING, **config_overrides: dict
+) -> None:
     """Apply basic logging configuration.
 
     Please call this function to format logging appropriately when opening
--- a/pymodbus/logging.py
+++ b/pymodbus/logging.py
@@ -40,7 +40,8 @@ class Log:
             "format": "%(asctime)s %(levelname)-5s %(module)s:%(lineno)s %(message)s",
             "datefmt": "%H:%M:%S",
         }
-        logging.basicConfig(**{**default_config, **config_overrides})
+        config_kwargs = {**default_config, **config_overrides}
+        logging.basicConfig(**config_kwargs)
         cls._logger.setLevel(level)
         cls.LOG_LEVEL = cls._logger.getEffectiveLevel()
 
--- a/pymodbus/logging.py
+++ b/pymodbus/logging.py
@@ -1,6 +1,6 @@
 """Pymodbus: Modbus Protocol Implementation.
 
-Released under the the BSD license
+Released under the BSD license.
 """
 
 import logging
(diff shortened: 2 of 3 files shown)


pymodbus/__init__.py
```diff
@@ -14,9 +14,10 @@
 __maintainer__ = "dhoomakethu, janiversen"
 
 
-def pymodbus_apply_logging_config(level=WARNING):
-    """Apply basic logging configuration used by default by Pymodbus maintainers.
+def pymodbus_apply_logging_config(level: int = WARNING, **config_overrides) -> None:
```
janiversen, line 17: config_overrides is not typed.

pymodbus/logging.py
```diff
@@ -32,12 +32,15 @@ class Log:
...
-            format="%(asctime)s %(levelname)-5s %(module)s:%(lineno)s %(message)s",
-            datefmt="%H:%M:%S",
-        )
+    def apply_logging_config(
+        cls, level: int = logging.WARNING, **config_overrides: dict
+    ) -> None:
+        """Apply basic logging configuration."""
+        default_config: dict = {
+            "format": "%(asctime)s %(levelname)-5s %(module)s:%(lineno)s %(message)s",
+            "datefmt": "%H:%M:%S",
```
janiversen, line 41: please add **config_overrides here.

pymodbus/logging.py
```diff
@@ -32,12 +32,15 @@ class Log:
...
+        }
+        logging.basicConfig(**{**default_config, **config_overrides})
```
janiversen, line 43: This is a highly complex statement and not needed, resolve config_overrides by adding it to default_config first.

janiversen (changes requested):

pymodbus/__init__.py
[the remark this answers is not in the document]
jamesbraza: I ran `./check_ci.sh` locally and it didn't complain. Nor does it complain in GitHub Actions.

What aspect will `pylint` complain about here?

jamesbraza: Ah thanks for pointing out, fixed!

pymodbus/logging.py
jamesbraza:
```python
default_config: dict = {
    **config_overrides,
    "format": "%(asctime)s %(levelname)-5s %(module)s:%(lineno)s %(message)s",
    "datefmt": "%H:%M:%S",
}
```

I think adding it here suggests the overrides are part of the default, so I would prefer not to do that.

pymodbus/logging.py
jamesbraza: Imo a dict merge and then a dict unpack is not complex.  What do you think of this then?

```python
config_kwargs = {**default_config, **config_overrides}
logging.basicConfig(**config_kwargs)
```

jamesbraza:
Thanks for the review! 👍 

> we need to secure that the default_format

The `default_format` lives as a local inside a `classmethod`, so it can't be changed.

> We need a specific format of the log when people open new issues.

Fwiw, it already says this in the docstring of `pymodbus_apply_logging_config`.  Also, the issue template doesn't pass any overrides, so I think when people open an issue they will not pass an override.  I think we're covered here.

Would you like me to add a comment on this to the issue template too?

pymodbus/logging.py
janiversen: it is only a local variable, so feel free to rename it to e.g. effective_config

pymodbus/logging.py
janiversen: Ugly but I can accept that.

janiversen:
I am not talking about changing docstrings. It is possible to change the layout of the log, this should not be permitted for level=DEBUG.

I suppose you want it configurable to match what you use in the application, this is a fair argument for log levels != DEBUG. It correspond to not calling pymodbus_apply_logging_config(). The main reason for level = DEBUG is to report an issue and for that the format must be fixed, in order for you not to change the program it needs to be done in pymodbus_apply_logging_config().

pymodbus/logging.py
jamesbraza: Alright, it's implemented!

jamesbraza:
Okay I hear what you're saying.  The thing is, I actually use debug logging for a whole instrument (`pymodbus` is one of many protocols), so I don't want to deny `logging.DEBUG` level configuration.  In other words, I actually override with the `logging.DEBUG` level.

pymodbus/logging.py
```diff
@@ -32,12 +32,16 @@ class Log:
...
-        )
+    def apply_logging_config(
+        cls, level: int = logging.WARNING, **config_overrides: dict
+    ) -> None:
+        """Apply basic logging configuration."""
+        default_config: dict = {
+            "format": "%(asctime)s %(levelname)-5s %(module)s:%(lineno)s %(message)s",
+            "datefmt": "%H:%M:%S",
+        }
+        config_kwargs = {**default_config, **config_overrides}
```
janiversen, line 43: This still allows to overwrite DEBUG. You are actually a good example, if you overwrite DEBUG, you are not able to submit issues, without somehow changing the overwrite.

In your case it would be a lot simpler, not to call our function, but do
```
  pymodbus.Log.LOG_LEVEL = <what you want>
```
A small change in the Log class could actually avoid that, by moving that line into build_msg.

have LOG_LEVEL = None 
and do:
```
   if not cls.LOG_LEVEL:
       cls.LOG_LOVEL = cls…….
```

IF you do not call our function your basic_config will rule.

janiversen (changes requested):

pymodbus/logging.py
jamesbraza: > ```python
>    if not cls.LOG_LEVEL:
>        cls.LOG_LOVEL = cls…….
> ```

I don't quite understand this part.  What does the `None` sentinel level mean?

Does this go into `build_msg` or `apply_logging_config`, and where within that method?  Why is it mutating the `Log.LOG_LEVEL` if `cls.LOG_LEVEL = None`?

janiversen: if you do not call apply_logging_config, cls.LOG_LEVEL still need to be set.

cls.LOG_LEVEL starts of being None, to secure it is set to effective log level, first time build_msg is called.

jamesbraza: I think `cls.LOG_LEVEL` will fail because `None` comparisons in checks like this `logging.INFO >= cls.LOG_LEVEL` will fail as a `TypeError` before we even get to `build_msg`.

jamesbraza:
Looks like you merged an upstream PR that basically nullifies the core of this PR.

@janiversen I am ready to abandon this, up to you.

pymodbus/logging.py
janiversen: no it will not.

janiversen:
your PR so it’s your decision not mine. I merged a PR with some pending cleanup, but it does not have support for overwriting the configuration.

pymodbus/logging.py
jamesbraza:
```python
>>> logging.INFO >= None
  File "<stdin>", line 1, in <module>
TypeError: '>=' not supported between instances of 'int' and 'NoneType'
```

I don't understand what you mean, I am sorry.

jamesbraza:
Closing since upstream no longer uses `logging.basicConfig`, which this PR centers around

PR #1413, opened 2023-03-07
Add value <-> registers converter helpers.

janiversen:
One of the more frequent problems, "I am trying to read/write a float register", this is of course not 1 register but 2 or 4 registers.

This PR add helpers functions to easily convert:
- int to registers (INT16, INT32, INT64)
- registers to int
- float to registers (FLOAT32, FLOAT64)
- registers to float
- string to registers
- registers to string

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -500,3 +500,99 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
         return self.execute(
             pdu_mei.ReadDeviceInformationRequest(read_code, object_id, **kwargs)
         )
+
+    # ------------------
+    # Converter methods
+    # ------------------
+
+    @classmethod
+    def convert_registers_to_int(cls, registers: list[int]) -> int:
+        """Convert registers to int.
+
+        :param registers: list of registers received from e.g. read_holding_registers()
+        :returns: int16 (1 reg), int32 (2 reg), int64 (4 reg)
+        :raises ModbusException: when size of registers is not 1, 2 or 4
+        """
+        if len(registers) not in {1, 2, 4}:
+            raise ModbusException("Illegal size of register array, use 1, 2 or 4")
+
+        return 0
+
+    @classmethod
+    def convert_int32_to_registers(cls, value: int) -> list[int]:
+        """Convert int to 2 registers (32 bit).
+
+        :param value: integer to be converted:
+        :returns: List of registers, can be used directly in e.g. write_registers()
+        """
+        value_bytes = int.to_bytes(value, 4, "big")
+        return [
+            int.from_bytes(value_bytes[:2], "big"),
+            int.from_bytes(value_bytes[-2:], "big"),
+        ]
+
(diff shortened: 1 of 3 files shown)


pymodbus/client/mixin.py
```diff
@@ -500,3 +502,66 @@ def read_device_information(
...
+        cls, value: Union[int, float, str], data_type: DATATYPE
+    ) -> list[int]:
+        """Convert int/float/str to registers (16/32/64 bit).
+
+        :param value: value to be converted:
+        :param data_type: data type to convert to
+        :returns: List of registers, can be used directly in e.g. write_registers()
+        """
+        if data_type == cls.DATATYPE.STRING:
+            byte_list = value.encode()  # type: ignore[union-attr]
```
janiversen, line 558: @alexrudd2 Here is a (in my mind) a good explanation why mypy is a real PITA. This is perfectly safe code, but mypy do not recognise that.

But admitted mypy have other advantages, so I live with the ignores 🥲

alexrudd2: In this case, I think `if isinstance(value, str):` satisfies mypy by narrowing the type with no ignore.

https://mypy.readthedocs.io/en/latest/type_narrowing.html#type-narrowing-expressions


The docstring for `param: data_type` isn't clear.  Would it be better as 'data type to be encoded as registers'?

Also, it makes no difference but `Union[float, int, str]` can be written as `Union[float, str]`

janiversen: We accept float, int and str that's why I added all 3, a way to tell the user what is allowed.

I will change the doc of param data_type later.

I will have to test if is instance works....I suspect that
```
  isinstance(cls.DATATYPE.STRING) ==  isinstance(cls.DATATYPE.IN16)
```
are true, and thus it will not work.

alexrudd2: > We accept float, int and str that's why I added all 3, a way to tell the user what is allowed.

Yes, the method accepts all 3 types ("broad").  But `encode()` should only work on `str` ("narrow").  I will prepare a PR to demonstrate.

alexrudd2:
I had undeleted the branch to get `dfc36b2`, but no longer need it. :)

janiversen:
you are living a dangerous life :-)

You restore branch "convert_values" and tell me you no longer need it, which I just deleted about a minute ago, but your new PR uses "convert-values"....not much room for errors. Luckily I was awake and not on auto-pilot.

Not a problem, just the highlight of my day.

alexrudd2:
> you are living a dangerous life :-)
> 
> You restore branch "convert_values" and tell me you no longer need it, which I just deleted about a minute ago, but your new PR uses "convert-values"....not much room for errors. Luckily I was awake and not on auto-pilot.
> 
> Not a problem, just the highlight of my day.

Haha, for danger I also use system Python and not a virtual environment. 😱 

 Actually, I was on my way to delete `convert_values` again when I noticed you deleted it 3 minutes before me...

Issue #1033, opened 2022-08-11
Decoding Holding Registers to 16bit binary format 


skar395:
### Versions

* Python: 3.9.13
* OS: Ubuntu 22.04
* Pymodbus: 2.5.3
* Modbus Hardware (if used): 

### Pymodbus Specific
* Server: tcp
* Client: tcp

### Description

In reference to the issue #634 

I am simulating a real case in which in a modbus server I have to read a holding register in binary format.

Reading the register and using the decode_bits() function I expect to receive a list of 16 bits.
decode_bits() only converts 1 byte.

### Code and Logs

```python
# register: 1101000000000000
result = self.connection.read_holding_registers(register, 1, unit=self.slave_unit)
decoder = BinaryPayloadDecoder.fromRegisters(result.registers, Endian.Big, wordorder=Endian.Little)
print(decoder.decode_bits())
# Result: [False, False, False, False, True, False, True, True]
# Expected: [False, False, False, False, False, False, False, False, False, False, False, False, True, False, True, True]

```

janiversen:
A holding register contains bytes (2 pr register) and not bits. Normally you interpret a holding register as an int.

You should be able to use result.registers directly, of course unless you want to do some special conversions.

skar395:
> 

Yes, I know this is not standard, but for what reason does decode_bits() only have to convert one byte?

janiversen:
Because that is how it is programmed 😃 

consider using e.g. decoder.decode_16bit_uint(), or make 2 calls to BinaryPayloadDecoder / decoder.decode_bits.

skar395:
> Because that is how it is programmed smiley
> 
> consider using e.g. decoder.decode_16bit_uint(), or make 2 calls to BinaryPayloadDecoder / decoder.decode_bits.

Mine was not meant to be a provocation!
I expected to receive 2 bytes :)

Thank you

janiversen:
I did not take it as a provocation, hence my smiley.

I suspect it is code that was not touched since the beginning of time....please keep the issue open, and I will see if I can get it fixed, but I cannot promise you a timeline since I am pretty busy with a major rewrite of the whole client structure.

Ps. if/when I fix it, it will be on dev and not on 2.5.3.

dhoomakethu:
AFAIK, `decode_bits` is historically used dealing with single byte registers and does not support multiple bytes.  It should be trivial to patch it support multiple bytes but not a hight priority task IMHO. PR's are welcome

Issue #634, opened 2021-05-18
Decoding Holding Registers to binary format


fess1100:
### Versions

* Python: 3.9.1
* OS: Windows 10 Pro 2004
* Pymodbus: 2.5.1
* Modbus Hardware (if used): 

### Pymodbus Specific
* Client: rtu

### Description
Need help with decoding Holding Registers to binary format. When reading, I get a value from 8 to 15 bits. I can't read the value from 0 to 7 bits. Thanks

Modbus Slave: ID:1 F:03
0: -47,2 (float32)
1:
2: -7565 (int)
3: -3 (int)
4: 4581 (uint)
5: 5555 (uint)
6: 67 (uint)
7: 76 (uint)
8: 0000 0001 0101 1001 (bool)
9: 0000 0100 0110 0011 (bool)

### Code and Logs

```python
# code
from pymodbus.client.sync import ModbusSerialClient
from pymodbus.payload import BinaryPayloadDecoder
from pymodbus.constants import Endian

client = ModbusSerialClient(method='rtu', port='COM5', baudrate=19200, timeout=3, parity='N', stopbits=1, bytesize=8)

if client.connect():
    res1 = client.read_holding_registers(0, 10, unit=1)
    res2 = client.read_holding_registers(0, 2, unit=1)
    res3 = client.read_holding_registers(2, 1, unit=1)
    res4 = client.read_holding_registers(3, 1, unit=1)
    res5 = client.read_holding_registers(8, 1, unit=1)
    if not res3.isError():
        print(res1)
        print(res1.registers)
        print("-" * 60)
        print(BinaryPayloadDecoder.fromRegisters(res2.registers, Endian.Big, Endian.Big).decode_32bit_float())
        print(BinaryPayloadDecoder.fromRegisters(res3.registers, Endian.Big, Endian.Little).decode_16bit_int())
        print(BinaryPayloadDecoder.fromRegisters(res4.registers, Endian.Big, Endian.Little).decode_16bit_int())
        print(BinaryPayloadDecoder.fromRegisters(res5.registers, Endian.Big, Endian.Little).decode_bits())
    else:
        print(res3, 'Error')
        # print('\n')
    print("-" * 60)
else:
    print('Cannot connect Modbus Server/Slave')


# logs
ReadHoldingRegistersResponse (10)
[49724, 52429, 57971, 65533, 4581, 5555, 67, 76, 345, 1123]
------------------------------------------------------------
-47.20000076293945
-7565
-3
[...]

PR #1420, opened 2023-03-09
Use type narrowing for mypy

alexrudd2:
See #1413.

This PR uses [type narrowing](https://mypy.readthedocs.io/en/latest/type_narrowing.html#type-narrowing-expressions) to satisfy `mypy` rather than `# type: ignore`.


pymodbus/client/mixin.py
```diff
@@ -508,7 +508,7 @@ def read_device_information(
     # ------------------
 
     class DATATYPE(Enum):
-        """Datatype enum for convert_* calls."""
+        """Datatype name and number of bytes, used for convert_* calls."""
```
janiversen, line 511: The enum is the name, the first part of the tuple is format code for struct.
```suggestion
        """Datatype enum used in convert_* calls."""
```
I think it is important to state that its an enum for the user, the content is not important.

pymodbus/client/mixin.py
```diff
@@ -523,7 +523,7 @@ class DATATYPE(Enum):
     @classmethod
     def convert_from_registers(
         cls, registers: List[int], data_type: DATATYPE
-    ) -> Union[int, float, str]:
+    ) -> Union[float, str]:
```
janiversen, line 526: But it also returns an int ! if you do not specify that, the user might believe he needs to write 
```
   int(result)
```
to get rid of the decimals.

pymodbus/client/mixin.py
```diff
@@ -546,16 +546,16 @@ def convert_from_registers(
...
         """Convert int/float/str to registers (16/32/64 bit).
 
         :param value: value to be converted:
-        :param data_type: data type to convert to
+        :param data_type: data type to be encoded as registers
         :returns: List of registers, can be used directly in e.g. write_registers()
         """
-        if data_type == cls.DATATYPE.STRING:
-            byte_list = value.encode()  # type: ignore[union-attr]
+        if isinstance(value, str):
```
janiversen, line 557: that is not the intention, and are wrong is the call is using value = "127" and DATATYPE.INT.

I want to check the datatype not the value.

pymodbus/client/mixin.py
```diff
@@ -546,16 +546,16 @@ def convert_from_registers(
 
     @classmethod
     def convert_to_registers(
-        cls, value: Union[int, float, str], data_type: DATATYPE
+        cls, value: Union[float, str], data_type: DATATYPE
```
janiversen, line 549: it is important for the user to know he can call with an "int" and do not need to convert to float.

e.g.
```
  value = 127.0, data_type = DATASTYPE.INT16
```
is not legal but:
```
  value = 127, data_type = DATASTYPE.INT16
  value = 127.2, data_type = DATASTYPE.FLOAT32
```
are both legal.

janiversen (changes requested):

pymodbus/client/mixin.py
janiversen: Please remember these documentations are mainly for external users, and they use DATATYPE as an enum. However I do not feel strongly about so if you also want to document the internals then please...

janiversen:
An alternative, which I think is more complicated to understand, would be to change the enum to something like

.BIT16
.BIT32
.BIT64
.STRING

wdyt?

pymodbus/client/mixin.py
alexrudd2: Hmm, let me think about this.  To me the important part for users to understand is that datatypes aren't always 1 byte when they become registers.  For instance:
https://github.com/pymodbus-dev/pymodbus/issues/634#issuecomment-845216074
https://github.com/pymodbus-dev/pymodbus/issues/1033#issuecomment-1212168222
https://github.com/pymodbus-dev/pymodbus/discussions/1303#discussioncomment-4794593

janiversen: actually most users do not think in bits or bytes....they think 1 register == 1 float/int/string.

I have that problem constantly in home assistant and quite frequently here. 

I am open to ideas.

pymodbus/client/mixin.py
alexrudd2: OK, `convert_to_registers("127", DATATYPE.INT)` is strange to me, but I have tried a different approach that permits this and satisfies `mypy`.  

Actually, the new code prevents a crash if you were to call `convert_to_registers(127, DATATYPE.STRING)`, I think.  Which is the purpose of tools like mypy :)

alexrudd2:
> An alternative, which I think is more complicated to understand, would be to change the enum to something like
> 
> .BIT16 .BIT32 .BIT64 .STRING
> 
> wdyt?

Interesting, but I don't think it's any better.  My only other idea is to subclass the built-in types:

```python
class INT16(int):
   bytes = 1
   code = "n"   # for f-string example, actually it should be 'h'

class UINT16(int):
   bytes = 1
   code = "H"

class FLOAT32(float):
  bytes: 2
  code = "f"

>>> cls = INT16()
>>> value = 123
>>> print(f'{value:{cls.code}}')
123
>>> isinstance(cls, int)
True
```

I'm using an f-string instead of the proper `encode()` just for simplicity.

janiversen:
but that does not change anything seen from an applications perspective. An application use the enum and do not care about the content.

Let me take a look at your new changes.

pymodbus/client/mixin.py
```diff
@@ -551,11 +551,11 @@ def convert_to_registers(
         """Convert int/float/str to registers (16/32/64 bit).
 
         :param value: value to be converted:
-        :param data_type: data type to convert to
+        :param data_type: data type to be encoded as registers
         :returns: List of registers, can be used directly in e.g. write_registers()
         """
-        if data_type == cls.DATATYPE.STRING:
-            byte_list = value.encode()  # type: ignore[union-attr]
+        if data_type == cls.DATATYPE.STRING and isinstance(value, str):
```
janiversen, line 557: I like the idea, but how about moving it inline and do a raise (as I do with the register count).
Something like:
```
if data_type == cls.DATATYPE.STRING:
    if not isinstance(value, str):
        raise RuntimeError(“wrong combination of datatype and value”)
```

Sorry I am on my ipad, so no nice formatting.

janiversen (changes requested):

pymodbus/client/mixin.py
janiversen: your suggestion works, but the side effects can be dangerous, since it goes into the int/float part.

alexrudd2:
> Interesting, but I don't think it's any better. My only other idea is to subclass the built-in types:

I experimented with this idea and it had some advantages, but ended up not being good.

pymodbus/client/mixin.py
```diff
@@ -551,14 +551,21 @@ def convert_to_registers(
...
 
         :param value: value to be converted:
-        :param data_type: data type to convert to
+        :param data_type: data type to be encoded as registers
         :returns: List of registers, can be used directly in e.g. write_registers()
+        :raises TypeError: when there is a mismatch between data_type and value
         """
-        if data_type == cls.DATATYPE.STRING:
-            byte_list = value.encode()  # type: ignore[union-attr]
+        if "STRING" in data_type.name:
```
janiversen, line 558: I think it is a bad idea to have an extra string "STRING", why do you see that as a better option ?

pymodbus/client/mixin.py
```diff
@@ -551,14 +551,21 @@ def convert_to_registers(
...
+            if not isinstance(value, str):
+                raise TypeError(f"Value should be string but is {type(value)}.")
+            byte_list = value.encode()
             if len(byte_list) % 2:
                 byte_list += b"\x00"
         else:
+            if "INT" in data_type.name and not isinstance(value, int):
```
janiversen, line 565: I have nothing against testing, but in that case you missed out on float. And I question if this test is worth while, because pack/unpack catches this.

janiversen (changes requested):

janiversen:
> I experimented with this idea and it had some advantages, but ended up not being good.

That might be, but I really like your new texts, they are a lot better than mine.

pymodbus/client/mixin.py
alexrudd2: I did it to be similar to `if "INT" in data_type.name:` below.

janiversen: How about dropping the "INT" check, and change "STRING" to cls.DATATYPE.STRING.name, then I am ready to merge.

alexrudd2: Hmm, OK but I am on mobile and can't do this today

janiversen: I can do it, if you are ok with that.

janiversen: I am converting unittest to pytest, which is 10 tons of silly editing...so something to break the rhythm is welcome.

PR #1423, opened 2023-03-09
Don't publish univeral (Python2 / Python 3) wheels

alexrudd2:
Releases on PyPI currently have a filename such as:
`pymodbus-3.2.0-py2.py3-none-any.whl`

Per PEP425, https://peps.python.org/pep-0425/#python-tag, the 
compatability tags `py2` and `py3` indicate the package is compatible with 
Python2 and Python3.

This PR drops Python2, which was accidentally left in.


setup.cfg
```diff
@@ -542,10 +542,6 @@ all_files  = 1
 upload_dir = build/sphinx/html
 
 
-[bdist_wheel]
```
janiversen, line 545: Do not remove the configuration entry point please.

janiversen (changes requested):

janiversen (approved):
LGTM. Thanks (I will just rebase dev, to get a clean CI)

Issue #1354, opened 2023-02-18
Feature request: add hook to detect reconnect


wlcrs:
Both the RTU and TCP client have functionality that automatically do a reconnect when the connection is interrupted.

I would like to be notified when such an event takes place, so that I can take some appropriate actions.

There is no clear way to add this functionality in the client currently, and with your upcoming work in #1199 it is unwise for me to try to implement something myself.

Proposal: add an argument to the constructor which allows you to pass a function that will be called when the event takes place. Maybe it even makes sense to add callbacks for different events?

```
callbacks : {
 'on_connection_made': ...,
 'on_connection_lost': ....,
 'on_reconnect': ....
}
```
On `AsyncModbusXXXClient` objects these callback functions are async, otherwise they are expected to by normal functions?

janiversen:
Your idea of adding callbacks is good ! Please remember this needs to be added for async as well as sync and for all types.

I am not sure why you think there are no clear way, I would add the arguments to the base client (making it available for all) and then insert the hook in the appropriate methods.

The first part of cleaning the client is to clean the base, but I do not think it will affect a PR from you a lot, should it affect I can help solve the problems.

The callbacks should be defined as callbacks not async (this is the same as as “connection_made” etc). It is a lot more work to make them async, since you need to start an async task.

So please do not let #1199 stand in the way for a nice new functionality.

Feel free to make the pull request early in your process, and request comments from the maintainers.

janiversen:
On dev.

PR #1425, opened 2023-03-10
Notify on reconnect

wlcrs:
As requested in https://github.com/pymodbus-dev/pymodbus/issues/1354#issuecomment-1435654383: a PR adding support a reconnect callback function .

I also modified a test so that it checks if the function is called as expected.

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -31,6 +31,7 @@ class ModbusBaseClient(ModbusClientMixin):
     :param broadcast_enable: (optional) True to treat id 0 as broadcast address.
     :param reconnect_delay: (optional) Minimum delay in milliseconds before reconnecting.
     :param reconnect_delay_max: (optional) Maximum delay in milliseconds before reconnecting.
+    :param on_reconnect_callback: (optional) Function that will be called just before a reconnection attempt.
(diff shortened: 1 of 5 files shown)


pymodbus/client/base.py
```diff
@@ -118,6 +120,7 @@ def __init__(
         self.params.broadcast_enable = bool(broadcast_enable)
         self.params.reconnect_delay = int(reconnect_delay)
         self.params.reconnect_delay_max = int(reconnect_delay_max)
+        self.params.on_reconnect_callback = on_reconnect_callback
```
janiversen, line 123: I miss :parma documentation of the parameter (needed in serial/tcp/udp/tls .py)

janiversen (changes requested):

pymodbus/client/base.py
wlcrs: For consistency, I added the `:param` documentation string in a similar fashion as for the related parameters `reconnect_delay` and `reconnect_delay_max`: these are only documented in [`base.py`](https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/client/base.py#L32-L33) and not in serial/tcp/udp/tls.py .

janiversen (approved):
LGTM. Thanks

PR #1427, opened 2023-03-10
Change unit= to slave= for framers.

janiversen:
There are still a fairly large number of APIs that have unit= instead of slave=.

This PR cleans the API as well as most internal parts.

It turns out that the changes dig deep into the core, so it will be done in parts.


janiversen:
that depends on the call you use, please read the documentation.

in the example it is slave=

From version 3.3 (see dev.) unit= is gone.

janiversen:
we recommend using slave= but depending on your device it might work with the default slave.

hasenradball:
For the call of the function it is not necessary to use `slave=` or `unit=` if the order of the parameters fiit to the function definition, so i deleted the parameter name when calling the functin.
So it works for all versions
😉

janiversen:
that is correct, but you then assume we do not change the position, something that will happen.

hasenradball:
Something that will happen or could?

PR #1431, opened 2023-03-13
Remove a colon from param description

krivulcik:
The extra colon caused the param parsing to fail in the REPL.


janiversen:
This pull request do not contain any changes. Apart from that the : you refer to is within a comment, so it is quite hard to understand how that affects REPL.

Furthermore (after checking the sources) REPL do not use convert_to_registers or convert_from_registers, either you have patched your sources or you use a version we do not know.

janiversen:
Please check with current dev, I am very sure you a different problem.

janiversen:
I had another look at your error, sorry did not check --help, but it is corrected.

Thanks for bringing it to our attention.

PR #1452, opened 2023-03-24
Add 'after_connect_delay' parameter

wlcrs:
Some devices are not ready to immediately answer requests after a connection has been established. This PR adds the ability to specify a wait period just after a client established a connection to a server. 

While a user of this library might just do a wait itself upon the initial connection, this is not possible in the case a reconnect happens. This PR solves that issue.


janiversen:
I am not sure what you want to solve here, even with reconnect it is the application that decides when to make calls. Reconnect do not repeat any calls.

test/test_client.py
```diff
@@ -274,6 +274,21 @@ async def test_client_instanciate(
...
+def test_delay_after_connect():
+    """Test protocol lost connection."""
+    with mock.patch("time.sleep", return_value=None) as mock_sleep:
+        client = lib_client.AsyncModbusTcpClient("127.0.0.1", after_connect_delay=10)
+        client.connection_made(client)
+
+        mock_sleep.assert_called_once_with(0.01)
+
+    with mock.patch("time.sleep", return_value=None) as mock_sleep:
+        client = lib_client.AsyncModbusTcpClient("127.0.0.1")
```
janiversen, line 286: you are not testing what happens if the appl. make a call in between.

pymodbus/client/base.py
```diff
@@ -242,6 +247,10 @@ def connection_made(self, transport):
         self.transport = transport
         Log.debug("Client connected to modbus server")
         self._connected = True
+
+        if self.params.after_connect_delay:
+            time.sleep(self.params.after_connect_delay / 1000)
```
janiversen, line 252: This is an async corutine, you cannot use time.sleep here, that blocks the whole loop.

wlcrs:
Yes, but the application does not know when a reconnect has happened, and it thus cannot wait the appropriate time before a new call is made.

An alternative solution is to add a callback instead in the connection_made function for an application to add some custom logic.

The on reconnect callback cannot be used here as we do not know how long it will take for the connection to be initialised.

janiversen (changes requested):
I am not convinced this is a needed change, but the suggested solution blocks all async work, like e.g. a second client.

janiversen:
Be aware that the bug you refer to, can very well be because the your lib requires async_serial which will conflict with the version pymodbus uses (integrated in the source due to a severe bug, have f8led a pull request, but the project seem abandoned).

pymodbus/client/base.py
wlcrs: `connection_made` is sadly not an async function, even while it is part of the `asyncio.Protocol` interface.

wlcrs:
> Be aware that the bug you refer to, can very well be because the your lib requires async_serial which will conflict with the version pymodbus uses (integrated in the source due to a severe bug, have f8led a pull request, but the project seem abandoned).

I can indeed remove that direct dependency, but:
- the copy of async_serial that you use is in another namespace, so the faulty async_serial library is not even used as you don't import it anymore?
- the problem is present for TCP connections, which do not use serial_asyncio

pymodbus/client/base.py
janiversen: I did not write it is an async method, because it is not, but it is a coroutine (called from create_protocol, which are async), and thus calling time.sleep blocks the whole asyncio loop.

Coroutines are meant to do something fast, but not make big work, IO or other items that can block the rest of the system.

janiversen:
It is your code, so you do what you want, I was just pointing out the potential problem, especially since I had it. And just for the record python do not have proper namespaces like we are used from C++/C# it has one global space.

janiversen:
Btw. you got this added: https://github.com/pymodbus-dev/pymodbus/commit/dec73002eafa954a338b0144ef39a1801c1ea517
so the app can be notified about reconnects if needed and then do whatever they need to do.

janiversen:
Any progress on this PR ?

wlcrs:
After your remarks, I did some refinement of this PR and implemented it in my library first by creating subclasses of `AsyncModbusTcpClient` and `AsyncModbusRtuClient` instead, cfr: 
https://github.com/wlcrs/huawei-solar-lib/blob/119a3e9b5b6f5456ddba9eeeedd4de9832b030e6/src/huawei_solar/modbus.py#L21-L40

I went with an `asyncio.Event` that is set/unset in `client_connection_made`/`client_connection_lost`, and that is checked against when doing requests:

https://github.com/wlcrs/huawei-solar-lib/blob/119a3e9b5b6f5456ddba9eeeedd4de9832b030e6/src/huawei_solar/huawei_solar.py#L133-L135

Does that seem like an appropriate solution to you that we can/should upstream into this library?

janiversen:
It might be an appropriate solution, but please explain why you cannot use the callback you got implemented (I assumed to solve cases like this).

wlcrs:
The callback that I previously added is called before the reconnect starts, not after it succeeded. 

Let's say that due to a network interruption a reconnect is triggered, the current callback will be called at the start of that outage, while I need a wait after the (re)connect succeeds. In this case that wait can only start after the reconnect succeeded. These 2 events can be minutes apart in case of a long outage

janiversen:
thats a good point. I cannot however not review your solution as it is difficult to see what the real changes are, and what is application stuff.

If you want a review (and potentially merged), then please either update this Pr, or close it and make a new one (please do not forget to update your dev branch, to ensure you work on the newest code).

wlcrs:
I'll close this PR for now. Thanks!

PR #1462, opened 2023-03-27
Use alpha version on dev to remove setuptools pin

alexrudd2:
See https://github.com/pymodbus-dev/pymodbus/issues/1372#issuecomment-1483774133

> > > > Can this setuptools<66.0 requirement be limited to development versions?
> > >
> > > at the moment no. But pull requests are welcome.
> > 
> > Actually as of [025ee8a](https://github.com/pymodbus-dev/pymodbus/commit/025ee8a4b0450e1d99e7d50d08c0eb4d08fb843f) the pin can be removed.
> > 
> > However, a commit like [6b27cd1](https://github.com/pymodbus-dev/pymodbus/commit/6b27cd1e03dfe2391c8216a850250b3337b9c285) will break things again, since it uses `version = Version("pymodbus", 3, 2, "x", "")`.
> > 
> > If instead `dev` switches to something like this: `version = Version("pymodbus", 3, 2, 3, "alpha")` the problem is permanently avoided.
>
> pull requests are welcome


requirements.txt
```diff
@@ -12,7 +12,6 @@
 # Required packages.
 # -------------------------------------------------------------------
 # install:required
-setuptools<66.0.0
```
janiversen, line 15: This does not look right (but I might be wrong). Why don't we need setup tools as a requirement?

I had expected that only the version limit to be removed.

pymodbus/version.py
```diff
@@ -37,4 +37,4 @@ def __str__(self) -> str:
         return f"[{self.package}, version {self.short()}]"
 
 
-version = Version("pymodbus", 3, 3, 0, "")
+version = Version("pymodbus", 3, 3, 0, "alpha")
```
janiversen, line 40: Just for information: can the text be any text e.g. "dev" ?

janiversen (changes requested):

requirements.txt
alexrudd2: https://github.com/pymodbus-dev/pymodbus/pull/1373

See the original pin here.  `setuptools` was never required, presumably since it's almost always installed with Python itself.

https://stackoverflow.com/questions/64811350/is-setuptools-always-installed-in-python-by-default

pymodbus/version.py
alexrudd2: It can't be arbitrary text, but it could be `dev` I think.  Do you want me to try?

https://peps.python.org/pep-0440/#implicit-development-release-number.

requirements.txt
dhoomakethu: `setuptools` I'd an internal requirement to make sure a target machine has required version satisfied even though the Python installation comes with it. Not everyone updates the default packages. It's better to have it pinned.

janiversen (approved):
but if it works with whatever is installed as default, then there are no need to make an extra demand

Issue #1455, opened 2023-03-25
pymodbus.simulator ignores command line arguments "http_port" and "json_file"


chkr-private:
### Versions

- Python: 3.10.10
- OS: Fedora 36
- Pymodbus: 3.2.2

### Pymodbus Specific

- Server: tcp
- Client: tcp

### Description

- when starting pymodbus.simulator with a specific json file, an error stating a different filename is shown
```
$ pymodbus.simulator --json_file /tmp/setup.json 
2023-03-25 20:06:03,532 INFO  logging:96 Start simulator
Traceback (most recent call last):
  File "/usr/bin/pymodbus.simulator", line 8, in <module>
    sys.exit(main())
  File "/usr/lib/python3.10/site-packages/pymodbus/server/simulator/main.py", line 119, in main
    asyncio.run(run_main(), debug=True)
  File "/usr/lib64/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib64/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/usr/lib/python3.10/site-packages/pymodbus/server/simulator/main.py", line 113, in run_main
    task = ModbusSimulatorServer(**cmd_args)
  File "/usr/lib/python3.10/site-packages/pymodbus/server/simulator/http_server.py", line 130, in __init__
    with open(json_file, encoding="utf-8") as file:
FileNotFoundError: [Errno 2] No such file or directory: './pymodbus/server/simulator/setup.json'
```
- the problem is in https://github.com/pymodbus-dev/pymodbus/blob/c2db53cad3c60834bffecd23db699f6982f914fb/pymodbus/server/simulator/main.py#L113 where the command line arguments are overwritten
- if it is the intention having default values, a patch keeping exactly these defaults would look like this:
```
diff --git a/pymodbus/server/simulator/main.py b/pymodbus/server/simulator/main.py
index 09af492..443e1a6 100755
--- a/pymodbus/server/simulator/main.py
+++ b/pymodbus/server/simulator/main.py
@@ -70,6 +70,7 @@ def get_commandline():
     parser.add_argument(
         "--http_port",
         help="use <http_port> as port to bind http listen",
+        default=8081,
         type=str,
     )
     parser.add_argument(
@@ -82,6 +83,7 @@ def get_commandline():
     parser.add_argument(
         "--json_file",
         help='name of json file, default is "setup.json"',
[...]

PR #1463, opened 2023-03-28
pymodbus.simulator fixes

chkr-private:
This PR is related to issue #1455: starting pymodbus.simulator.

- correct misplaced default value: 8081 applies to the http_port and not the http_host
- makes the default path for setup.json independent of the current working directory


janiversen (approved):
LGTM. thanks

PR #1470, opened 2023-04-03
Test case for UDP issue

wirehell:


janiversen:
Reopening to check if it can be converted to a general test case.

PR #1473, opened 2023-04-07
Remove old aliases to OSError

alexrudd2:
> _exception_ socket.**error**
>
> A deprecated alias of [`OSError`](https://docs.python.org/3/library/exceptions.html#OSError).
> 
> Changed in version 3.3: Following [PEP 3151](https://peps.python.org/pep-3151/), this class was made an alias of [`OSError`](https://docs.python.org/3/library/exceptions.html#OSError).

https://docs.python.org/3/library/socket.html#socket.error


janiversen (approved):
Nice one.

PR #1476, opened 2023-04-07
Replace flake8 with ruff

alexrudd2:
`ruff` is rapidly absorbing lots of linters.  I used `flake8-to-ruff` to generate a config which basically matches the current `flake8` configuration.

Integration of bandit is left for later, since it uses the rule names from [`flake8-bandit`](https://beta.ruff.rs/docs/rules/#flake8-bandit-s) which are slightly different.

Note: HomeAssistant has already switched, although [their rules](https://github.com/home-assistant/core/blob/7eccef87c2cccb834f5ff394f765e697831390e3/pyproject.toml#L240) are more strict so I did not copy them (yet).

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -5,7 +5,6 @@ exclude = [
     ".venv",
     ".git",
     "build",
-    "examples/v2.5.3",
 ]
 ignore = [
     "D202", # No blank lines allowed after function docstring (to work with black)
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,0 +1,24 @@
+target-version="py38"
+exclude = [
+    "pymodbus/client/serial_asyncio",
+    "venv",
+    ".venv",
+    ".git",
+    "build",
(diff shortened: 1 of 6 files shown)


janiversen:
A couple of comments.
If it makes things simpler then please make a PR that changes from bandit to flake8-bandit.

Can you please ( if you have not already) the rules from HA, commented out, that way it is easier to activate when you find t8mr.

Now to the review

ruff.toml
```diff
@@ -0,0 +1,24 @@
+target-version="py38"
+exclude = [
+    "pymodbus/client/serial_asyncio",
+    "venv",
+    ".venv",
+    ".git",
+    "build",
+    "examples/v2.5.3",
```
janiversen, line 8: No longer in our code.

janiversen (changes requested):
I like this one.

I just had one small comment.

PR #1480, opened 2023-04-10
Implement pyflakes-simplify

alexrudd2:
Implement pyflakes-simplify rules used by HomeAssistant

(I tried enabling all rules and did not like it combining `if` statements, so the HA rules make sense.)


janiversen (approved):
LGTM. Thanks

PR #1481, opened 2023-04-10
Enable ruff-specific rules

alexrudd2:
Enable ruff-specific `"RUF"` [rules](https://textual.textualize.io/blog/2023/02/11/the-heisenbug-lurking-in-your-async-code/). 

In the case of `RUF006`, this is only relevant when garbage collection [occurs](https://textual.textualize.io/blog/2023/02/11/the-heisenbug-lurking-in-your-async-code/).  The two places where this rule triggered are not-lived enough to worry about GC (IMO), so I silenced them instead.

Note!  This is branched off #1480

### DIFF
--- a/examples/contrib/redis_datastore.py
+++ b/examples/contrib/redis_datastore.py
@@ -1,11 +1,10 @@
 """Datastore using redis."""
 # pylint: disable=missing-type-doc
+from contextlib import suppress
 
 
-try:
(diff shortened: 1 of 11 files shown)


janiversen (approved):
LGTM. Thanks

I do not like silencing, but I agree with you here.

PR #1488, opened 2023-04-11
Move bandit to pre-commit

alexrudd2:
- Remove [obsolete](https://github.com/PyCQA/bandit/issues/675) B325 rule.
- Move bandit to pre-commit, so it's included in `check_ci.sh` and doesn't need its own CI runner.


janiversen (approved):
LGTM, and you decide when it is time to move it to ruff.

PR #1489, opened 2023-04-11
Refactor REPL client to reduce complexity

alexrudd2:
This splits up a large function (complexity 21) into smaller ones within a class.  The tradeoff is more LOC.  What do you think?

### DIFF
--- a/pymodbus/repl/client/main.py
+++ b/pymodbus/repl/client/main.py
@@ -151,86 +151,103 @@ def _process_args(args: list, string: bool = True):
     return kwargs, execute
 
 
-def cli(client):  # pylint: disable=too-complex
-    """Run client definition."""
-    use_keys = KeyBindings()
-    history_file = pathlib.Path.home().joinpath(".pymodhis")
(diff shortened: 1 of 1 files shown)


dhoomakethu:
This looks good to me on a first pass, I will spend some more time to see if everything is fine.

My original idea was not to bloat the core  libraries with all these helpers/scripts but add them as separate projects which consumes pymodbus. It will be easier to maintain in that way.

janiversen (approved):
LGTM, but I will wait for @dhoomakethu 

@dhoomakethu I get your point, but in that case we should move the whole REPL out as a separate repo, and I am not sure that it is worth it.

janiversen:
If @dhoomakethu have concerns later then we will address them.

dhoomakethu:
@janiversen yes, I had plans to move repl as an add-on package for pymodbus. I am not sure if it's worth the effort though.

janiversen:
I think it is not worth the effort, it does in no way disturb the library.

PR #1492, opened 2023-04-13
New common transport layer.

janiversen:
Internally both client and server uses async transport, in order to optimise code maintenance.

The transport layer is responsible for handing:
- connect/disconnect
- receive data, with frame handling
- send data, with automatic retry (if configured)
- automatic reconnect (if configured)
- Diagnostic (if configured)

The transport layer is solely internal to pymodbus and do not expose any API to the outside world. States like e.g. connected, are NOT exposed directly. but must be added in the different interfaces.


pymodbus/transport/base.py
```diff
@@ -0,0 +1,109 @@
...
+    # -------------------------- #
+    def connection_made(self, transport: Any):
+        """Call from asyncio, when a connection is made.
+
+        :param transport: socket etc. representing the connection.
+        """
+        self.transport = transport
+        Log.debug("Connected {}", self.comm_name)
+        if self.on_connection_made:
+            self.on_connection_made(self.comm_name)  # pylint: disable=not-callable
```
janiversen, line 63: @alexrudd2 please do me a favor and have a look at this puzzle.

I declare on_connection_made in line 45, and pylint complains here in line 63 (same goes for on_connection_lost). I simply cannot make it right for both pylint and mypy, while keeping it working.

The idea is that the inherited classes can set a callback and get called, if the callback is None no action is taken.

I am sure it is something simple, but I have run out of ideas.

Thanks in advance for your help.

alexrudd2: Well, I don't know if it's a _good_ solution, but it works:

`self.on_connection_made: Callable[[str], None] = lambda foo : None`

Then delete the conditional and always call the callback.  😆

janiversen: Thanks, at least a solution.

alexrudd2:
This introduced a tiny type error (https://github.com/pymodbus-dev/pymodbus/issues/1511), which was easy enough to solve.

janiversen:
Thanks, and sorry for that.

PR #1496, opened 2023-04-16
pymodbus.server: listen on ID 1 by default

chkr-private:
- default was lost in commit 670123338f8f4ba41035c790609fe346b52ac4e8 during the migration to typer


janiversen (approved):
LGTM.

PR #1499, opened 2023-04-18
Refactor REPL server to reduce complexity

alexrudd2:
See https://github.com/pymodbus-dev/pymodbus/pull/1489

This splits up another large function (complexity 23) into smaller ones.

This was much easier/cleaner than the client.  HOWEVER, there are no tests.  I ran the examples in the readme- both servers at least _start_, but I didn't test in detail.


alexrudd2:
Hahahaha, pylint and ruff calculate complexity slightly differently <_<

PR #1513, opened 2023-04-24
Switch to cache included in setup-python action

alexrudd2:
Currently the GH actions CI uses `syphar/restore-virtualenv@v1` to restore cache/restore pip dependencies.

It should be possible to do this with `actions/setup-python@v4` directly, since it has [built-in caching](syphar/restore-virtualenv@v1).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
           cache: 'pip'
           cache-dependency-path: requirements.txt
 
-      - name: venv create
+      - name: create venv
         if: steps.setup-python.outputs.cache-hit != 'true'
         run: pip install -e . -r requirements.txt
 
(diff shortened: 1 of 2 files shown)


janiversen (commented):
I like your idea ! but please do not mix code changes and CI changes. I really prefer a PR with only the CI change.

(could this be due to the PR I just merged, in which case it is just a rebase)

alexrudd2:
> I like your idea ! but please do not mix code changes and CI changes. I really prefer a PR with only the CI change.
> 
> (could this be due to the PR I just merged, in which case it is just a rebase)

Yes, I had to start with the PR you just merged or the CI fails :).  I am going to rebase again, which will drop the code change.

alexrudd2:
Actually, this PR makes things slower because it only installs the downloads, not the install.  https://github.com/actions/setup-python/issues/330.

Therefore, closing.

Issue #1500, opened 2023-04-18
Evaluate  flake8-pytest-style as a linter


alexrudd2:
> @janiversen I would recommend checking out [`flake8-pytest-style`](https://github.com/m-burst/flake8-pytest-style), you can hook it right into `pre-commit`.
> 
> It really helps one best adopt `pytest`'s DSL

@jamesbraza 
If you are still interested, it's fast and easy to enable in ruff.
https://beta.ruff.rs/docs/rules/#flake8-pytest-style-pt

Personally on my other projects it only found `PT001` and `PT018`.  I found those to be useless and kept it disabled.

```
examples/client_calls.py:213:5: PT018 [*] Assertion should be broken down into multiple parts
examples/client_calls.py:218:5: PT018 [*] Assertion should be broken down into multiple parts
test/test_client.py:41:5: PT006 [*] Wrong name(s) type in `@pytest.mark.parametrize`, expected `tuple`
test/test_client.py:209:5: PT006 [*] Wrong name(s) type in `@pytest.mark.parametrize`, expected `tuple`
test/test_client.py:364:2: PT023 [*] Use `@pytest.mark.skip()` over `@pytest.mark.skip`
test/test_client.py:543:5: PT006 [*] Wrong name(s) type in `@pytest.mark.parametrize`, expected `tuple`
test/test_client_multidrop.py:18:5: PT005 Fixture `_framer` returns a value, remove leading underscore
test/test_client_multidrop.py:23:5: PT005 Fixture `_callback` returns a value, remove leading underscore
test/test_example_client_server.py:77:26: PT006 [*] Wrong name(s) type in `@pytest.mark.parametrize`, expected `tuple`
test/test_example_client_server.py:105:26: PT006 [*] Wrong name(s) type in `@pytest.mark.parametrize`, expected `tuple`
test/test_exceptions.py:30:17: PT017 Found assertion on exception `exc` in `except` block, use `pytest.raises()` instead
test/test_factory.py:99:5: PT004 Fixture `setup` does not return anything, add leading underscore
test/test_framers.py:19:2: PT001 [*] Use `@pytest.fixture()` over `@pytest.fixture`
test/test_framers.py:25:2: PT001 [*] Use `@pytest.fixture()` over `@pytest.fixture`
test/test_logging.py:31:9: PT006 [*] Wrong name(s) type in `@pytest.mark.parametrize`, expected `tuple`
[...]

PR #1515, opened 2023-04-24
Defaulted pylint

jamesbraza:
Before working on https://github.com/pymodbus-dev/pymodbus/issues/1500, I noticed the `pylint` config was way too verbose (repeated many defaults).  This makes it:
- More arduous to update to newer `pylint` and get updated defaults/args
- Hard to determine what non-default behaviors the config is actually calling out

This PR:
- Removed defaulted options
- Removed some over-constraining options that prevented `pylint` from discerning useless suppressions

To see `pylint` defaults: https://github.com/pylint-dev/pylint/blob/v2.15.10/examples/pylintrc

### DIFF
--- a/pymodbus/client/tcp.py
+++ b/pymodbus/client/tcp.py
@@ -259,9 +259,7 @@ class ModbusTcpClient(ModbusBaseClient):
 
         return b"".join(data)
 
-    def _handle_abrupt_socket_close(
-        self, size, data, duration
-    ):  # pylint: disable=missing-type-doc
+    def _handle_abrupt_socket_close(self, size, data, duration):
         """Handle unexpected socket close by remote end.
 
         Intended to be invoked after determining that the remote end
--- a/pymodbus/constants.py
+++ b/pymodbus/constants.py
@@ -191,7 +191,7 @@ class ModbusStatus:  # pylint: disable=too-few-public-methods
 
     Waiting = 0xFFFF
     Ready = 0x0000
-    On = 0xFF00  # pylint: disable=invalid-name
+    On = 0xFF00
     Off = 0x0000
     SlaveOn = 0xFF
     SlaveOff = 0x00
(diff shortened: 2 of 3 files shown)


jamesbraza:
Cc @alexrudd2 @janiversen for review.  not sure why Lock Threads action fails

alexrudd2 (approved):
LGTM; I'm in favor of deleting stuff!

that said, I don't use pylint myself so I'll defer to @janiversen

alexrudd2:
> not sure why Lock Threads action fails

I'm fairly sure it's because you're opening this from a fork, which won't have the read/write permissions to this repo.

The [Action in question](https://github.com/pymodbus-dev/pymodbus/pull/1508/files) locks old issues (which I personally disagree with but it's Jan's call).  I don't see any reason to run it on pull requests, so it should be possible to remove that trigger.  That will fix this PR from your fork.
```yaml
  pull_request:
    branches:
      - dev
  schedule:
    # Every night at 02:20 UTC.
    - cron: '20 2 * * *'
```

jamesbraza:
I just tested what it's like to bump `pylint` to the current release, and it's pretty easy after this PR.  Just sharing another benefit

alexrudd2:
> I just tested what it's like to bump `pylint` to the current release, and it's pretty easy after this PR. Just sharing another benefit

On my repos I've enabled renovatebot, but I haven't suggested that here yet. o_0

janiversen (approved):
LGTM.

janiversen:
@alexrudd2 just for the record, I actually agree with you too, but for some reason in this repo most issues are reported on a closed issue and users do not open a new issue. That way we “loose” history (e.g. issues vs version).


The lock forces users to open a new issue (the issue remains open for 10 days, to allow comments relating to the close).

janiversen:
Thanks, looking forward to see more PR, like the bot you talked about (haven’t checked what it does).

PR #1519, opened 2023-04-25
pylint and pre-commit autoupdate

jamesbraza:
Modernizes `pre-commit` and `pylint` versions, implementing all suggested fixes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ default_language_version:
 
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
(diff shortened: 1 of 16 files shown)


janiversen:
Looks good to me, will wait for green light, to make the real review.

Would be nice if this could be split in 2, one with the code changes (first) and one with the configuration changes (second), that makes it easier to bisect if we run into problems later.

jamesbraza:
Alright, got the green light 🟢 

For stuff like this, I have seen it where old versions of tools have bugs that got fixed.  If one is to break the PR into two pieces, one would have to temporarily add more disable comments, just to remove them in the next PR.

janiversen:
Good argument merged.

PR #1521, opened 2023-04-25
Correct pylint configuration.

janiversen:
Allow pylint to work with broad exceptions again.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -194,7 +194,7 @@ deprecated-modules=regsub,TERMIOS,Bastion,rexec
 [pylint.exceptions]
 
 # Exceptions that will emit a warning when being caught.
-overgeneral-exceptions=builtins.Exception
+overgeneral-exceptions=Exception
 
 [pylint.deprecated_builtins]
 


janiversen:
@jamesbraza seems your latest commit broke pylint, I have no idea why our CI did not catch it

janiversen:
That was exactly what caused the problem.

jamesbraza:
Oh actually that change from `Exception` to `builtins.Exception` was intentional: https://stackoverflow.com/questions/75370815/pylint-specifying-exception-names-in-the-overgeneral-exceptions-option-without

This message comes with newer `pylint`, which https://github.com/pymodbus-dev/pymodbus/pull/1519 moved to. 
 `overgeneral-exceptions=builtins.Exception` is the correct configuration option.

Now after this PR, on my machine, I see all over the place:

```
pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.Exception' ?) instead.
pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.Exception' ?) instead.
pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.Exception' ?) instead.
```

Note the warning is actually asking for `builtins.Exception`.

janiversen:
sounds like a good idea, unless pylint continue telling me to remove the disable for a lot of lines (which for some reason seems solved with this PR).

PR #1520, opened 2023-04-25
Adding `flake8-pytest-style` to `ruff`

jamesbraza:
- Closes https://github.com/pymodbus-dev/pymodbus/issues/1500.
- Alphabetizes some `ruff` config lists

PR #1523, opened 2023-04-25
fallout from pylint change.

janiversen:
Seems that CI once again did not catch everything.

### DIFF
--- a/test/test_framers.py
+++ b/test/test_framers.py
@@ -83,7 +83,7 @@ def test_framer_initialization(framer):
 
 
 @pytest.mark.parametrize("data", [(b"", {}), (b"abcd", {"fcode": 98, "slave": 97})])
-def test_decode_data(rtu_framer, data):  # pylint: disable=redefined-outer-name
+def test_decode_data(rtu_framer, data):
     """Test decode data."""
     data, expected = data
     decoded = rtu_framer.decode_data(data)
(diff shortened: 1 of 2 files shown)


alexrudd2:
The [pylint cache](https://pylint.pycqa.org/en/v2.10.2/faq.html#where-is-the-persistent-data-stored-to-compare-between-successive-runs) is being preserved across versions.  On the CI runner the cache is `~/.cache/pylint`, which I guess gets included in the cache/restore Action.

I can reproduce locally:
1. Run `check_ci.sh` (or pylint directly) on 7b585d9ca292d402b34e4e8c08d53f9e83165397.  (This will pass, and create/update the cache).
2. Checkout b6384d02d3ff0a5a066eb39c9ce894aa1b3fb325
3. Run pylint again.  This will pass, but it should not
4. Clear the cache (in my case `~/Library/Caches/pylint`)
5. Run pylint again.  This will fail.

janiversen:
That's a good explanation.

jamesbraza:
> Seems that CI once again did not catch everything.

I can explain the issue here, it's a common issue when a repo doesn't require branches to be up-to-date before merging:

1. `pylint` update PR https://github.com/pymodbus-dev/pymodbus/pull/1519 was merged
2. Note that `ruff` in https://github.com/pymodbus-dev/pymodbus/pull/1520 made changes that `pylint` would have opinions about, but that branch was still using the old `pylint`
3. https://github.com/pymodbus-dev/pymodbus/pull/1520 was merged without a `git rebase` or a `git merge`
    - And its green checkmark was last attained with an old `pylint`
4. Thus, when CI was next run, `pylint` finally was able to express opinions on `ruff`

If `pymodbus` required branches to be up-to-date (a GitHub rule which I would advocate for), this issue would have surfaced within https://github.com/pymodbus-dev/pymodbus/pull/1520 after a `rebase` or `merge`.

Hope this explains things!

PR #1539, opened 2023-05-16
Added mechanism to determine if server did not start cleanly

couling:
Same rationale as #715.  I wonder if it got accidentally removed in the great 3.rewrite or if it was deliberately removed.

The key issue is starting a server and waiting for it to be up, also requires the unhappy path to be handled in determining that it will not come up.

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -534,6 +534,7 @@ class ModbusUnixServer:
 
         # asyncio future that will be done once server has started
         self.serving = asyncio.Future()
+        self.serving_done = asyncio.Future()
         # constructors cannot be declared async, so we have to
         # defer the initialization of the server
         self.server = None
@@ -552,6 +553,7 @@ class ModbusUnixServer:
                 Log.info("Server(Unix) listening.")
                 await self.server.serve_forever()
             except asyncio.exceptions.CancelledError:
+                self.serving_done.set_result(True)
                 raise
             except Exception as exc:  # pylint: disable=broad-except
                 Log.error("Server unexpected exception {}", exc)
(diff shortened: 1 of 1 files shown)


janiversen:
The patch looks good,  except it breaks a bunch of tests for some reason, please fix that.

couling:
Sure I'll look into it. @janiversen i'm wondering whether it's better to set(False) or set the exception. I'm not 100% sure myself.

janiversen:
Setting an exception is strange, you could raise an exception.

couling:
@janiversen I think that's what I mean. This could be written so that one or the other of these works either:

```python
if not await connection.serving:
    # Handle failed connection 
    ...
```
```python
try:
    await connection.serving
except IOError:
    # Handle failed connection 
    ...
```


The difference in this PR is between:
```python
self.serving.set_result(False)
```
```
self.serving.set_exception(sys.exc_info()[1])
```

janiversen:
The last can be made a lot easier:
```python
except IOError as exc:
    self.serving.set_exception(exc)
    raise
```

But independent of that:
1) you need to consider which values serving can have and how to test them "if serving" will run for both True and exc, and I suppose that is not your intention.
2) serving needs to be documented, especially if it contains more than True/False
3) the tests are still failing

janiversen:
Cleaned up the PR.

couling:
Apologies for not getting back sooner. It's been a busy few weeks.

janiversen:
I added a new future: serving_done.

Your concept could not work, because you cannot do set_result() on a future one, not multiple times.

couling:
Just to check that serving_done would fit the need, the aim is to have a single code path complete on success or failure.

The problem with the existing design was that you could not "continue when server ready":

- if you await the existing future your code hangs infinitely when the server cannot start
- the failure is only reported to the task that otherwise serves infinitely on success

so no single code path is certain to continue on both success and failure... here continue does not need to be success, it could be continue by raising an exception but it must exit and not await indefinitely.

 It's particularly problematic in unit tests, but has other uses.

PR #1543, opened 2023-05-22
Fix installation instructions

Ghostkeeper:
There was a typo here. And also, to be consistent with the rest of the instructions, I added the arrow on the line that must actually be executed.


Ghostkeeper:
I think the failing of the MacOS test is unrelated to this change.

janiversen (approved):
LGTM. Thanks.

PR #1544, opened 2023-05-22
Fix missing web content for simulator

corollaries:
Running `pymodbus.simulator` after installing pymodbus via pypi results in exceptions that certain files were not found.
Specifically the `simulator/web/generator/*` files. Upon further inspection it seems that these files are not packaged in the actual wheel, the simulator fails to start without these files. 

 Added the needed files for the simulator to `setup.py` to ensure that they're installed in the wheel. Verified locally that this works.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,5 +30,7 @@ del dependencies["required"]
 setup(
     install_requires=install_req,
     extras_require=dependencies,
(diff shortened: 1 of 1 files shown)


janiversen (approved):
LGTM. Thanks.

Issue #1551, opened 2023-05-27
client.read_coils with simulator fails


tstorek:
### Versions

- Python: 3.10 (fresh environment)
- OS: WIN10
- Pymodbus: 3.2.0 and 3.3.0a0
- Modbus Hardware (if used): None

### Pymodbus Specific

- Server: tcp async with simulator datastore
- Client: tcp sync/async

### Description

Hello, 
I am new to Modbus and trying to use the new simulator server to develop and test a Modbus to MQTT Gateway. 
I like the idea of having the values changing values during testing due to testing.
Although it seems rather complex to implement own actions like running a simulation. 

Just to make sure: Is the simulator with actions the right way to go? Or would be a datastore with Callbacks the better idea?

Using the example, getting the simulator up and running `examples\server_simulator.py` was pretty easy.
Having the simulator started with the example config, I was able to read the holding registers with sync and async client with `client.read_holding_registers`. This works like a charm for all datatypes, including `float32`. Surprisingly this also reads the "bits" section of the server.
As I understand from the docs (“bits”, defines registers that contain bits (discrete input and coils), this section would configure the coils of the virtual device.
But using, e.g., `client.read_coils(2, 1)` always fails with `Exception Response(129, 1, IllegalAddress)`.

Additionally, I understand the BITS for having boolean values. How can they have int values, then?

Using the `examples/server_async`, this works like a charm, but that does not have the `action` functionality. Do I need to write my own data store?

### Code and Logs

Here is the modified `run_sync_client` from the example. I only added the read commands.

```python

def run_sync_client(client, modbus_calls=None):
    """Run sync client."""
    _logger.info("### Client starting")
    client.connect()
    if modbus_calls:
        modbus_calls(client)
    rr = client.read_holding_registers(2, 1)
    print(rr)
    print(rr.registers)
    rr = client.read_coils(2,1)
    print(rr)
    print(rr.bits)
    client.close()
    _logger.info("### End of Program")

if __name__ == "__main__":
    testclient = setup_sync_client(description="Run synchronous client.")
    run_sync_client(testclient)
```

janiversen:
The simulator allow you to simulate a complete device, in examples you see an example configuration, and I think in examples/contrib are a couple of extra devices.

Using a datastore (there are no callbacks) is easier but a lot less flexible.

server_async.py is just the servers actions are are in the datastore simulator.

tstorek:
[...]

Issue #1472, opened 2023-04-06
Simulator returns IllegalAddress on read_coil functions using the correct register as address


corollaries:
### Versions

- Python: 3.10.6
- OS: Pop!_OS 22.04 LTS
- Pymodbus: 3.1.3

### Pymodbus Specific

- Server: Any
- Client: Any

### Description

I have configured registers with address 3 and 4 under `"bits"` and trying to execute a `READ_COIL` request on these specific registers, but keep on getting `IllegalAddress` as a response. After inspecting the simulator code I encountered the following line: [https://github.com/pymodbus-dev/pymodbus/blob/f815003721f91496874d677b3a1255bbe00b6f53/pymodbus/datastore/simulator.py#L556](https://github.com/pymodbus-dev/pymodbus/blob/f815003721f91496874d677b3a1255bbe00b6f53/pymodbus/datastore/simulator.py#L556) What is the purpose of dividing the address by 16? This line means that my address 3 gets turned into 0 which is why I'm getting the `IllegalAddress` exception. 

Invoking `READ_COIL` with address 48  does make this work but I don't think this behavior is desired. Based on the PR description that introduced this change: [https://github.com/pymodbus-dev/pymodbus/pull/1195](https://github.com/pymodbus-dev/pymodbus/pull/1195) it should be possible to read an address using both `read_coil` and `read_holding_register` but using the same address for both of these functions will result in `read_coil` failing. 

The sample code I've provided on the bottom should show this situation. Would love to know if this behavior is expected and/or there is possibly a bug needed to be fixed here. 

### Code and Logs

Client code:
```python
import time

from pymodbus.client import ModbusTcpClient
client = ModbusTcpClient('127.0.0.1', port=5022)

connected = client.connect()

if not connected:
    print("Connection failed")
    exit(1)

while True:
    response = client.read_coils(3, 1)
    if response.isError():
        print(f"Error occurred: {response}")
    else:
        print(f"Response: {response}")

    response = client.read_holding_registers(1, 1)
    if response.isError():
        print(f"Error occurred: {response}")
    else:
        print(f"Response: {response}")
    time.sleep(2)
```

Simulator code:
```python
import asyncio
from pymodbus.server import StartAsyncTcpServer
from pymodbus.datastore import ModbusServerContext, ModbusSimulatorContext

config = {
    "setup": {
        "co size": 250,
        "di size": 250,
        "hr size": 250,
        "ir size": 250,
        "shared blocks": True,
[...]

PR #1552, opened 2023-05-27
Allow reading/writing address 0 in the simulator

corollaries:
This PR is related to #1551 and #1472. Currently trying to attempt to read coils from bit 0 to 15 is not possible in the simulator and will result in an invalid address error. These bits get converted to the real address, which for bit 0 to 15 would be zero, and thus results in the invalid address error because of `validate()` in `simulator.py`. This is not only related to `read_coil`, but also other functions fail on address 0 while technically it should be possible to execute. 

From reading the documentation on all functions it seems that the API expects addresses and not registers, which leads to the expectation that address 0 (or bit 0-15) should be readable. The simulator also accepts setting address 0 in the configuration, but in it's current state it's not possible to read or write to this address.


janiversen:
Technically register 0 do not exist, some very old devices use it, but the simulator do not support it.

 If you define shared memory first possible coil is bit 16 (register 1). If you have separate blocks you should be able to do a read_coil of bit 0.

Which device are you trying to simulate that uses register 0 ?

janiversen:
btw the API clearly talks about registers or register address not addresses, because the modbus protocol only talks about registers, please see the simulator documentation where there are a drawing explaining the difference.

pymodbus/datastore/simulator.py
```diff
@@ -554,8 +554,9 @@ def validate(self, func_code, address, count=1):
             # Bit count, correct to register count
             count = int((count + WORD_SIZE - 1) / WORD_SIZE)
             address = int(address / 16)
+
         real_address = self.fc_offset[func_code] + address
-        if real_address <= 0 or real_address > self.register_count:
```
janiversen, line 558: This is not correct we do not use address 0, nor do we use register 0.

janiversen (changes requested):
You need some very convincing arguments why this should be allowed.

corollaries:
@janiversen 

Thanks for the review.

> Which device are you trying to simulate that uses register 0 ?

In our case we are using the simulator to test a client. The simulator isn't necessarily used to "reflect" a real world device, just to verify whether a client works correctly or not. Based on many examples / documentation we expected to be able to read/write on address 0. 

> btw the API clearly talks about registers or register address not addresses

If I look at the documentation for `read_coils`, `read_holding_registers`, etc the parameters are always `address`, hence the assumption that we're talking about addresses and not registers . As far as I can tell from the code these can be 0-indexed. Just not in combination with the simulator.
Example for `read_coil` docs:
> In the PDU Coils are addressed starting at zero. Therefore coils
    numbered 1-16 are addressed as 0-15.

The simulator in turn uses `addr` which are actually `registers`. It just makes using the simulator and the client IMO very confusing. Also given that we can actually set `"addr": 0` in the simulator configuration but somehow cannot use it. Even the example on the README of this repository would fail to run with the simulator as it's attempting to read/write coil 1. 

I'm fine with dropping this PR but maybe the following might then be a good course of action:
- Specify it very clearly in the docs that bit 0-15 and address 0 using the client cannot be used in combination with the simulator.
- Maybe rewrite the simulator configuration to:
```jsonc
"uint16": [
    { "register": 0, "value": 1 } /// making it clear that we're talking about registers here.
]
```
- Rewrite examples to not use address bit 0-15. People new to this library and/or modbus might copy examples to test something out real quick, to then run into trouble is confusing.

Hope this explains the rationale a bit.

janiversen:
It is true that both the modbus documentation and our API uses "register addresses" which are not the same as memory addresses (hence the reference to the official drawing). The difference between memory addresses and modbus protocol register addresses are indeed confusing, but see it as 2 different worlds that have 0 in common a server can map the register address to whatever memory it wants.

There are a lot of examples also in the real world that allows you to read coil 0, but in all these examples you do not have shared blocks, did you try to configure without shared blocks.

We have tried to be very clear in the documentation, especially regarding the difference between shared and non-shared blocks, which HIGHLY affects how to address registers and coils, but corrections are always welcome.

I have updated your branch to match the current dev, because I think you miss a couple of changes we made to the simulator (and other parts).

corollaries:
> did you try to configure without shared blocks.

Just tested it and it fails there as well. In the end the `validate()` method is always called, for bits `0-15` the real_address will be always be 0 and thus an `InvalidAddress` will be given. In whatever configuration we currently set the simulator, coils 0-15 will not work.

janiversen (approved):
I highly disagree, but admitted we have no tests to prove it is wrong.

I cannot guarantee this functionality will be kept when we do an overhaul of the simulator configuration (currently only ideas on a drawing board).

PR #1556, opened 2023-05-30
A few more type hints

alexrudd2:
- More type hints ( I had them lying around my computer and forgot)
- Shorter Exception [syntax](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement)


pymodbus/events.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, **kwargs):
         self.listen = kwargs.get("listen", False)
         self.broadcast = kwargs.get("broadcast", False)
 
-    def encode(self):
+    def encode(self) -> str:
```
janiversen, line 57: I think a message is bytes ?

pymodbus/events.py
```diff
@@ -64,7 +64,7 @@ def encode(self):
         packet = pack_bitstring(bits)
         return packet
 
-    def decode(self, event):
+    def decode(self, event) -> None:
```
janiversen, line 67: Would be nice to type event as well (as bytes).

pymodbus/framer/base.py
```diff
@@ -18,15 +19,16 @@ class ModbusFramer:
 
     name = ""
 
-    def __init__(self, decoder, client=None):
+    def __init__(self, decoder, client=None) -> None:
```
janiversen, line 22: Would be nice to type decoder and client as well.

janiversen (commented):
I have just made a couple of comments, you decide if you want to do them.

pymodbus/events.py
alexrudd2: Thanks, you are correct.  I'll annotate `pack_bitstring` and improve its docstring.

https://github.com/pymodbus-dev/pymodbus/blob/9c6acc385d2cb5362863cad740a90f693785f1e0/pymodbus/utilities.py#L109-L119

alexrudd2: Fixed in f5cd9fa817a5ec4fe19281c079d45e5307c3484e.

I think b212e79faf94379f59a0fb429230064d8c212ef6 is technically an API change (changing the name of an argument), although all current callers pass by value.

pymodbus/framer/base.py
alexrudd2: decoder I could do in f7b7517bc9dc5a3df5b01220ecb95b331fa579c6, client creates a circular import :(

pymodbus/events.py
janiversen: you are right technically It is, but there are no changes in the app, so lets not mark it.

API_Changes.rst signals changes the user need to consider/modify (at least that is the intention).

PR #1557, opened 2023-05-30
Switch to ruff's version of bandit

alexrudd2:
Ruff has now incorporated basically all Bandit rules we care about.  (The remaining ones pertain to e.g. Django or imports).

Rather than explicitly enabling rules using `bandit.yaml`, I just enabled them all and disabled two useless ones.

Note that ruff uses `Sxxx` to mean Bandit rule `Bxxx`


alexrudd2:
Default `bandit` actually flags the following code as violating [`B110`],(https://docs.openstack.org/bandit/1.4.0/plugins/try_except_pass.html):
https://github.com/pymodbus-dev/pymodbus/blob/67d875b785d4f79da7b76413ae62ad71a1259fc3/pymodbus/client/serial_asyncio/__init__.py#L434-L438

Even through `ruff` doesn't find this since it's in a module `__init__()` it should be easy enough to fix, by narrowing the caught Exception.

pymodbus/server/simulator/http_server.py
```diff
@@ -118,7 +118,7 @@ def __init__(
         self,
         modbus_server: str = "server",
         modbus_device: str = "device",
-        http_host: str = "0.0.0.0",
+        http_host: str = "localhost",
```
janiversen, line 121: That is wrong, local host is 127.0.0.1 whereas 0.0.0.0 allows bind on all interfaces.

janiversen (changes requested):

janiversen:
Apart from the wrong change it looks good to me.

alexrudd2:
> Or disable bandit for that line, I suppose. I have started rewriting that module so please do what is easiest for you.

I can't (yet) figure out what exceptions are actually thrown, ha.

pymodbus/server/simulator/http_server.py
janiversen: Please add The B code, 0.0.0.0 is the standard for listen.

janiversen (approved):
LGTM. Thanks

PR #1573, opened 2023-06-02
Fix non-shared block doc in config.rst.

haydenroche5:
The doc says holding registers come before input registers, but that's not what the code does. handle_setup_section puts the input registers first.


janiversen (approved):
Thanks for catching this.

PR #1575, opened 2023-06-03
Fix incorrect callback for transport_connect

### LINKED ISSUE #1574: Possibly incorrect invocation of transport_connect in reconnect_timer
```
```

alexrudd2:
Closes #1574

I would like to catch this in a test, but don't quite know how to write 
the test.


janiversen:
I know how (now). I will make it later today and make a new release as well.

janiversen (approved):
Thanks.

janiversen:
Seems something is not right with your change, let me check

alexrudd2:
Nope, this is incorrect.

PR #1576, opened 2023-06-03
Suppress asyncio error in callback for transport_connect

### LINKED ISSUE #1574: Possibly incorrect invocation of transport_connect in reconnect_timer

alexrudd2:
Closes #1574 

(Supercedes #1575, which was incorrect analysis of the problem).

I do not know why the lamda suppresses the `RuntimeWarning: coroutine 'BaseTransport.transport_connect' was never awaited` warning, but it does :)

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -368,8 +368,7 @@ class BaseTransport:
         )
         self.reconnect_timer = self.loop.call_later(
             self.reconnect_delay_current,
-            asyncio.create_task,
-            self.transport_connect(),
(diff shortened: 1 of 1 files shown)


janiversen:
I am not convinced this is the correct solution, but lets see if CI gives green light (the 2 errors was due to timeout on reconnect).

I am writing several test cases right now, to pin point the 2 problems (see #1577)

janiversen (changes requested):
Just refreshed my memory on lambda, they to do work with async methods, so this change only hides the fundamental problem.

I think we should solve the problem instead. Hope to have a solution during today.

PR #1577, opened 2023-06-03
Proof transport reconnect works.

### LINKED ISSUE #1574: Possibly incorrect invocation of transport_connect in reconnect_timer

janiversen:
I back ported a part from #1572 in order to keep life simple.

Add new test file test_reconnect.py to verify reconnect.

closes #1574


janiversen:
@alexrudd2 I know why pytest passed, that is being corrected here, and I will use this PR to chase down
- why windows have a reconnect problem
- and the missing await transport_connect
If you have tests or something that show the bug, please share.

janiversen:
@alexrudd2 seems my understanding also needs some updating 🥇 this is going to be interesting.

Actually I might try to avoid the call_later at all, use create_task and sleep in the method.

janiversen:
First reconnect now works nice, but second reconnect never happens...something about the tasks.

alexrudd2:
>   If you have tests or something that show the bug, please share.

https://github.com/numat/midas/actions/runs/5159705871/jobs/9294966924#step:7:24

My test does not fully fail, but if you expand the `pytest` step, you can see the warning in `3.3.0` (but not in others)

janiversen:
Thanks, I am hopeful that its all solved now.

janiversen:
@alexrudd2 please try dev, when you have time...if everything works for you, I can cut a release 3.3.1 in a couple of days.

alexrudd2:
> @alexrudd2 please try dev, when you have time...if everything works for you, I can cut a release 3.3.1 in a couple of days.

The latest `dev` makes the pytest warning I reported go away.  (I have not tested on real hardware yet).  

I still do not understand, but thank you!  :)

PR #1598, opened 2023-06-13
Fix write function codes

Fredo70:
Replace 0x15 with 0x0F and 0x16 with 0x10
Remove 0x17 (it should be 0x11) since this is a machine dependent code


janiversen:
Lgtm, thanks.

PR #1599, opened 2023-06-13
Fix validate and setValues in RemoteSlaveContext

Fredo70:
Previously validate wrote the data, but count instead of values.
validate has been changed to read the data and then write it again.
If it works, setValues can then write the "correct" data.
It's extra traffic and I don't like it very much. But this is how it works.


Fredo70:
Perhaps a better variant would be to pass values to validate?

janiversen:
In general we need the 2 functions, and normally validate() is non-intrusive so the more correct solution is to do a remote call in get_values / set_values. We should never do 2 calls since it can have bad consequences (e.g. for a stepper).

Fredo70:
Ok. I agree. I didn't like the solution with the 2 calls.
But since it is a "remote" context, validate has no chance to check the validity of the addresses without making a call and trying to perform the operation.
I see 2 possibilities:

1. validate makes the request (that's what was done before).  But in this case validate needs values as argument. setValues would then do nothing.
2. validate does nothing and returns only True. In this case, however, the calling execute method in the requesters must catch the exception and pass it on with doException. Otherwise this ends in an "Unknown exception forcing disconnect". 

What do you think is better? Or are there other solutions?
If I have time today, I will try solution 2. That would satisfy the condition that validate is not intrusive.

janiversen:
I think its best for validate() to always return true.

janiversen:
Now I got back to my computer, and I have to correct myself.

The correct procedure is as follows:

validate():

for read commands pass it on, preferable store the result as something like self.__class__.values.

Remark if we get a python exception, something is very wrong ! device exception (like illegal function) are returned in the response.

For write commands it need to convert the fc to the corresponding read command, and continue with the first part.


get_values():

use self.__class_.values (this is optional just to avoid a second read)

set_values():
Issues the write command as normal.

Sure it causes 2 requests for every write, but write should be rare compared to read, so it does not really matter.

Sorry for the confusion, I had my mind on the part that is using the datastore, for another reason.

Fredo70:
Ok. For getValues() it's not a problem. I don't touch this function.
But sending a read command to validate the write is not enough. Because it could be that an attempt is made to write to a readonly address. In this case, the validation would be ok (registers can be read). But writing to setValues would throw an IllegalAddress Exception.
This exception is not caught and so the connection is disconnected.

janiversen:
This is what modbus devices do, the world is not perfect.

In case of a readonly register, there are NO python exception and NO is disconnect, the response from the device is passed to the application

```
rr = write_register
if rr.is_error():
   do something
```

Fredo70:
Well, I could catch is_error in setValues. But the WriteMultipleRegistersRequest sould call `doException(merror.IllegalAddress)`.
That's why I think this should be caught in validate. setValues returns nothing. In WriteMultipleRegistersRequest the execute method would return a WriteMultipleRegistersResponse normally. But nothing was written.

> In case of a readonly register, there are NO python exception and NO is disconnect

Ok. If the excecption is not catched in setValues, then why do I get Unknown `exception "..." on stream ... forcing disconnect`?

janiversen:
Your exception must come from something else, please show the trace, you might have found another bug

In the normal server setup, validate() is called but in reality not used for anything. In other server setups like the simulator the validate() call is used actively. In any case the caller do not always have the values to write when calling validate().

In the medium term, I would like to combine the 2 calls. This is however a bit more simple search/replace. It needs to be done in all datastore, the calling code and the test cases.

janiversen:
This is the typical pattern of how we use datastore in the server (register_read_message.py):
```
    def execute(self, context):
        """Run a read holding request against a datastore.

        :param context: The datastore to request from
        :returns: An initialized :py:class:`~pymodbus.register_read_message.ReadHoldingRegistersResponse`, or an :py:class:`~pymodbus.pdu.ExceptionResponse` if an error occurred
        """
        if not (1 <= self.count <= 0x7D):
            return self.doException(merror.IllegalValue)
        if not context.validate(self.function_code, self.address, self.count):
            return self.doException(merror.IllegalAddress)
        values = context.getValues(self.function_code, self.address, self.count)
        return ReadHoldingRegistersResponse(values)
```

Fredo70:
Ok. I have expressed myself wrongly. I'll try it again. When reading or writing on an illegal address, the result of getValues will be an ExceptionResponse.
The ExceptionResponse cannot be encoded (has no len()). So it ends in an "Unknown exception" and the connection is broken.
The result of setValues is nothing (setValues returns nothing). So nothing happens. It cannot be written, but the client does not "notice" that.
I will think again about the best way to solve this. But if you have a solution, I'd love to hear it.

janiversen:
The client do receive a response to a write request. Our test harness do test that. If you issue a write in the datastore, then you act as a client and must catch exception etc as a client should do. (see examples/client_calls.py that contains template calls).

janiversen:
As I wrote earlier the best, but also with most code change, is to integrate validate() into get_values() and set_values(), that way set_values can return true/false

janiversen:
Did you by the way get a stack trace of the exception, I need it to see what happens..

It also helps you a lot to run the server with "--log debug" that allows you to see what is being sent and received (e.g. see the device response to the write request).

Fredo70:
```
  File "/usr/local/lib/python3.8/dist-packages/pymodbus/server/async_io.py", line 213, in handle
    self.framer.processIncomingPacket(
  File "/usr/local/lib/python3.8/dist-packages/pymodbus/framer/socket_framer.py", line 173, in processIncomingPacket
    self._process(callback, tid)
  File "/usr/local/lib/python3.8/dist-packages/pymodbus/framer/socket_framer.py", line 188, in _process
    callback(result)  # defer or push to a thread?
  File "/usr/local/lib/python3.8/dist-packages/pymodbus/server/async_io.py", line 215, in <lambda>
    callback=lambda x: self.execute(x, *addr),
  File "/usr/local/lib/python3.8/dist-packages/pymodbus/server/async_io.py", line 285, in execute
    self.send(response, *addr, skip_encoding=skip_encoding)
  File "/usr/local/lib/python3.8/dist-packages/pymodbus/server/async_io.py", line 301, in send
    pdu = self.framer.buildPacket(message)
  File "/usr/local/lib/python3.8/dist-packages/pymodbus/framer/socket_framer.py", line 211, in buildPacket
    data = message.encode()
  File "/usr/local/lib/python3.8/dist-packages/pymodbus/register_read_message.py", line 90, in encode
    result = struct.pack(">B", len(self.registers) * 2)
TypeError: object of type 'ExceptionResponse' has no len()
2023-06-15 08:21:57,268 ERROR logging:114 Unknown exception "object of type 'ExceptionResponse' has no len()" on stream ('192.168.32.1', 64543) forcing disconnect
```

janiversen:
ahh if you add debug, you will see that your device did not respond and most likely closed the connection.

Fredo70:
> ahh if you add debug, you will see that your device did not respond and most likely closed the connection.

No. My device responds. And it does not close the connection either.

janiversen:
Then please use pymodbus_apply_log("DEBUG") and let's see what pymodbus receive and of course the trace that follows. No len is a very abnormal situation

janiversen:
No problem, just be aware that for us a debug log is worth a lot more than an explanation. We use the debug log to try and reproduce the problem, without a debug log and a stack trace when available, it's all guess work and not really worth the time.

I look forward to see the PR updated so we can review it again.

janiversen:
That looks a lot like a bug in the socket_framer, let me simulate it.

janiversen:
Finally I understand why you get a python exception. It does not happen in the client, it happens because get_values() returns the response from the client, without validation, this means the exception response is returned, but the caller of get_values expects an array, hence the exception.

By having validate() always return true, you are effectively doing what I suggested (without changing a lot of source), but it could very well work.

Why not submit your change as a new PR (or update this one), so that we at least can get that part fixed, while you continue to think about set_values() ?

janiversen:
If CI passes I will approve it and merge it, so it can be basis for your further work.

Fredo70:
> Finally I understand why you get a python exception. It does not happen in the client, it happens because get_values() returns the response from the client, without validation, this means the exception response is returned, but the caller of get_values expects an array, hence the exception.

It was exactly what I tried to explain. :-)

I have developed a new proposal. My tests were positive. Can you still see if it fits for you and if I haven't forgotten anything? Thank you.

janiversen:
will do, in the meantime you have pylint errors in line 35 and 54  (see CI run).

Unused arguments needs to start with _ to signal they are unused.

janiversen (approved):
LGTM, but you need fix the CI problem

Fredo70:
> LGTM, but you need fix the CI problem

I am struggling with it... sorry python is not my cup of tea ;-)

Fredo70:
pytest failed. No idea why and what I have to do. Sorry.

janiversen:
?? if you look at the CI output, it tells you that your problem is in

test/test_remote_datastore.py:67: AssertionError

as I suspected we test that validate returns false sometimes that test need to be corrected for remote_datastore.

Fredo70:
Ok. I see. I won't get around to correcting it until tomorrow.

janiversen:
No problem take your time.

janiversen (approved):
I like your test changes.

janiversen:
Thanks, looking forward to see more from you.

PR #1603, opened 2023-06-14
Python 3.11, versioning (pep 440).

janiversen:
It seems 3.4.0alfa is no longer allowed in python 3.11, it needs to be 3.4.0.alfa


janiversen:
False alarm, this was a caching issue.

Issue #1549, opened 2023-05-24
writing register does not work when running through forwarder


t-dan:
### Versions

- Python: 3.11.3
- OS: Arch Linux
- Pymodbus: 3.2.2
- Modbus Hardware (if used): none

### Pymodbus Specific

- Server: tcp - async
- Client: tcp - async

### Description

I am experimenting with writing serial to ETH MODBUS forwarder. I have started from [https://pymodbus.readthedocs.io/en/latest/source/examples.html#serial-forwarder](serial-forwarder) example, modyfying it a bit to forward TCP at one port to another (for testing purposes):

code snippet of the forwarder is:
```c = ModbusTcpClient(host=127.0.0.1', port=target_port)
c = ModbusServerContext(slaves=RemoteSlaveContext(c, slave=0), single=True)

p = int('1'+str(target_port))  # forwarder port, forwards to target_port
s = ModbusTcpServer(c, address=('127.0.0.1', p), allow_reuse_address=True)
s = asyncio.create_task(s.serve_forever())
```

When a client tries to write to the server directly (on target_port), the value is changed and the new value is correctly read.
However, when I try to do the same through forwarder (port f"1{target_port}"), the wrong value is written.

The client looks like:
```client = ModbusClient(host='127.0.0.1', port=port)

rd = client.read_holding_registers(4,count=1)
print('read orig: ',rd.registers)

client.write_register(4,7)  # write new value

rd = client.read_holding_registers(4,count=1)
print('read changed: ',rd.registers)
```

There is ModbusSimulator serving as a server, holding registers are 16bit.

### Code and Logs

Calling twice, going directly to target port, writing 7 when an original value is 5

```
read orig:  [5]
read changed:  [7]
```
```
read orig:  [7]
read changed:  [7]
```

And going through forwarder  to f"1{target_port}", again writing 7 when an original value is 5

```
read orig:  [5]
read changed:  [1]
```
```
read orig:  [1]
read changed:  [1]
```

janiversen:
hmmm looks like a bug, let me see if I can reproduce it.

janiversen:
Just noted that you are using an example in contrib. As documented contrib contains user contributions that we do not check.

Please try with example modbus_forwarder.py which we have tested.

janiversen:
Close as it affects the "as is" examples.

t-dan:
I have tried to use `modbus_forwarder.py`, starting it as:

```
[...]

PR #1625, opened 2023-06-23
Change to "sync client" in forwarder example

Fredo70:
I replace the `AsyncModbusTcpClient` with `ModbusTcpClient`.
The problems caused with the Async Client are discussed here:
https://github.com/pymodbus-dev/pymodbus/discussions/1593#discussioncomment-6145862
https://github.com/pymodbus-dev/pymodbus/issues/1549#issuecomment-1563466215

An error in the `ModbusTcpClient` in the method `connected` was also fixed.

### DIFF
--- a/examples/modbus_forwarder.py
+++ b/examples/modbus_forwarder.py
@@ -20,7 +20,7 @@ import asyncio
 import logging
 
 from examples import helper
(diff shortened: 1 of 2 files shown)


janiversen:
Thanks.

PR #1627, opened 2023-06-23
fix getFrameStart for ExceptionResponse

Fredo70:


janiversen (approved):
Good catch, this might also be present in the other frame handler. Do you mind checking?

janiversen:
If you run with our venv (see README.RST) then most things are caught in precommit.

For mypy etc run ./check_ci.sh

Fredo70:
> this might also be present in the other frame handler. Do you mind checking?

I'm not sure. I didn't see this construct in the other Framers.

Fredo70:
> If you run with our venv (see README.RST) then most things are caught in precommit.

Unfortunately, I develop under Windows. I already have an old Linux laptop somewhere. But it's too slow.
I have to write down that I have to keep these things in mind. I like to be forgetful ;-)

janiversen:
Python is the same on windows command line, The only script we have is check_ci.sh and that is just a few python commands.

PR #1634, opened 2023-06-28
ModbusUdpServer now respects loop kwargs and updates docstrings

znazter:
Harmonize loop handling since TCP and RTU servers allow users to supply their own loop UDP should as well.

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -319,6 +319,7 @@ class ModbusTcpServer(Transport):
                         False to treat 0 as any other slave_id
         :param response_manipulator: Callback method for manipulating the
                                         response
+        :param loop: Optionally allow users to supply their own event loop
         """
         if not address:
             address = ("", 502)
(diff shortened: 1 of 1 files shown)


janiversen (commented):
Please explain the advantage of this change. We removed loop= a while ago in the documentation, because we cannot see a case where you need to supply the loop.

the code in general does "self.loop = asyncio,get_running_loop()", actually in several places.

janiversen (changes requested):
If loop is a legal parameter, it needs to be specified in parameter list. kwargs are only for experimental and internal parameters.

But first we need to understand why you need this change back ?

znazter:
Hi again, 

The short answer is that we have some code working for RTU and TCP servers where the loop is handled elsewhere, when adding the UDP server now it is no longer possible to supply the loop, which causes me to have to rewrite a bunch of things. I understand now that i should not supply the loop manually according to asyncio, but i understand that now. But as this seems to be the way to go moving forward i will just rewrite the code now rather than later.

Thank you.

Issue #1661, opened 2023-07-17
Simulator with defalut config gives back different byte order between float32 and uint32


fhydralisk:
### Versions

- Python: 3.9.x
- OS: macos 13.4
- Pymodbus: 3.2.2
- Modbus Hardware (if used): None

### Pymodbus Specific

- Server: tcp, simulator
- Client: telegraf

### Description

Hi, I run my pymodbus simulator by starting main.py in pymodbus/datastore/simulator with default config file and started a telegraf program which fetches data from this pymodbus server. 

We can see in config that register of address [04, 05] holds value of uint32: 617001 and [06, 07] holds float32: 404.17. 
The telegraf(client or master) side is configured to fetch uint32 value from 04, 05 address and float32 value from 06, 07. It is also configured to use BIG ENDIAN  byte order to decode fetched data.

What I expected is that both of uint32 value and float32 value are fetched correctly, or at least, all of them are fetched incorrectly. However, the result shows that uint32 is correctly fetched with decoded value of 617001 and the float32 value is decoded with wrong value of  -299.5801696777344.

I did some study and found that bytes [195 149 202 67] represent 404.17 in little endian byte order. Is this a  feature or a bug?

Sorry for my bad Englsih, hope someone can understand what I'm talking about.

### Telegraf Logs

```
2023/07/17 17:12:10 D! [inputs.modubs] Reading slave 1 for tcp://localhost:5020...
2023/07/17 17:12:10 D! [inputs.modubs] trying to read input@4[4]...
2023/07/17 17:12:10 D! [inputs.modubs] got input@4[4]: [0 9 106 41 195 149 202 67]
2023/07/17 17:12:10 D! [inputs.modubs]   field int_test with offset 0 with len 4: [0 9 106 41] --> 617001
2023/07/17 17:12:10 D! [inputs.modubs]   field float_test with offset 4 with len 4: [195 149 202 67] --> -299.5801696777344
```

janiversen:
Modbus always transmit in big endian, the simulator convert the value locally if the CPU uses little endian.

Please make a debug log in the simulator, so we can see what is actually transmitted. all types of 32bit should be converted (if needed).

fhydralisk:
FYI

```
(pymodbus) localhost:pymodbus hydra$ python pymodbus/server/simulator/main.py --log debug
2023-07-17 18:45:19,380 INFO  logging:96 Start simulator
[...]

PR #1664, opened 2023-07-18
Bug fix for #1661

fhydralisk:


pymodbus/datastore/simulator.py
```diff
@@ -781,5 +781,5 @@ def build_value_from_registers(cls, registers, is_int):
         if is_int:
             value = int.from_bytes(value_bytes, "big")
         else:
-            value = struct.unpack("f", value_bytes)[0]
+            value = struct.unpack(">f", value_bytes)[0]
```
janiversen, line 784: Good catch !!, funny thing is that I am on a M1 MacMini and did not see it.

janiversen (approved):
Thanks.

PR #1679, opened 2023-07-20
Run Linters on python 3.11 as well.

janiversen:


janiversen:
This cannot work, because pylint do not have a "disable for specific python version" switch.

PR #1685, opened 2023-07-20
Fix broken sync client test

### LINKED ISSUE #1684: Invalid client tests that always pass
https://github.com/pymodbus-dev/pymodbus/blob/152ad8dfd64882026b8b7a9284d7912879a78c23/test/test_client_sync.py#L192-L203

https://github.com/pymodbus-dev/pymodbus/blob/152ad8dfd64882026b8b7a9284d7912879a78c23/test/test_client_sync.py#L281-L292

These tests will _always_ pass.  even if the call were changed to `client.register('WTFBBQ')`, `client.register(NotImplementedError)`, or anything else.

(1) `mock.called_once_with()` does not exist.  It should be `mock.assert_called_once_with()`.
(2) The `assert` at the beginning of the line will pass on any truth-y value, including a _mocked method_.

alexrudd2:
Closes #1684

### DIFF
--- a/test/test_client_sync.py
+++ b/test/test_client_sync.py
@@ -200,7 +200,7 @@ class TestSynchronousClient:  # pylint: disable=too-many-public-methods
         client = ModbusTcpClient("127.0.0.1")
         client.framer = mock.Mock()
         client.register(CustomRequest)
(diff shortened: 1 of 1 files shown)


janiversen (approved):
LGTM, thanks.

PR #1688, opened 2023-07-21
Removed `requests` dependency from `contrib/explain.py` 

jamesbraza:
I learned the built-in `urllib.request` can do the trick, so let's just roll with built-in only code, no need for third party libraries.


jamesbraza:
@janiversen I think this is fine, the test errors are unrelated to this PR, seem to be flaky tests.  Leaving them here so you can look at them.

janiversen (approved):
LGTM, thanks.

PR #1687, opened 2023-07-21
Fix help for REPL client missing some client methods

alexrudd2:
Closes #1686

PR #1689, opened 2023-07-21
Update ruff to 0.0.277

alexrudd2:
See #1687 and #1685; those should be merged first.

We have decided that implicit Optional is OK already (example `foo: Optional[bar] = None` is 
unnecessary`).  The `typing.ClassVar` rule also seems noisy.


janiversen (approved):
LGTM, thanks

I learned something here, I did not know that !s worked.

alexrudd2:
> I learned something here, I did not know that !s worked.

I didn't either :). The `pyupgrade` rules in `ruff` have taught me many new things!  I'm also impressed with the subtle bugs it's starting to catch.

Issue #1695, opened 2023-07-24
RTU regression in 3.3.0 for custom message class


wlcrs:
### Versions

- Python: 3.11.4
- OS: Linux
- Pymodbus: Works fine up until 3.2.2. Regression started in 3.3.0
- Modbus Hardware (if used):

### Pymodbus Specific

- Client: rtu async

### Description

I have multiple people reporting a failure to communicate over Modbus RTU using a custom message class which has `_rtu_byte_count_pos` set. When they downgrade to pyModbus 3.2.2 it starts working again. The same code also works fine for connections over Modbus TCP.

cfr. https://github.com/wlcrs/huawei_solar/issues/426 and https://github.com/wlcrs/huawei_solar/issues/422#issuecomment-1646764646

The logs show that the data arrives in two parts, and that the first part gets discarded because the frame is invalid. However, the data should be kept until the frame is complete.

## Code

```python

class AsyncHuaweiSolarModbusSerialClient(AsyncModbusSerialClient):
    """Custom SerialClient with support for custom Huawei modbus messages"""

    def __init__(self, port, baudrate, timeout: int, **serial_kwargs):
        super().__init__(port, **serial_kwargs, baudrate=baudrate, reconnect_delay=RECONNECT_DELAY, timeout=timeout)
        super().register(PrivateHuaweiModbusResponse)

...

class PrivateHuaweiModbusResponse(ModbusResponse):
    """Response with the private Huawei Solar function code"""

    function_code = 0x41
    _rtu_byte_count_pos = 3

    def __init__(self, **kwargs):
        ModbusResponse.__init__(self, **kwargs)

        self.sub_command = None
        self.content = b""

    def decode(self, data):
        self.sub_command = int(data[0])
        self.content = data[1:]

    def __str__(self):
        return f"{self.__class__.__name__}({self.sub_command})"

...

class StartUploadModbusRequest(ModbusRequest):
    """
    Modbus file upload request
    """

    function_code = 0x41
    sub_function_code = 0x05

    def __init__(self, file_type, customized_data: t.Optional[bytes] = None, **kwargs):
        ModbusRequest.__init__(self, **kwargs)
        self.file_type = file_type

        if customized_data is None:
            self.customised_data = b""
        else:
            self.customised_data = customized_data

    def encode(self):
        data_length = 1 + len(self.customised_data)
        return struct.pack(">BBB", self.sub_function_code, data_length, self.file_type) + self.customised_data

    def decode(self, data):
[...]

PR #1696, opened 2023-07-24
Transport emulator

janiversen:
This is part of streamlining of our test suite, to avoid the spontaneous CI errors, and make tests even faster.

Add possibility to manipulate send/receive messages.

Needed to solve #1695 (and future problems that needs manipulation of the data stream).


janiversen:
wrong branch name.

Issue #1711, opened 2023-07-30
pymodbus.server does not start in 3.4.1


chkr-private:
### Versions

- Python: 3.11.4
- OS: Fedora 37
- Pymodbus: 3.4.1
- Modbus Hardware (if used): None

### Pymodbus Specific

- Server: tcp

### Description

- "pymodbus.server run" does not start

Error message:
```pymodbus.server run
Traceback (most recent call last):

  File "/usr/bin/pymodbus.server", line 8, in <module>
    sys.exit(app())
             ^^^^^

  File "/usr/lib/python3.11/site-packages/pymodbus/repl/server/main.py", line 190, in run
    app = ReactiveServer.factory(
          ^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/lib/python3.11/site-packages/pymodbus/server/reactive/main.py", line 505, in factory
    server = server(
             ^^^^^^^

TypeError: ModbusTcpServer.__init__() got an unexpected keyword argument 'loop'
```

### Code and Logs
As a workaround, the following patch for class ModbusTcpServer helps but I'm unsure whether the "loop" parameter is actually needed or not.

```
--- server/async_io.py.old	2023-07-30 11:11:19.538544170 +0200
+++ server/async_io.py	2023-07-30 11:11:32.463766887 +0200
@@ -269,6 +269,7 @@
         broadcast_enable=False,
         response_manipulator=None,
         request_tracer=None,
+        **kwargs,
     ):
         """Initialize the socket server.
 ```

The problem looks similar to #1662, but this time with the REPL stand-alone tool pymodbus.server.

janiversen:
It works as a workaround, but clearly the loop is not used as parameter and adding warns just hides the problem.

The caller should not add loop, that is the correct solution.

Feel free to submit a PR and I will review it.

chkr-private:
> It works as a workaround, but clearly the loop is not used as parameter and adding warns just hides the problem.

Unfortunately, there were additional superfluous parameters.

> Feel free to submit a PR and I will review it.

Please see #1714.

janiversen:
You are right, the main problem is that there are no tests trying to start the REPL server so problems are not detected.

janiversen:
This works now on dev.

chkr-private:
> This works now on dev.

I tested dev at commit 08529612b11966e4d16d085b7550d74737472b67 and the problem is still reproducible. It looks like that #1714 is not merged yet.

PR #1714, opened 2023-07-30
RFC: Reduce parameters for REPL server classes

chkr-private:
In order to avoid using **kwargs in Modbus{Tcp,Tls,Udp}Server, I had to remove all superfluous arguments. I did a little bit of testing, but I don't have the full overview of the impact of all changes. Please review and let me know your thoughts!

- allows to run "modbus.server run" with server classes which use specific argument list without "**kwargs" in their init methods
- remove "handler" (all handles are now mapped to ModbusServerRequestHandler)
- remove "allow_reuse_port" (not used otherwise)
- loop (not used anymore)
- rename paramter "broadcast" to "broadcast_enable" as used in server classes
- fixes issue #1711

### DIFF
--- a/API_changes.rst
+++ b/API_changes.rst
@@ -8,6 +8,7 @@ Version 3.5.0 (future)
 - Remove handler parameter from ModbusUdpServer
 - Remove handler from default config
 - Remove allow_reuse_port from default config
+- Remove loop parameter from ModbusSerialServer
 
 -------------
 Version 3.4.2
--- a/API_changes.rst
+++ b/API_changes.rst
@@ -6,7 +6,7 @@ PyModbus - API changes.
 Version 3.5.0 (future)
 -------------
 - Remove handler parameter from ModbusUdpServer
-- Remove handler from default config
+- Remove handler from repl default config
 - Remove allow_reuse_port from default config
 - Remove loop parameter from ModbusSerialServer
 
--- a/API_changes.rst
+++ b/API_changes.rst
@@ -3,6 +3,13 @@ PyModbus - API changes.
 =======================
 
 -------------
+Version 3.5.0 (future)
+-------------
+- Remove handler parameter from ModbusUdpServer
+- Remove handler from default config
+- Remove allow_reuse_port from default config
+
+-------------
 Version 3.4.2
 -------------
 -ModbusSerialServer now accepts request_tracer=.
--- a/API_changes.rst
+++ b/API_changes.rst
@@ -7,7 +7,7 @@ Version 3.5.0 (future)
 -------------
 - Remove handler parameter from ModbusUdpServer
 - Remove handler from repl default config
-- Remove allow_reuse_port from default config
+- Remove allow_reuse_port from repl default config
 - Remove loop parameter from ModbusSerialServer
 
 -------------
--- a/pymodbus/repl/server/main.py
+++ b/pymodbus/repl/server/main.py
@@ -1,5 +1,6 @@
 """Repl server main."""
 import asyncio
+import contextlib
 import json
 import logging
 import sys
@@ -180,19 +181,26 @@ def run(
 
     modbus_config["randomize"] = randomize
     modbus_config["change_rate"] = change_rate
-    app = ReactiveServer.factory(
-        modbus_server,
-        framer,
-        modbus_port=modbus_port,
-        slave=modbus_slave_id,
-        single=False,
-        data_block_settings=data_block_settings,
-        **web_app_config,
-        **modbus_config,
-    )
-    loop.run_until_complete(app.run_async(repl))
+
+    async def _wrapper():
+        app = ReactiveServer.factory(
+            modbus_server,
+            framer,
+            modbus_port=modbus_port,
+            slave=modbus_slave_id,
+            single=False,
+            data_block_settings=data_block_settings,
+            **web_app_config,
+            **modbus_config,
+        )
+        await app.run_async(repl)
+        return app
+
+    app = loop.run_until_complete(_wrapper())
     if repl:
-        loop.run_until_complete(run_repl(app))
+        with contextlib.suppress(asyncio.CancelledError):
+            loop.run_until_complete(run_repl(app))
+
     else:
         loop.run_forever()
 
--- a/pymodbus/repl/server/main.py
+++ b/pymodbus/repl/server/main.py
@@ -185,7 +185,6 @@ def run(
         framer,
         modbus_port=modbus_port,
         slave=modbus_slave_id,
-        loop=loop,
         single=False,
         data_block_settings=data_block_settings,
         **web_app_config,
--- a/pymodbus/repl/server/main.py
+++ b/pymodbus/repl/server/main.py
@@ -16,7 +16,6 @@ from pymodbus.repl.server.cli import run_repl
 from pymodbus.server.reactive.default_config import DEFAULT_CONFIG
 from pymodbus.server.reactive.main import (
     DEFAULT_FRAMER,
-    DEFUALT_HANDLERS,
     ReactiveServer,
 )
 
@@ -114,7 +113,7 @@ def server(
         "repl": repl,
         "host": host,
         "web_port": web_port,
-        "broadcast": broadcast_support,
+        "broadcast_enable": broadcast_support,
     }
 
 
@@ -178,13 +177,7 @@ def run(
     data_block_settings = modbus_config.pop("data_block_settings", {})
     modbus_config = modbus_config.get(modbus_server, {})
     modbus_config = process_extra_args(extra_args, modbus_config)
-    if modbus_server != "serial":
-        handler = modbus_config.pop("handler", "ModbusConnectedRequestHandler")
-    else:
-        handler = modbus_config.pop("handler", "ModbusSingleRequestHandler")
-    handler = DEFUALT_HANDLERS.get(handler.strip())
 
-    modbus_config["handler"] = handler
     modbus_config["randomize"] = randomize
     modbus_config["change_rate"] = change_rate
     app = ReactiveServer.factory(
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -572,7 +572,7 @@ class ModbusSerialServer(ModbusProtocol):
             True,
         )
 
-        self.loop = kwargs.get("loop") or asyncio.get_event_loop()
+        self.loop = asyncio.get_running_loop()
         self.handle_local_echo = kwargs.get("handle_local_echo", False)
         self.ignore_missing_slaves = kwargs.get("ignore_missing_slaves", False)
         self.broadcast_enable = kwargs.get("broadcast_enable", False)
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -449,8 +449,6 @@ class ModbusUdpServer(ModbusProtocol):
         :param framer: The framer strategy to use
         :param identity: An optional identify structure
         :param address: An optional (interface, port) to bind to.
-        :param handler: A handler for each client session; default is
-                            ModbusDisonnectedRequestHandler
         :param ignore_missing_slaves: True to not send errors on a request
                             to a missing slave
         :param broadcast_enable: True to treat slave_id 0 as broadcast address,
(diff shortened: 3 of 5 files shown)


pymodbus/server/async_io.py
```diff
@@ -472,7 +470,7 @@ def __init__(
             True,
         )
 
-        self.loop = asyncio.get_running_loop()
```
janiversen, line 475: This looks wrong, it creates a new loop, the loop must be created outside the class.

pymodbus/server/async_io.py
```diff
@@ -574,7 +572,7 @@ def __init__(
             True,
         )
 
-        self.loop = kwargs.get("loop") or asyncio.get_event_loop()
+        self.loop = asyncio.get_event_loop()
```
janiversen, line 575: please use get_running_loop

janiversen (changes requested):
A few small changes, you also need to update API_CHANGES.rst to make users aware of the API changes.

janiversen:
Feel free to make another PR that tests starting the repl server. Just starting/stopping the server would be a huge jump from what we have.

pymodbus/server/async_io.py
chkr-private: When I do this, pymodbus.server doesn't start with the following error message:
```
pymodbus.server --verbose run -s serial -p ./device/p1 -f rtu

  File "/usr/bin/pymodbus.server", line 8, in <module>
    sys.exit(app())
             ^^^^^

  File "/usr/lib/python3.11/site-packages/pymodbus/repl/server/main.py", line 183, in run
    app = ReactiveServer.factory(
          ^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/lib/python3.11/site-packages/pymodbus/server/reactive/main.py", line 498, in factory
    server = server(context, framer=framer, identity=identity, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/lib/python3.11/site-packages/pymodbus/server/async_io.py", line 575, in __init__
    self.loop = asyncio.get_running_loop()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^

RuntimeError: no running event loop
```
I'm afraid I don't fully understand the design here - especially which class is responsible for creating the event loop...

pymodbus/server/async_io.py
chkr-private: It looks like that an event loop is created (outside) in pymodbus/repl/server/main.py and before my patch is was set as an explicit parameter when creating the servers:

```
    app = ReactiveServer.factory(
        modbus_server,
        framer,
        modbus_port=modbus_port,
        slave=modbus_slave_id,
        loop=loop,
        single=False,
        data_block_settings=data_block_settings,
        **web_app_config,
        **modbus_config,
    )
```
- ModbusSerialServer was indeed using it (via evaluating kwargs.get("loop") )
- ModbusTcpServer complaint about an superfluous parameter (didn't use the event loop)
- ModbusUdpServer complaint about an superfluous "loop" parameter and I removed it, it tried to use "get_running_loop()" but it didn't work either because there was no running loop

Should the following work: as before, use the loop parameter for the Modbus{Udp,Tcp,Tls,Serial}Server, make it explicit and in case the specific server doesn't need it (looks like that the TCP-based classes don't use it), just ignore the parameter?

pymodbus/server/async_io.py
janiversen: I am afraid there are no bigger design, but merely a not well maintained corner.

main needs to start the app with asyncio.run, that way the loop will exist. Which might need to be done in 2 steps using an async helper method because app is not async.

asyncio.py contains convenience functions to do that (startXserver), but repl uses some old structure.

pymodbus/server/async_io.py
chkr-private: I just pushed 2 commits which would implement the variant described above. Please have a look - any feedback is welcome.

pymodbus/server/async_io.py
```diff
@@ -432,6 +435,7 @@ class ModbusUdpServer(ModbusProtocol):
     def __init__(
         self,
         context,
+        loop=None,
```
janiversen, line 438: no please.

pymodbus/server/async_io.py
```diff
@@ -472,7 +474,7 @@ def __init__(
             True,
         )
 
-        self.loop = asyncio.get_running_loop()
+        self.loop = loop or asyncio.get_running_loop()
```
janiversen, line 477: No loop parameter please.

janiversen (changes requested):
Your last commits caused CI to fail. I normally do not review a PR before CI passes, but decided to make an exception, please in the future solve CI before asking for a review

I gave you a simple solution to the loop problem, we do not want to reintroduce the loop parameter especially not as a dummy parameter.

chkr-private:
> Your last commits caused CI to fail. I normally do not review a PR before CI passes, but decided to make an exception, please in the future solve CI before asking for a review

Yes, I was aware of that but didn't had the time yesterday to continue to work on it.

I have just set the Pull Request to "Draft" to indicate, that it is not ready for review. Once it is, I'll remove the Draft state and request another review. Github should indicate that to the maintainers accordingly.

> I gave you a simple solution to the loop problem, we do not want to reintroduce the loop parameter especially not as a dummy parameter.

Ok. I think I understand now the idea of the async loop better and will work on a solution.

pymodbus/server/async_io.py
chkr-private: > This looks wrong, it creates a new loop, the loop must be created outside the class.

Fixed in latest commits.

pymodbus/server/async_io.py
chkr-private: > please use get_running_loop

Fixed in latest commits.

pymodbus/server/async_io.py
```diff
@@ -366,6 +367,7 @@ class ModbusTlsServer(ModbusTcpServer):
     def __init__(  # pylint: disable=too-many-arguments
         self,
         context,
+        loop=None, # not used
```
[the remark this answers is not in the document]
chkr-private: Fixed in latest commits.

pymodbus/server/async_io.py
```diff
@@ -411,6 +413,7 @@ def __init__(  # pylint: disable=too-many-arguments
         )
         super().__init__(
             context,
+            loop=loop,
```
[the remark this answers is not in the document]
chkr-private: Fixed in latest commits.

pymodbus/server/async_io.py
chkr-private: Fixed in latest commits.

pymodbus/server/async_io.py
chkr-private: Fixed in latest commits.

janiversen:
? this is still draft status ??

chkr-private:
> ? this is still draft status ??

Now it is ready for review. It took a little bit longer since I had to familiarize myself with asyncio.

For the current version I tested pymodbus.console/server (I'm gathering some ideas for (semi-)automatic tests here: https://github.com/chkr-private/pymodbus/wiki/Tests-for-pymodbus.console---pymodbus.server) and the CI passed as well.

API_changes.rst
```diff
@@ -6,6 +6,7 @@ PyModbus - API changes.
 Version 3.4.2
 -------------
 -ModbusSerialServer now accepts request_tracer=.
+-ModbusSerialServer does not use loop parameter anymore
```
janiversen, line 9: I think you have changed a lot more parameters, every API change needs to be documented.

Please add a new headline 
Version 3.5.0 (future)

since we do not accept API changes in 3.4.x versions.

pymodbus/repl/server/main.py
```diff
@@ -178,29 +178,29 @@ def run(
     data_block_settings = modbus_config.pop("data_block_settings", {})
     modbus_config = modbus_config.get(modbus_server, {})
     modbus_config = process_extra_args(extra_args, modbus_config)
```
janiversen, line 180: This is missing in API_CHANGES

pymodbus/repl/server/main.py
```diff
@@ -178,29 +178,29 @@ def run(
...
+            slave=modbus_slave_id,
+            single=False,
+            data_block_settings=data_block_settings,
+            **web_app_config,
+            **modbus_config,
+        )
+        await app.run_async(repl)
+        return app
+
+    app = loop.run_until_complete(_wrapper())
```
janiversen, line 199: This seems wrong, at least you are repeating it within the "if repl".

pymodbus/server/async_io.py
```diff
@@ -449,8 +449,6 @@ def __init__(
         :param framer: The framer strategy to use
         :param identity: An optional identify structure
         :param address: An optional (interface, port) to bind to.
```
janiversen, line 451: API doc.

janiversen (changes requested):
Look quite good, you miss a couple of API documentation and there are one line that I believe is wrong.

pymodbus/repl/server/main.py
chkr-private: As far as I understand, the wrapper function (and so app.run_async()) does not actually start the full server. It only does the initialization work (which, nevertheless, must be started in a running loop).

The _wrapper() just initialises the app (the same code for repl and non-repl case) and starts the web_app but does not run the modbus server itself.

The original [code](https://github.com/pymodbus-dev/pymodbus/blob/08529612b11966e4d16d085b7550d74737472b67/pymodbus/repl/server/main.py#L201C1-L205C27) did exactly the same 2-step process (first create the objects + start web server and as second step run the whole app):

```
    loop.run_until_complete(app.run_async(repl))


    if repl:
        loop.run_until_complete(run_repl(app))
    else:
        loop.run_forever()
```

The issue of the original code was, that the Server objects (and the start of the web app) must be initialized in a running loop in some cases. Otherwise, there would be a problem like this e.g. in the UDP case:

```
pymodbus.server run -s udp
#Traceback (most recent call last):

  File "/usr/bin/pymodbus.server", line 8, in <module>
    sys.exit(app())
             ^^^^^

  File "/usr/lib/python3.11/site-packages/pymodbus/repl/server/main.py", line 183, in run
    app = ReactiveServer.factory(
          ^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/lib/python3.11/site-packages/pymodbus/server/reactive/main.py", line 499, in factory
    server = server(
             ^^^^^^^

  File "/usr/lib/python3.11/site-packages/pymodbus/server/async_io.py", line 473, in __init__
    self.loop = asyncio.get_running_loop()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^

RuntimeError: no running event loop
```

janiversen: sound just about right.

pymodbus/repl/server/main.py
chkr-private: That specific change does not change the API, it rather corrects the parameter name when instantiating the various Modbus*Server objects. Neither the servers' __init__ methods nor the parameter when calling the console.server was changed.

API_changes.rst
chkr-private: Thank you for the last review. I have added all API changes as requested.

pymodbus/server/async_io.py
chkr-private: fixed in latest version

API_changes.rst
```diff
@@ -2,6 +2,14 @@
 PyModbus - API changes.
 =======================
 
+-------------
+Version 3.5.0 (future)
+-------------
+- Remove handler parameter from ModbusUdpServer
+- Remove handler from default config
```
janiversen, line 9:
```suggestion
- Remove handler from repl default config
```

API_changes.rst
```diff
@@ -2,6 +2,14 @@
...
+- Remove allow_reuse_port from default config
```
janiversen, line 10:
```suggestion
- Remove allow_reuse_port from repl default config
```

janiversen (approved):

janiversen:
Thanks.

PR #1715, opened 2023-07-31
Request tracer

wriswith:
Getting the request_tracer from kwargs instead of setting it to None in ModbusSerialServer. Works as expected.
Removed line from ModbusUdpServer that overwrote the request_tracer set on line 483 with None.

If there was a reason for disabling the request_tracer in the Serail and UDP server please let me know as I use it in my project.

### DIFF
--- a/API_changes.rst
+++ b/API_changes.rst
@@ -3,6 +3,12 @@ PyModbus - API changes.
 =======================
 
 -------------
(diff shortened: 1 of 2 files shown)


janiversen (approved):
LGTM, thanks.

The same should be done with the response_tracer

PR #1716, opened 2023-07-31
Fix custom message parsing in RTU framer

wlcrs:
As the regression reported in #1695 still isn't resolved in pyModbus 3.4.1, I've been tinkering with the code to try to reproduce the issue.

I noticed that `ModbusRtuFramer` caches the function codes registered in the decoder lookup table. As the framer is already initialized before a client can call the `register()` function to register custom Modbus message types, this prevents those custom Modbus messages from being processed correctly.

This PR uses `.keys()` instead of `set()` which prevents `function_codes` from becoming outdated if a custom message class is registered.

Open question: if there a valid case for the decoder being None? Otherwise we can drop the fallback to `{}`?

### DIFF
--- a/pymodbus/framer/rtu_framer.py
+++ b/pymodbus/framer/rtu_framer.py
@@ -62,7 +62,7 @@ class ModbusRtuFramer(ModbusFramer):
         self._hsize = 0x01
         self._end = b"\x0d\x0a"
         self._min_frame_size = 4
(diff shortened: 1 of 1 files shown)


janiversen:
May I politely correct you, the issue raised in #1695 is solved ! seems you are having another issue in addition to #1695.

janiversen (approved):
LGTM.

janiversen:
When you call register() it should update the local copy, but your solution is more robust.

PR #1720, opened 2023-08-03
Impoved the example "server_updating.py"

aloishockenschlohe:
Added some comments to explain the code a bit better.
Changed the update routine to have a more expressive example.

### DIFF
--- a/examples/server_updating.py
+++ b/examples/server_updating.py
@@ -105,7 +105,6 @@ def setup_updating_server(cmdline=None):
 async def run_updating_server(args):
     """Start updating_task concurrently with the current task"""
     task = asyncio.create_task(updating_task(args.context))
-    """start the server"""
     await run_async_server(args)  # start the server
     task.cancel()
 
--- a/examples/server_updating.py
+++ b/examples/server_updating.py
@@ -1,7 +1,7 @@
-#!/usr/bin/env python3
-"""Pymodbus asynchronous Server Example.
+"""Pymodbus asynchronous Server with updating subroutine Example.
 
-An example of a multi threaded asynchronous server.
+An example of an asynchronous server with multiple threads and
+a subroutine that runs continuously alongside the server and updates values.
 
 usage: server_async.py [-h] [--comm {tcp,udp,serial,tls}]
                        [--framer {ascii,binary,rtu,socket,tls}]
@@ -16,6 +16,7 @@ options:
   --comm {tcp,udp,serial,tls}
                         "serial", "tcp", "udp" or "tls"
   --framer {ascii,binary,rtu,socket,tls}
+
                         "ascii", "binary", "rtu", "socket" or "tls"
   --log {critical,error,warning,info,debug}
                         "critical", "error", "warning", "info" or "debug"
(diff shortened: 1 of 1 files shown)


examples/server_updating.py
```diff
@@ -1,7 +1,7 @@
-#!/usr/bin/env python3
```
janiversen, line 1: This line is needed to run the file as a script.

examples/server_updating.py
```diff
@@ -1,7 +1,7 @@
...
-"""Pymodbus asynchronous Server Example.
+"""Pymodbus asynchronous Server with updating subroutine Example.
 
-An example of a multi threaded asynchronous server.
+An example of an asynchronous server with multiple threads and
```
janiversen, line 3: Actually it does not have threads, the correct wording is tasks. Threads can be done in both sync and async, but task are solely async.

examples/server_updating.py
```diff
@@ -42,20 +43,38 @@
...
 
 async def updating_task(context):
-    """Run every so often and update live values of the context.
+    """
+    This subroutine will run continously beside the server and
+    increment each two seconds some values (via asyncio.create_task
+    in run_updating_server).
 
-    It should be noted that there is a race condition for the update.
+    It should be noted that getValues and setValues are not thread safe.
```
janiversen, line 51: Not only are they not threadsafe, they are not safe from any concurrent access

examples/server_updating.py
```diff
@@ -42,20 +43,38 @@
...
-    _logger.debug("updating the context")
+
     fc_as_hex = 3
     slave_id = 0x00
-    address = 0x10
-    values = context[slave_id].getValues(fc_as_hex, address, count=5)
-    values = [v + 1 for v in values]  # increment by 1.
-    txt = f"new values: {values!s}"
-    _logger.debug(txt)
+    address = 7
```
janiversen, line 56: You are changing the addresses I hope you tested with our standard server. We use the same addresses for all examples.

examples/server_updating.py
```diff
@@ -76,8 +95,8 @@ def setup_updating_server(cmdline=None):
 
 async def run_updating_server(args):
     """Start updater task and async server."""
-    asyncio.create_task(updating_task(args.context))  # noqa: RUF006
-    await run_async_server(args)
+    asyncio.create_task(updating_task(args.context)) # start updating_task concurrently with the current task
```
janiversen, line 98: Please make comments on a seperate line, otherwise it gets too long.

janiversen:
Please also solve the CI linter error.

examples/server_updating.py
janiversen: Till open

examples/server_updating.py
```diff
@@ -42,20 +44,40 @@
...
-    _logger.debug("updating the context")
+
     fc_as_hex = 3
     slave_id = 0x00
     address = 0x10
-    values = context[slave_id].getValues(fc_as_hex, address, count=5)
-    values = [v + 1 for v in values]  # increment by 1.
-    txt = f"new values: {values!s}"
-    _logger.debug(txt)
+    c = 6
```
janiversen, line 59: Please use variable names that mean something e.g. count (just like the other variables)

janiversen (changes requested):
You still have a spelling problem see CI error

janiversen (approved):
LGTM, thanks.

PR #1725, opened 2023-08-08
Fix faulty not check

Booplicate:
### Issue:
If you check how you're handling writing to multiple registers, you notice you're using [explicit `is None` check](https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/register_write_message.py#L171), which allows passing "falsy" values.

However for coils, you're doing [`not` check](https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/bit_write_message.py#L188), which means it's impossible to write `0` as it will get converted to an empty `list`.

It's not a big deal and it's a good practice to explicitly wrap values into a list, but since you do support it for input registers, can make sense to support for coils as well.

### Changelog:
- replaces `not` check with explicit `is None`
- added a test case for this

### Note:
I'm not too familiar with `pytest`, but I hope I added a valid test case for this.

### DIFF
--- a/pymodbus/bit_write_message.py
+++ b/pymodbus/bit_write_message.py
@@ -185,7 +185,7 @@ class WriteMultipleCoilsRequest(ModbusRequest):
         """
         ModbusRequest.__init__(self, slave=slave, **kwargs)
         self.address = address
-        if not values:
+        if values is None:
             values = []
         elif not hasattr(values, "__iter__"):
             values = [values]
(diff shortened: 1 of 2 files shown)


janiversen (approved):
LGTM, thanks.

Just wondering if the opposite solution would have been better (bring coil inline).

Booplicate:
> Just wondering if the opposite solution would have been better (bring coil inline).

Up to you, wasn't sure about that since it'd be a breaking change.

janiversen:
good point, lets keep it the way you made it.

Issue #1728, opened 2023-08-12
Retry logic fails when retries parameter is set to 1


sefakeles:
### Versions

- Python: 3.11
- OS: Linux
- Pymodbus: 3.4.1

### Pymodbus Specific

- Client: tcp - async

### Description

The `async_execute` method of `ModbusBaseClient` includes a retry logic, which does not work properly if the value of `retries` parameter is 1. In this case, even if the response is received successfully, the method raises `ModbusIOException`.

It can be assumed that this issue is related to #1702. Because, setting the value of `retries` parameter to 1 can be used as a workaround for #1702.

### Code and Logs

```python
import asyncio
import logging

from pymodbus import pymodbus_apply_logging_config
from pymodbus.client import AsyncModbusTcpClient
from pymodbus.exceptions import ModbusException

logger = logging.getLogger()


async def main() -> None:
    pymodbus_apply_logging_config()
    try:
        async with AsyncModbusTcpClient("10.10.2.23", retries=1) as client:
            response = await client.read_holding_registers(38, 2, slave=247)
            if not response.isError():
                logger.info(response.registers)
    except ModbusException as err:
        logger.error(err)


if __name__ == "__main__":
    asyncio.run(main())
```

```shell
2023-08-13 02:04:38,188 DEBUG logging:102 Connecting to 10.10.2.23:502.
2023-08-13 02:04:38,188 DEBUG logging:102 Connecting comm
2023-08-13 02:04:40,117 DEBUG logging:102 Connected to comm
2023-08-13 02:04:40,118 DEBUG logging:102 callback_connected called
2023-08-13 02:04:40,119 DEBUG logging:102 send: 0x0 0x1 0x0 0x0 0x0 0x6 0xf7 0x3 0x0 0x26 0x0 0x2
2023-08-13 02:04:40,119 DEBUG logging:102 Adding transaction 1
2023-08-13 02:04:40,487 DEBUG logging:102 recv: 0x0 0x1 0x0 0x0 0x0 0x7 0xf7 0x3 0x4 0xc1 0xe 0xa8 0x0 addr=None
2023-08-13 02:04:40,488 DEBUG logging:102 Processing: 0x0 0x1 0x0 0x0 0x0 0x7 0xf7 0x3 0x4 0xc1 0xe 0xa8 0x0
2023-08-13 02:04:40,488 DEBUG logging:102 Factory Response[ReadHoldingRegistersResponse': 3]
[...]

PR #1733, opened 2023-08-13
Fix #1702 and #1728

### LINKED ISSUE #1702: Pymodbus 3.4.0, bug in async_execute
Python: 3.8
Pymodbus: 3.4.0
used on: UDP Client (for our example connection to nowhere, or not existed address)

Bug:
https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/client/base.py#L193
Future is created only ONCE

code tries to use retries, first time it is ok, when more times then `wait_for` using CANCELLED future!

FIX:
move `req = self._build_response(request.transaction_id)` to wait_for

```py3
async def async_execute(self, request=None):
        """Execute requests asynchronously."""
        request.transaction_id = self.transaction.getNextTID()
        packet = self.framer.buildPacket(request)
        self.transport_send(packet)        
        if self.params.broadcast_enable and not request.slave_id:
            resp = b"Broadcast write sent - no response expected"
        else:
            count = 0
            while count < self.params.retries:
                count += 1
                try:
                    req = self._build_response(request.transaction_id)
                    resp = await asyncio.wait_for(
                        req, timeout=self.comm_params.timeout_connect
                    )
                    break
                    pass
            if count == self.params.retries:
                self.close(reconnect=True)
                raise ModbusIOException(
                    f"ERROR: No response received after {self.params.retries} retries"
                )
        return resp
```

sefakeles:


janiversen (approved):
LGTM, thanks.

sefakeles:
You are welcome.

PR #1737, opened 2023-08-15
Broadcast with Handle Local Echo

lsgunth:
When using a broadcast transaction with handle_local_echo=True the broadcast skips reading a the local echo and leaves
the echo in the receive buffer. This results in a "Cleanup recv buffer before send" warning on the next transaction.

Fix this by ensuring the local echo is read for broadcast transactions.

Also add a test for this behaviour.


janiversen:
LGTM, thanks

PR #1738, opened 2023-08-16
Handle partial message in ReadDeviceInformationResponse

lsgunth:
When a partial message is received for a ReadDeviceInformationResponse, an exception occurs when calculating the frame size:
 
The caller is expecting an IndexError if there isn't enough data in the packet. So to fix this, catch the struct error and raise an IndexError.

### DIFF
--- a/pymodbus/mei_message.py
+++ b/pymodbus/mei_message.py
@@ -125,11 +125,14 @@ class ReadDeviceInformationResponse(ModbusResponse):
         size = 8  # skip the header information
         count = int(buffer[7])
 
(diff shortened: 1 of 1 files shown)


janiversen (changes requested):
The change looks good, please solve the linter problem in CI.

lsgunth:
Sorry, I was rushing to get that done last night. I had fixed the linter error but forgot to commit the fix. Should pass now.

Thanks!

janiversen:
Thanks, looking forward to see other pull requests from you.

PR #1743, opened 2023-08-21
Use enums for constants

### LINKED ISSUE #1741: Should `constants.Endian` be an enum?
### Versions

- Python: 3.11
- OS: Arch
- Pymodbus: 3.4.1

### Pymodbus Specific

- Server: tcp/rtu/ascii - all
- Client: tcp/rtu/ascii - all

### Description

I think it'd be more convenient if `Endian` was an actual enum, could help with type safety. For compatibility it could be made "str enum" and be fully compatible with literals.

All target python versions have support for enums.

If desired, I could work on this.

Booplicate:
Closes #1741, went a bit further and converted all static classes to enums. This should provide type safety while maintaining compatibility with raw ints and strings.

### Changelog:
- `ModbusStatus` > inherits from `int` and `enum.Enum`
- `Endian` > inherits from `str` and `enum.Enum`
- `ModbusPlusOperation` > inherits from `int` and `enum.Enum`
- `DeviceInformation` > inherits from `int` and `enum.Enum`
- `MoreData` > inherits from `int` and `enum.Enum`

### DIFF
--- a/pymodbus/constants.py
+++ b/pymodbus/constants.py
@@ -11,39 +11,39 @@ INTERNAL_ERROR = "Pymodbus internal error"
 class ModbusStatus(int, enum.Enum):
     """These represent various status codes in the modbus protocol.
 
-    .. attribute:: Waiting
+    .. attribute:: WAITING
 
        This indicates that a modbus device is currently
        waiting for a given request to finish some running task.
 
-    .. attribute:: Ready
+    .. attribute:: READY
 
        This indicates that a modbus device is currently
        free to perform the next request task.
 
-    .. attribute:: On
+    .. attribute:: ON
 
        This indicates that the given modbus entity is on
 
-    .. attribute:: Off
+    .. attribute:: OFF
 
        This indicates that the given modbus entity is off
 
-    .. attribute:: SlaveOn
+    .. attribute:: SLAVE_ON
 
        This indicates that the given modbus slave is running
 
-    .. attribute:: SlaveOff
+    .. attribute:: SLAVE_OFF
 
        This indicates that the given modbus slave is not running
     """
 
-    Waiting = 0xFFFF
-    Ready = 0x0000
-    On = 0xFF00
-    Off = 0x0000
-    SlaveOn = 0xFF
-    SlaveOff = 0x00
+    WAITING = 0xFFFF
+    READY = 0x0000
+    ON = 0xFF00
+    OFF = 0x0000
+    SLAVE_ON = 0xFF
+    SLAVE_OFF = 0x00
 
     def __str__(self):
        return str(int(self))
@@ -52,16 +52,16 @@ class ModbusStatus(int, enum.Enum):
 class Endian(str, enum.Enum):
     """An enumeration representing the various byte endianness.
 
-    .. attribute:: Auto
+    .. attribute:: AUTO
 
        This indicates that the byte order is chosen by the
        current native environment.
 
-    .. attribute:: Big
+    .. attribute:: BIG
 
        This indicates that the bytes are in big endian format
 
-    .. attribute:: Little
+    .. attribute:: LITTLE
 
        This indicates that the bytes are in little endian format
 
@@ -69,9 +69,9 @@ class Endian(str, enum.Enum):
        python struct module for my convenience.
     """
 
-    Auto = "@"
-    Big = ">"
-    Little = "<"
+    AUTO = "@"
+    BIG = ">"
+    LITTLE = "<"
 
     def __str__(self):
        return str.__str__(self)
@@ -80,19 +80,19 @@ class Endian(str, enum.Enum):
 class ModbusPlusOperation(int, enum.Enum):
     """Represents the type of modbus plus request.
 
-    .. attribute:: GetStatistics
+    .. attribute:: GET_STATISTICS
 
        Operation requesting that the current modbus plus statistics
        be returned in the response.
 
-    .. attribute:: ClearStatistics
+    .. attribute:: CLEAR_STATISTICS
 
        Operation requesting that the current modbus plus statistics
        be cleared and not returned in the response.
     """
 
-    GetStatistics = 0x0003
-    ClearStatistics = 0x0004
+    GET_STATISTICS = 0x0003
+    CLEAR_STATISTICS = 0x0004
 
     def __str__(self):
        return str(int(self))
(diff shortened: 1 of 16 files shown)


pymodbus/constants.py
```diff
@@ -44,12 +45,8 @@ class ModbusStatus:  # pylint: disable=too-few-public-methods
     SlaveOn = 0xFF
     SlaveOff = 0x00
 
-    def __init__(self):
-        """Prohibit objects."""
-        raise RuntimeError(INTERNAL_ERROR)
 
-
-class Endian:  # pylint: disable=too-few-public-methods
```
janiversen, line 52: This should be replaced by the system constant.

test/sub_messages/test_mei_messages.py
```diff
@@ -93,7 +93,6 @@ def test_read_device_information_encode(self):
         )
         result = handle.encode()
         assert result == message
-        assert str(handle) == "ReadDeviceInformationResponse(1)"
```
janiversen, line 96: Please do not remove a test to make your code work, if the return now is different then its most likely because something is broken.

The very minimum is to adapt the test.

test/sub_messages/test_mei_messages.py
```diff
@@ -133,7 +132,6 @@ def test_read_device_information_encode_long(self):
...
```
janiversen, line 136: Same here.

janiversen (changes requested):
You need to resolve the linter errors. Changing to upper-case seems most correct, but there were no lint problems before, so maybe it is due to how you changed the classes.

janiversen:
> Upd.
> 
> 
> 
> I updated the tests, at first I wanted to replace 
> 
> ```py
> 
> assert str(handle) == "ReadDeviceInformationResponse(1)"
> 
> ```
> 
> with
> 
> ```py
> 
> assert handle.read_code == DeviceInformation.Basic
> 
> ```
> 
> but there's already a test for that, and I don't see a reason to test value of `read_code` on `encode`. I just removed them.
> 
> 
> 
> Lint fails and I agree constants should be in UPPER_CASE, but for compatibility we have to leave them as is.

The change you wanted to make, is quite a different test, the test is checking the string representation of handle, not an individual component.

pymodbus/constants.py
Booplicate: Not sure I understood, do you mean `sys.byteorder`?

test/sub_messages/test_mei_messages.py
Booplicate: Sorry, I took a look and it didn't make much sense to me to test string representation, as far as I could tell the only thing that could affect it was `read_code`.

I will undo this and adjust string representation of the enums instead.

Booplicate:
@janiversen just to note, at least `Endian` looks like part of the public API. One way we could handle this is by suppressing the warnings and adding aliases for compatibility.

However if you're fine with breaking things, then I will just rename as you proposed.

janiversen:
> @janiversen just to note, at least `Endian` looks like part of the public API. One way we could handle this is by suppressing the warnings and adding aliases for compatibility.
> 
> However if you're fine with breaking things, then I will just rename as you proposed.

It is the default in many payload calls, and therefore it will not change anything if using the "official" constant instead.

If you rename things that are real part of the API (meaning people will have to change their app) then add a line in API_CHANGES.rst.

Yes I am happy with renaming, we have had changes like this in every 3.x version but never in 3.x.y versions.

Booplicate:
@janiversen, could you please force re-run checks, I believe there was an issue with the runner.

pymodbus/constants.py
```diff
@@ -3,155 +3,157 @@
...
-    SlaveOn = 0xFF
-    SlaveOff = 0x00
+    WAITING = 0xFFFF
+    READY = 0x0000
+    ON = 0xFF00
+    OFF = 0x0000
+    SLAVE_ON = 0xFF
+    SLAVE_OFF = 0x00
 
-    def __init__(self):
```
janiversen, line 47: Why should we permit objects ? I do not see the need in contrary.

pymodbus/constants.py
```diff
@@ -3,155 +3,157 @@
...
     """
 
-    Auto = "@"
-    Big = ">"
-    Little = "<"
+    AUTO = "@"
+    BIG = ">"
+    LITTLE = "<"
 
-    def __init__(self):
```
janiversen, line 76: We should not permit to instantiate any constant class.

janiversen (changes requested):
The only bigger item is the remove of __init__.

Please add a short description in API_CHANGES.rst (section 3.5) that constant now uses enums and are capital letters.

janiversen:
A question, why do you inherit from int, str ? I do not see anything in the class that makes use of it...you do not make a special int.

pymodbus/constants.py
Booplicate: Enum members are singletons by default, we don't need any special logic to guard them, Python handles that for us.

janiversen: It does not prohibit that I can assign the class to a variable and make changes, with unexpected results.

Booplicate:
It's the way to create [derived enums](https://docs.python.org/3.10/library/enum.html#derived-enumerations). By default enum members are just unique constants with *some* value. However, in our case the values have meaning (like modbus code), so we want to control what the values are, and we want to use the enum members as the values. Thus, for hex constants I used `int`, for string, I used `str`.
For example, in case of `ModbusStatus`, we want
```py
assert ModbusStatus.WAITING == 0xFFFF
```
Without the inheritance, we'd have to do
```py
assert ModbusStatus.WAITING.value == 0xFFFF
```
which is more error-prone and less convenient to use.

pymodbus/constants.py
Booplicate: Not in the case of enums, they are using metaclasses underneath. Say, if somebody does `ModbusStatus(0xFFFF)`, they will get `ModbusStatus.WAITING`.

janiversen:
Good point !

Apart from when you look at comm_types in transport, that class inherits "class comm_types(Enum)" and the values are addressed directly as comm_types.TCP

Why did you decide to inherit Enum.enum seems to complicate matters. Or have I overlooked something ?

janiversen:
A short example with you need to disable __init__:
```
Python 3.11.4 (main, Jul 25 2023, 17:36:13) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from constants import ModbusStatus
>>> test = ModbusStatus(ModbusStatus.WAITING)
>>> test.WAITING = 0
>>> print(test.WAITING)
0
>>> print(ModbusStatus.WAITING)
65535
>>> 
```
Having different values for the same constant are likely to cause problems (and have done so).

Booplicate:
I will see if it's possible to change behaviour of `__init__`, but I doubt so.

I believe that LS/typechecker/IDE should prevent that kind of errors:
image

because you can override even if you don't init:
```py
test = ModbusStatus.WAITING
test.WAITING = 0
assert test.WAITING != ModbusStatus.WAITING
```

janiversen:
IDE is no good to us, we depend on CI to be the final judge.

I simply do not like to remove this kind of protection unless it is absolutely nessecary , it will not provide 100% protection, just like constants can be assigned to, but it's better than nothing. I made the example to show that your argument about singleton doesn't not hold in this case.

janiversen:
We are getting close to release time, so if you want this in please solve the outstanding issue, and request a new review.

janiversen:
Did you try to call super().__init__()

janiversen:
the __init__() is only called if you instanciate the class, so something sounds wrong.

Booplicate:
Enums use metaclasses, so logic from ordinary classes cannot be applied, `__init__` is being called for each enum member. I tried  playing around with removing/changing `__init__` and `__call__` (they use both!), but nothing seems to result in the behaviour you want (most result in exceptions of sorts).

If you try:
```py
class Example(enum.Enum):
    def __init__(self, value):
        raise Exception("need __init__")
    A = "A"
    B = "B"
```
You will get `Exception: need __init__` right away.

janiversen:
Seems you are right.

janiversen (approved):
LGTM, thanks

Looking forward to your next PR.

PR #1744, opened 2023-08-22
Local Echo Broadcast with Async Clients

lsgunth:
Using broadcast messages with local echo almost works with asynchronous clients but it is racy.

If the echo for a broadcast comes in after the next transaction has been sent, then the sent_buffer gets overridden and the echo of the broadcast is forgotten. This corrupts the data stream.

This PR has a couple commits to make it more robust: for one allow the echo to be split along multiple packets, then ensure a broadcast waits for the echo before returning from async_execute.

The last two commits add tests for both these issues.

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -391,7 +391,7 @@ class ModbusProtocol(asyncio.BaseProtocol):
         """
         Log.debug("send: {}", data, ":hex")
         if self.comm_params.handle_local_echo:
-            self.sent_buffer = data
+            self.sent_buffer += data
         if self.comm_params.comm_type == CommType.UDP:
             if addr:
                 self.transport.sendto(data, addr=addr)  # type: ignore[attr-defined]
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -312,10 +312,23 @@ class ModbusProtocol(asyncio.BaseProtocol):
     def datagram_received(self, data: bytes, addr: tuple):
         """Receive datagram (UDP connections)."""
         if self.comm_params.handle_local_echo and self.sent_buffer:
-            Log.debug("recv skipping (local_echo): {} addr={}", data, ":hex", addr)
-            if self.sent_buffer in data:
-                data, self.sent_buffer = data.replace(self.sent_buffer, b"", 1), b""
+            if data.startswith(self.sent_buffer):
+                Log.debug(
+                    "recv skipping (local_echo): {} addr={}",
+                    self.sent_buffer,
+                    ":hex",
+                    addr,
+                )
+                data = data[len(self.sent_buffer) :]
+                self.sent_buffer = b""
+            elif self.sent_buffer.startswith(data):
+                Log.debug(
+                    "recv skipping (partial local_echo): {} addr={}", data, ":hex", addr
+                )
+                self.sent_buffer = self.sent_buffer[len(data) :]
+                return
             else:
+                Log.debug("did not receive local echo: {} addr={}", data, ":hex", addr)
                 self.sent_buffer = b""
             if not data:
                 return
--- a/test/sub_transport/test_basic.py
+++ b/test/sub_transport/test_basic.py
@@ -155,6 +155,12 @@ class TestBasicModbusProtocol:
         client.datagram_received(test_data, ("127.0.0.1", 502))
         assert client.recv_buffer == test_data + test_data
         assert not client.sent_buffer
+        client.recv_buffer = b""
+        client.transport_send(b"partial")
+        client.datagram_received(b"par", ("127.0.0.1", 502))
+        client.datagram_received(b"tialresponse", ("127.0.0.1", 502))
+        assert client.recv_buffer == b"response"
+        assert not client.sent_buffer
 
     async def test_transport_close(self, server, dummy_protocol):
         """Test transport_close()."""
--- a/test/sub_transport/test_basic.py
+++ b/test/sub_transport/test_basic.py
@@ -162,6 +162,31 @@ class TestBasicModbusProtocol:
         assert client.recv_buffer == b"response"
         assert not client.sent_buffer
 
+    async def test_broadcast_local_echo(self, client):
+        """Test transport_send() with broadcast and echo packets"""
+        client.comm_params.handle_local_echo = True
+        client.transport = mock.Mock()
+        client.recv_buffer = b""
+        client.transport_send(b"broadcast")
+        client.transport_send(b"message")
+        client.datagram_received(b"broadcast", ("127.0.0.1", 502))
+        client.datagram_received(b"messageresponse", ("127.0.0.1", 502))
+        assert client.recv_buffer == b"response"
+        assert not client.sent_buffer
+        client.recv_buffer = b""
+        client.transport_send(b"broadcast")
+        client.transport_send(b"message")
+        client.datagram_received(b"broadcastmessageresponse", ("127.0.0.1", 502))
+        assert client.recv_buffer == b"response"
+        assert not client.sent_buffer
+        client.recv_buffer = b""
+        client.transport_send(b"broadcast")
+        client.transport_send(b"message")
+        client.datagram_received(b"broadcastmessa", ("127.0.0.1", 502))
+        client.datagram_received(b"geresponse", ("127.0.0.1", 502))
+        assert client.recv_buffer == b"response"
+        assert not client.sent_buffer
+
     async def test_transport_close(self, server, dummy_protocol):
         """Test transport_close()."""
         dummy_protocol.abort = mock.MagicMock()


pymodbus/client/base.py
```diff
@@ -191,6 +191,14 @@ async def async_execute(self, request=None):
         packet = self.framer.buildPacket(request)
         self.transport_send(packet)
         if self.params.broadcast_enable and not request.slave_id:
+            if self.comm_params.handle_local_echo:
```
janiversen, line 194: This will not work in all cases. Local echo is something that might happen, but are not guaranteed to happen in all cases.

You mix transport and client, which secures your PR will not work for the server.

pymodbus/transport/transport.py
```diff
@@ -310,12 +311,26 @@ def data_received(self, data: bytes):
         self.datagram_received(data, None)
 
     def datagram_received(self, data: bytes, addr: tuple):
-        """Receive datagram (UDP connections)."""
+        """Receive datagram"""
```
janiversen, line 314: Why change this ?

pymodbus/transport/transport.py
```diff
@@ -310,12 +311,26 @@ def data_received(self, data: bytes):
...
+                Log.debug(
+                    "recv skipping (local_echo): {} addr={}",
+                    self.sent_buffer,
+                    ":hex",
+                    addr,
+                )
+                data = data[len(self.sent_buffer) :]
+                self.sent_buffer = b""
+                self.wait_for_local_echo.set_result(True)
+            elif self.sent_buffer.startswith(data):
```
janiversen, line 326: If that happens then it is not a local echo.

test/sub_client/test_client.py
```diff
@@ -360,6 +360,31 @@ async def test_client_protocol_handler():
     assert result == reply
 
 
+async def test_client_broadcast_echo():
+    """Test the client protocol handles broadcasts with local echos correctly"""
```
janiversen, line 364: Local echo, is also used in the server.

janiversen (changes requested):
Seems you have missed out that local echo must work for both client and server, since transport is used in both.

test/sub_transport/test_basic.py
```diff
@@ -155,6 +155,12 @@ async def test_handle_local_echo(self, client):
         client.datagram_received(test_data, ("127.0.0.1", 502))
         assert client.recv_buffer == test_data + test_data
         assert not client.sent_buffer
+        client.recv_buffer = b""
```
janiversen, line 158: This is a good test, and show that transport do have a problem !

If you change the PR to avoid the future, but only have the changes in transport I am happy to merge it.

janiversen (changes requested):
Nice test!

pymodbus/client/base.py
lsgunth: If the local echo doesn't happen then the future will be completed anyway and this will work.

There is a test in the suite for when a local echo does not occur.

pymodbus/transport/transport.py
lsgunth: Per the commit message, it was confusing and incorrect. This path is used for serial clients.

pymodbus/transport/transport.py
lsgunth: How so? So a partial read of a local echo is not a local echo??

This path is used by the serial port and it is definitely possible to have a partial read of the data.

lsgunth:
> I do not understand your racing condition...there are no "next transaction" before a response is received from the first transaction.

When sending a broadcast that will result in an echo, the current code sets the send_buffer to what was previously sent. The broadcast packet does not wait for any response and returns immediately after the packet is sent. The user code continues while the packet is transmitting on the line, often before the local echo is fully received. If the user code sends another packet immediately after the broadcast packet, it will overwrite send_buffer with the new packet before the echo was received. When the echo of the broadcast packet is then received it is not seen as an echo because send_buffer contains the contents of the next packet. This results in communication failures seeing there's unexpected data on the bus. 

> "local echo" does not involve any serial/network communication but are purely at device driver level in the local computer, so there are no delay in receiving the data.

I'm working with hardware that uses an MAX13487E with auto direction control to a half-duplex RS485 bus.  The auto direction control is necessary seeing the microprocessor cannot control the direction quick enough to reliably receive the other devices' responses. Auto direction control automatically controls whether the transmitter is enabled but it doesn't do anything for the receiver, and thus everything sent on the bus will be received by the hardware. In this case then, the local echo is at the hardware level, cannot be disabled and is delayed for as long as it takes to send the buffer and receive it back through the kernel. Assuming that a local echo is always instant is wrong.
 
> Are you basing this PR on a real life problem ? if so please open an issue, with the debug logs,

Yes, this is a real life problem. I'm not able to capture the issue with debug logs enabled seeing it changes the timing: it adds enough delay that the local echo is received before the next packet is sent.

Sorry, I should have included more context in the PR.

test/sub_transport/test_basic.py
lsgunth: The future is the main fix for my bug. I need it or another method to accomplish the same result.

There's another way I considered solving the issue: the send_buffer could be appended instead of reset when a new packet is sent. 

I chose to do it this way seeing I felt it would be more robust for every transaction to return when it is completed including receiving the local echo. If we append to send_buffer then the broadcast will still return early and the next transaction will continue before the local echo is received. This seemed more risky to me.

pymodbus/transport/transport.py
janiversen: No that path is used for datagrams which are udp, but the data method which handles all other types calls this, in order not to duplicate code.

Please revert this change.

pymodbus/client/base.py
janiversen: Yes but the future should go away, echo is handled solely in the transport class! Read the definition of the transport class.

pymodbus/transport/transport.py
janiversen: Of course, but it looks suspect to me, maybe because it is all mixed with the future.

test/sub_client/test_client.py
janiversen: Can a server not use a usb adapter for serial communication ?
so local echo can happen there too.

janiversen:
Broadcast do not exist in modbus, especially not in rs485 communication, you can set slave=0 meaning address all devices on the line.

The client still looks for a response and waits until timeout, at least when you use the mixin methods.

If you send multiple requests on the serial line without giving the devices time to respond (go from listen mode to write mode), and that breaks the modbus protocol.

Part of the PR makes sense to me, especially since the added test shows a clear bug, but the future is not acceptable nor that transport logic goes into client base. The transport class is to be seen as a closed box with a few methods and a few callbacks, and it handles everything relevant to to the transport layer.

janiversen:
Just made a test with client_async_calls, change SLAVE=0x00 (broadcast as you call it) and ran it on a serial line.

All requests waited for a response....could the problem be, you forgotten to await the client call: "await client.read_.....", that would allow your code to continue, but break a lot of things in pymodbus.

lsgunth:
> Broadcast do not exist in modbus, especially not in rs485 communication, you can set slave=0 meaning address all devices on the line.

Broadcasts do exist in serial modbus:

> In broadcast mode, the master can send a request to all slaves.
> No response is returned to broadcast requests sent by the master. The broadcast requests are necessarily writing commands. All devices must accept the broadcast for writing function. The address 0 is reserved to identify a broadcast exchange.

(From [MODBUS over serial line specification and implementation guide V1.02 ](https://www.modbus.org/docs/Modbus_over_serial_line_V1_02.pdf) -- page 7. Broadcasts are also referenced in other specifications.)

In fact pymodbus supports this behaviour -- it is just optional. You must specify broadcast_enable=True when creating the client. This works, and does not wait for the response in the current pymodbus code. You can see this in the code I'm changing in async_execute(), it simply returns the text "Broadcast write sent - no response expected" and does not wait for any response.  This implementation just has this bug I'm trying to solve when it is combined with handle_local_echo=True.

I suspect the test you tried did not set broadcaste_enable in the client. Then you would see what you are seeing. I can tell you that I am definitely await-ing the calls correctly.

Broadcast support on the server side is different. In order for a server to support broadcasts it has to handle requests with a slave ID of zero and not send the response for such requests. There is a similar broadcast_enable flag in the server context and is implemented [here](https://github.com/pymodbus-dev/pymodbus/blob/cbbdf0f4b2f0102c4c1b98419cca3ac8e2980b76/pymodbus/server/async_io.py#L193).  The server side does not have an equivalent of the problem I'm trying to solve with handle_local_echo seeing it doesn't actually send anything for broadcast packets.

If you prefer the approach of appending to send_buffer, I can try that. However I won't get to it until next week.

janiversen:
you are right, and I stand corrected....however your idea about the server is a bit off, remember a server can sit on a RS485 with many other devices so it is quite a lot more complicated.

Anyhow, my main point is, Keep transport clean and do not spread transport issues into the upper levels. We have worked hard the last few releases to clean especially the client code !

When you have it ready, then add me as reviewer.

janiversen:
Just passed over transport, the solution is quite simple. 

All you need to do is in transport_send check if send_buffer is empty and wait if not.

lsgunth:
> All you need to do is in transport_send check if send_buffer is empty and wait if not.

Ok, that sounds reasonable. Just a quick question:

As it is now, transport_send() does not have any knowledge of whether it is a broadcast request or not. 

If it doesn't know it's a broadcast then we are adding an additional wait on every request which is unnecessary and reduces efficiency.

Would you be ok with passing a broadcast flag to this function which adds the wait in when handle_local_echo is also true?

janiversen:
It should not know about broadcast the is message layer.

It only needs to wait if send_buffer contains something, for non broadcast that will not be the case so no problem

lsgunth:
> It only needs to wait if send_buffer contains something, for non broadcast that will not be the case so no problem

send_buffer will be set for *all* messages when handle_local_echo is true, not just broadcast messages. In 99% of the cases it does not need to wait for the local echo seeing the code will eventually be waiting for a response which includes that. The wait is only required when broadcast packets are sent *and* handle_local_echo is true. send_buffer will still be set for non-broadcast cases. 

In my opinion adding a wait for all packets is just adding inefficiencies to common use cases when it's only really needed in the niche case. So I think it's worthwhile passing a broadcast flag to transport_send() or finding another way.

janiversen:
you forget in all other cases, there are no extra send until data is received which will clear the send_buffer so there are no extra wait.

But it's your PR so feel free to do what you want, I will review what you submit. To me it is important to keep the layers separated, and broadcast relate to message handling not transport,

janiversen:
We are getting close to release time, so if you want this in please solve the outstanding issue, and request a new review.

lsgunth:
I attempted solving the problem in the way that we discussed, however, it was not possible. transport_send() is not an async function so awaiting on a future is not possible in that function. transport_send() also called in places that were not async functions so it was not possible to easily convert it to one. (The classic [function colouring problem](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/).)

Instead, I've pushed a fix using the other solution I alluded to: append to `sent_buffer`  in transport_send. This solves the problem by not dropping the expected echo of the broadcast packet. The difference between this solution and the previous one is that sending a broadcast packet will not wait for the echo for that packet. It will instead be processed in due course, perhaps prepended to the receive buffer in the next packet. As best I can tell, there isn't an issue with this.

janiversen (approved):
LGTM, thanks.

Looking forward to see more PRs from you, this one was not an easy one and we are always looking for a helping hand.

lsgunth:
Thanks!

PR #1750, opened 2023-08-28
Fix retries in ModbusBaseClient.async_execute()

lsgunth:
When using the asynchronous client and no response is received when a request is sent, the code sort of retries, but doesn't actually send the packets to retry.  This PR fixes that issue, fixes the test for async_execute() and adds two additional tests to verify the behaviour of retries.

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -189,25 +189,27 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
         """Execute requests asynchronously."""
         request.transaction_id = self.transaction.getNextTID()
         packet = self.framer.buildPacket(request)
-        self.transport_send(packet)
-        if self.params.broadcast_enable and not request.slave_id:
-            resp = b"Broadcast write sent - no response expected"
-        else:
-            count = 0
-            while count <= self.params.retries:
-                try:
-                    req = self._build_response(request.transaction_id)
-                    resp = await asyncio.wait_for(
-                        req, timeout=self.comm_params.timeout_connect
-                    )
-                    break
-                except asyncio.exceptions.TimeoutError:
-                    count += 1
-            if count > self.params.retries:
-                self.close(reconnect=True)
-                raise ModbusIOException(
-                    f"ERROR: No response received after {self.params.retries} retries"
+
+        count = 0
+        while count <= self.params.retries:
+            self.transport_send(packet)
+            if self.params.broadcast_enable and not request.slave_id:
+                resp = b"Broadcast write sent - no response expected"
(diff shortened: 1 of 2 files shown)


pymodbus/client/base.py
```diff
@@ -189,25 +189,27 @@ async def async_execute(self, request=None):
...
-                    )
-                    break
-                except asyncio.exceptions.TimeoutError:
-                    count += 1
-            if count > self.params.retries:
-                self.close(reconnect=True)
-                raise ModbusIOException(
-                    f"ERROR: No response received after {self.params.retries} retries"
+
+        count = 0
```
janiversen, line 193: This is not correct, retries is "retry to read" not to send. We cannot automatically send a request multiple time, at least not all requests.

Example: App sends a write request to step up a motor 1 step, if you repeat that it might be fatal. Some devices act on read requests and e.g. initiate an expensive remote request.


Only the app knows if it is safe to resend a request, so we have on purpose left that to the app.

janiversen (changes requested):

janiversen (changes requested):
The parameter is defined as "param retries: (optional) Max number of retries per request." to indicate it is pr request, but that text could be more clear.

janiversen:
An alternative would be to add a "request_retry=False" to the request calls, If true the code should repeat the send as well as build_response.

It must be at request level, because a single device might have a mixture of "dangerous" and free commands.

lsgunth:
I personally find that quite non-intuitive. But it explains the confusion I had around retries in this code base. 

I really don't understand what the point of a retried "read" is. The same thing can be accomplished by multiplying the timeout by 3. 

I just thought I was fixing a bug when I was debugging some communications issues and the debug log said failed after 3 retries even though I only saw one request happen on the wire. I ended up fixing the underlying communications issue so I don't really need the changes in this PR.

Seems to me like it would be more sensible to support automatic retries (including the send) and then just make the default to only send one and make it opt-in to retry. Dangerous applications need not apply.  If someone needs per-request retries that would be another feature they should implement. Ultimately this is sounding like a larger conceptual change that would cut across the whole project.  

Sorry, but I don't really have time for such a project at the moment. Feel free to use my changes as part of a larger effort, if you like. Or, patch 2 fixes one of the skipped tests in a way that aided writing a test for retries, that might be useful.

janiversen:
Your PR was sound which is why I took time to explain the reasoning. 

A change as you suggest is basically the same I suggested, all it takes is an extra parameter in base.py and then test it on in your patch.

Might be I will take a look to amend your patch.

I will take your tests and add them, thanks.

Issue #1761, opened 2023-09-04
Exceptions are returned instead of raised.


matteo-zanoni:
Here it looks like `ModbusIOException` is caught to handle special actions that are needed, after those the exceptions shoud be re-raised but it is returned instead.

This will result in clients behaving like the request completed succesfully but having an exception instead of the usual `ModbusResponse`.

https://github.com/pymodbus-dev/pymodbus/blob/d1af6910bb05b61b0a97e6679f58b0df4e3f58d2/pymodbus/transaction.py#L246C1-L252C27

janiversen:
Well behaving application test ModbusResponse.isError() so how can the application assume the request completed successfully. See our examples.

In general exceptions originating from the device are returned as ModbusResponse, and exceptions are only used for problems in the library (which in theory should never occur).

Your reference is a good example for a couple of reasons, we have received garble from the device
-so we tell the application (with isError) that something went wrong. Without having dived into the specific code, I assume that mixin packs the exc in a ModbusResponse if not that is a bug.
- and depending on the client parameters the library automatically retries.

matteo-zanoni:
The issue I encountered was, infact, that the exception returned in the example I linked was never wrapped in a `ModbusResponse`.

Also some of the confusion comes from the type hints and docstrings of functions like `read_holding_registers`:

```python
    def read_holding_registers(
        self, address: int, count: int = 1, slave: int = 0, **kwargs: Any
    ) -> ModbusResponse:
        """Read holding registers (code 0x03).

        :param address: Start address to read from
        :param count: (optional) Number of coils to read
        :param slave: (optional) Modbus slave ID
        :param kwargs: (optional) Experimental parameters.
        :raises ModbusException:
        """
    ...
```

Here it seems like `ModbusException` could be raised from the function but what happens is that (at least some times, not sure if everytime) the exception is actually returned.

If the intention is to always return and never raise then I would suggest (as @janiversen mentioned) wrapping the exception in a `ModbusResponse` and remove the `:raises ModbusException:` from the docstrings.

janiversen:
I tried to explain that we both use isError and exception, depending on the type of error.

Generally speaking:
- isError is returned when it is something the app should handle as part of the normal sequence.
- Exception is returned for an abnormal situation, mostly from something going wrong within the library itself. This is typically something the app cannot handle without a reset of the client.

janiversen:
[...]

PR #1766, opened 2023-09-08
ModbusIOException as ModbusResponse (#1761)

matteo-zanoni:
This fixes issue #1761


janiversen:
How exactly does this one help (apart from not working), can you please show how the app can have simpler code, because I do not understand it.

janiversen:
What happened ?

I did not even get around to reviewing it, because you still had a lot of CI errors.

I did not say anything negative about the PR, I believe it is fair to question how it helps the app, since you say it solves a problem. To be clear I have nothing against your PR, that is, once the CI show green, we get quite a few PR that I do not personally agree with, but they get merged if they do not break anything.

But of course, your choice.

PR #1793, opened 2023-10-06
Update simulator.py method docstring

yyokusa:


janiversen (approved):
Good catch, thanks.

Copy/paste is a nice but dangerous tool :-)

PR #1796, opened 2023-10-07
CI caching

janiversen:
Let the python build do the caching.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,45 +18,33 @@ on:
 
 jobs:
   linters:
-    name: Linters
-    runs-on: ubuntu-latest
+    name: Linters - ${{ matrix.os.on }} - ${{ matrix.python.version }}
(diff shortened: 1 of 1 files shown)


alexrudd2:
aaaah, no don't do that!  (too late 😆 )

`setup-python` only caches Python (which is fast), and not `pip` (which is slow).  See https://github.com/actions/setup-python/issues/330

Before:
Screenshot 2023-10-16 at 5 01 42 PM

After:
Screenshot 2023-10-16 at 5 01 14 PM

janiversen:
According to the doc it does cache pip, see the parameters used.

alexrudd2:
> According to the doc it does cache pip, see the parameters used.

Yes, that's what the docs say :)  But if you look more closely it only caches the downloads, not the install.  So it's quite slow.  (see screenshots)

janiversen:
I am aware that python-checkout caches the cache, but it a step I needed to make, before making a cache mechanism tailored to us.

PR #1804, opened 2023-10-08
fixes access to asyncio loop via loop property of SerialTransport

MAKOMO:
The property loop of SerialTransport (`transport_seria.py`) is defined as

```
    @property
    def loop(self):
        """Return asyncio event loop."""
        return self._protocol.loop
```

with `self._protocol` of type `asyncio.protocols.BaseProtocol` which does not feature an attribute `loop`, but instead holds the asyncio event loop in the attribute `_loop` thus this definition should read correctly

```
    @property
    def loop(self):
        """Return asyncio event loop."""
        return self._protocol._loop
```

There are two further occurrences of accessing the event loop of the `_protocol` instances via the attribute `loop` instead of `_loop` (in `_read_ready` and `_write_ready`). This PR fixes those and adds a test that demonstrates the issue.

### DIFF
--- a/test/sub_transport/test_basic.py
+++ b/test/sub_transport/test_basic.py
@@ -292,17 +292,6 @@ class TestBasicSerial:
         """Test null modem init"""
         SerialTransport(asyncio.get_running_loop(), mock.Mock(), "dummy")
 
-    async def test_loop(self):
-        """Test access to asyncio event loop of the underlying protocol"""
-        loop = asyncio.get_event_loop()
-        reader = asyncio.StreamReader(loop=loop)
-        comm = SerialTransport(loop, asyncio.StreamReaderProtocol(reader, loop=loop))
-        assert isinstance(comm.loop, asyncio.AbstractEventLoop)
-        protocol = comm.get_protocol()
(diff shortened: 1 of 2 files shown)


janiversen:
Please solve the CI errors.

test/sub_transport/test_basic.py
```diff
@@ -292,6 +292,17 @@ async def test_init(self):
         """Test null modem init"""
         SerialTransport(asyncio.get_running_loop(), mock.Mock(), "dummy")
 
+    async def test_loop(self):
```
janiversen, line 295: We do not test that the libraries work correctly, so I see this test as not needed.

janiversen (changes requested):
The production code change is ok, but please remove the test, for the reasons mentioned.

test/sub_transport/test_basic.py
MAKOMO: I added this test more as illustration that something is wrong in the original code. I will remove it.

Please check again the production code change. Is the "protocol" event loop necessarily always the same as the one of the transport? Anyhow, the original access was faulty.

janiversen (approved):
LGtM, thanks.

PR #1814, opened 2023-10-11
Update doc, remove md files.

janiversen:


janiversen:
And just an advice, if you find bugs in the code, then please add another pull request that solves the bug. This pull request should only be typing.

MAKOMO:
> And just an advice, if you find bugs in the code, then please add another pull request that solves the bug. This pull request should only be typing.

Please discard this Pull Request. I am not able to do what you ask me for. Sorry.

janiversen:
I think the problem is that you try to solve the whole world in one go, that is difficult, even though I know the code quite good, I would have problems doing that.

Instead of giving up, submit a pull request with the simple changes, basically "| None", easy to review and easy to get merged...then you can focus on the more difficult items in smaller chunks.

But of course its all up to you.

PR #1817, opened 2023-10-14
eliminates implicit Optional

MAKOMO:
PEP 484 prohibits implicit Optional for good reasons. See for [Any advantages of enabled mypy `strict_optional?](https://stackoverflow.com/questions/62211237/any-advantages-of-enabled-mypy-strict-optional) some arguments.

Current mypy configuration ignores implicit Optional

```
[tool.mypy]
strict_optional = false
```

with `strict_optional` set mypy raises 87 errors in 13 files. This commit resolves all of them.

While this is mostly about adding/changing type annotations I had to apply few code changes too as well as one (minor) API change (see the marks below). Here are non-trivial issues I encountered. Maybe you find better resolutions than the ones proposed by this PR for those.

1) Subclass should follow the type declarations of its superclass

```python
base.py:ModbusBaseClient
  self.last_frame_end: float = 0

serial.py:ModbusSerialClient(ModbusBaseClient)
  self.last_frame_end = None
```

resolved by widening

```python
self.last_frame_end: float | None = 0
```

2) simple type violation

```python
base.py:ModbusBaseClient(ModbusClientMixin, ModbusProtocol)
   def callback_disconnected(self, _reason: Exception) -> None

transport.py:ModbusProtocol(asyncio.BaseProtocol):
  def callback_disconnected(self, exc: Exception) -> None
  ..
  def transport_close(self, intern: bool = False, reconnect: bool = False) -> None:
    ..
      value.callback_disconnected(None)
```

resolved by widening

```
  def callback_disconnected(self, _reason: Exception | None)
```

3) Maybe frame should not be an optional argument but mandatory? [API CHANGE]

```python
base.py:ModbusBaseClient(ModbusClientMixin, ModbusProtocol)

    def __init__(  # pylint: disable=too-many-arguments
           self, framer: type[ModbusFramer] | None = None,

        self.framer = framer(ClientDecoder(), self)
```

resolved by a cast

```python
        self.framer = cast(type[ModbusFramer], framer)(ClientDecoder(), self)
```

which make pylint to fail under Python 3.8:
```
pymodbus/client/base.py:121:27: E1136: Value 'type' is unsubscriptable (unsubscriptable-object)
```

so I removed the cast and made the framer argument mandatory

```python
    def __init__(  # pylint: disable=too-many-arguments
           self, framer: type[ModbusFramer],
```

This is an API change!

4) self.last_frame_end can be None?

base.py:ModbusBaseClient(ModbusClientMixin, ModbusProtocol)

```python
    def __init__(  # pylint: disable=too-many-arguments
       ...
       self.last_frame_end: float = 0
       ...

    def idle_time(self) -> float:
        ...
        if self.last_frame_end is None or self.silent_interval is None:
            return 0
```

resolved by extending the type declaration of `self.last_frame_end

```python
    def __init__(  # pylint: disable=too-many-arguments
       ...
       self.last_frame_end: float | None = 0
       ...
```

5) Module missing

server/simulator/http_server.py

```python
try:
    from aiohttp import web
except ImportError:
    web = None
```

mypy reports

```
pymodbus/server/simulator/http_server.py:14: error: Incompatible types in assignment (expression has type "None", variable has type Module)  [assignment]
```

resolved by using a cast

```
try:
    from aiohttp import web
except ImportError:
    web = cast(Any, None)
```

see https://github.com/python/mypy/issues/10512

6) Type violation: None is not a dict

mypy reports
```
pymodbus/server/simulator/http_server.py:159: error: Argument 2 to "ModbusSimulatorContext" has incompatible type "Any | None"; expected "dict[str, Callable[..., Any]]"  [arg-type]
```

We have

```python
        self.datastore_context = ModbusSimulatorContext(
            device, custom_actions_dict or None )
```

but

```python
ModbusSimulatorContext:
    def __init__(self, config: Dict[str, Any], custom_actions: Dict[str, Callable]
    ) -> None:


resolved by replacing the None by an empty dict

```python
        self.datastore_context = ModbusSimulatorContext(
            device, custom_actions_dict or {}
        )
```

7) ModbusProtocol.transport not a tuple [CODE CHANGE]

transport/transport.py:ModbusProtocol

```python
def __init__(
        self,
        params: CommParams,
        is_server: bool,
    ) -> None:
...
self.transport: asyncio.BaseTransport = None


async def transport_listen(self) -> bool:

            self.transport = await self.call_create()
            if isinstance(self.transport, tuple):
                self.transport = self.transport[0]
```

How can self.transport is of type asyncio.BaseTransport and thus cannot hold a tuple.

```python
            transport = await self.call_create()
            if isinstance(transport, tuple):
                self.transport = transport[0]
            else:
                self.transport = transport
```

8) baudrate in ModbusSerialClient can be None

client/serial.py

```python
class ModbusSerialClient(ModbusBaseClient):

    def __init__(
        self,
        ...
        baudrate: int = 19200,
        ...


        ModbusBaseClient.__init__(
            self,
            ...
            baudrate=baudrate,
            ...
        )


        self._t0 = float(1 + bytesize + stopbits) / self.comm_params.baudrate
```

but from `transport/transport.py` we see that `self.comm_params.baudrate` could be `None`:

```python
@dataclasses.dataclass
class CommParams:
    ...
    baudrate: int | None = None
    ...
```

resolved by adding an assertion

```python
        assert isinstance(self.comm_params.baudrate, int)
        self._t0 = float(1 + bytesize + stopbits) / self.comm_params.baudrate
```

9) Return type declaration of init_setup_serial wrong [CODE CHANGE]

transport/transport.py

```python
class ModbusProtocol(asyncio.BaseProtocol):

    def __init__(..) -> None:
        ...
            host, port = self.init_setup_serial(host, port)
            if not host and not port:
                return
        ...
    def init_setup_serial(self, host: str, _port: int) -> tuple[str, int]:
        ...
        return None, None
```

The return type of `init_setup_serial` declaration does not match its implementation.

Resolved by extending this type declaration and taking care that in the assignment of `host` and `port` in `__init__` does not violate its declaration:

```python
    def __init__(..) -> None:
        ...
            host_serial, port_serial = self.init_setup_serial(host, port)
            if host_serial is None or port_serial is None:
                return
            host, port = host_serial, port_serial
        ...
    def init_setup_serial(
        self, host: str, _port: int) -> tuple[str, int] | tuple[None, None]:
        ...
        return None, None
```

10) self.loop can be None

transport/transport.py:ModbusProtocol

```python
class ModbusProtocol(asyncio.BaseProtocol):
    def __init__(..) -> None:
        ..
        self.loop: asyncio.AbstractEventLoop | None = None
        ..


    def init_setup_connect_listen(self, host: str, port: int) -> None:
        """Handle connect/listen handler."""
        if self.comm_params.comm_type == CommType.UDP:
            if self.is_server:
                self.call_create = lambda: self.loop.create_datagram_endpoint(
                    self.handle_new_connection,
                    local_addr=(host, port),
                )
```

resolved by adding a cast

```python
    def init_setup_connect_listen(self, host: str, port: int) -> None:
        """Handle connect/listen handler."""
        if self.comm_params.comm_type == CommType.UDP:
            if self.is_server:
                self.call_create = lambda: cast(
                    asyncio.AbstractEventLoop, self.loop
                ).create_datagram_endpoint(
                    self.handle_new_connection,
                    local_addr=(host, port),
                )
```

11) dummy lambda does not return Coroutine as declared [CODE CHANGE]

transport/transport.py:ModbusProtocol
```python
  self.call_create: Callable[[], Coroutine[Any, Any, Any]] = lambda: None
```

None is not a Coroutine

Resolved by using None instead of the dummy lambda

```python
  self.call_create: Callable[[], Coroutine[Any, Any, Any]] | None = None
```

and adding an assertion to

```python
    async def transport_connect(self) -> bool:
        ...
        try:
            self.transport, _protocol = await asyncio.wait_for(
                self.call_create(),
                timeout=self.comm_params.timeout_connect,
            )
        except USEEXCEPTIONS as exc:
```

here and in some related occurrences

```python
    async def transport_connect(self) -> bool:
        ...
        try:
            assert self.call_create is not None
            self.transport, _protocol = await asyncio.wait_for(
                self.call_create(),
                timeout=self.comm_params.timeout_connect,
            )
        except USEEXCEPTIONS as exc:
```

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -22,7 +22,7 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
 
     **Parameters common to all clients**:
 
-    :param framer: (optional) Modbus Framer class.
+    :param framer: Modbus Framer class.
     :param timeout: (optional) Timeout for a request, in seconds.
     :param retries: (optional) Max number of retries per request.
     :param retry_on_empty: (optional) Retry on empty response.
@@ -66,7 +66,7 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
 
     def __init__(  # pylint: disable=too-many-arguments
         self,
-        framer: type[ModbusFramer] | None = None,
+        framer: type[ModbusFramer],
         timeout: float = 3,
         retries: int = 3,
         retry_on_empty: bool = False,
@@ -118,7 +118,7 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
         self.slaves: list[int] = []
 
         # Common variables.
-        self.framer = cast(type[ModbusFramer], framer)(ClientDecoder(), self)
+        self.framer = framer(ClientDecoder(), self)
         self.transaction = DictTransactionManager(
             self, retries=retries, retry_on_empty=retry_on_empty, **kwargs
         )
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import asyncio
 import socket
 from dataclasses import dataclass
-from typing import Any, Callable
+from typing import Any, Callable, cast
 
 from pymodbus.client.mixin import ModbusClientMixin
 from pymodbus.exceptions import ConnectionException, ModbusIOException
@@ -53,20 +53,20 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
     class _params:
         """Parameter class."""
 
-        retries: int = None
-        retry_on_empty: bool = None
-        close_comm_on_error: bool = None
-        strict: bool = None
-        broadcast_enable: bool = None
-        reconnect_delay: int = None
+        retries: int | None = None
+        retry_on_empty: bool | None = None
+        close_comm_on_error: bool | None = None
+        strict: bool | None = None
+        broadcast_enable: bool | None = None
+        reconnect_delay: int | None = None
 
-        source_address: tuple[str, int] = None
+        source_address: tuple[str, int] | None = None
 
-        server_hostname: str = None
+        server_hostname: str | None = None
 
     def __init__(  # pylint: disable=too-many-arguments
         self,
-        framer: type[ModbusFramer] = None,
+        framer: type[ModbusFramer] | None = None,
         timeout: float = 3,
         retries: int = 3,
         retry_on_empty: bool = False,
@@ -118,14 +118,14 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
         self.slaves: list[int] = []
 
         # Common variables.
-        self.framer = framer(ClientDecoder(), self)
+        self.framer = cast(type[ModbusFramer], framer)(ClientDecoder(), self)
         self.transaction = DictTransactionManager(
             self, retries=retries, retry_on_empty=retry_on_empty, **kwargs
         )
-        self.reconnect_delay_current = self.params.reconnect_delay
+        self.reconnect_delay_current = cast(float, self.params.reconnect_delay)
         self.use_udp = False
         self.state = ModbusTransactionState.IDLE
-        self.last_frame_end: float = 0
+        self.last_frame_end: float | None = 0
         self.silent_interval: float = 0
 
     # ----------------------------------------------------------------------- #
@@ -164,7 +164,7 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
             return 0
         return self.last_frame_end + self.silent_interval
 
-    def execute(self, request: ModbusRequest = None) -> ModbusResponse:
+    def execute(self, request: ModbusRequest | None = None) -> ModbusResponse:
         """Execute request and get response (call **sync/async**).
 
         :param request: The request to process
@@ -210,7 +210,7 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
 
         return resp
 
-    def callback_data(self, data: bytes, addr: tuple = None) -> int:
+    def callback_data(self, data: bytes, addr: tuple | None = None) -> int:
         """Handle received data
 
         returns number of bytes consumed
@@ -218,7 +218,7 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
         self.framer.processIncomingPacket(data, self._handle_response, slave=0)
         return len(data)
 
-    def callback_disconnected(self, _reason: Exception) -> None:
+    def callback_disconnected(self, _reason: Exception | None) -> None:
         """Handle lost connection"""
         for tid in list(self.transaction):
             self.raise_future(
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -502,7 +502,7 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
     # code 0x2B sub 0x0D: CANopen General Reference Request and Response, NOT IMPLEMENTED
 
     def read_device_information(
-        self, read_code: int = None, object_id: int = 0x00, **kwargs: Any
+        self, read_code: Union[int, None] = None, object_id: int = 0x00, **kwargs: Any
     ) -> ModbusResponse:
         """Read FIFO queue (code 0x2B sub 0x0E).
 
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -141,6 +141,7 @@ class ModbusSerialClient(ModbusBaseClient):
 
         self.last_frame_end = None
 
+        assert isinstance(self.comm_params.baudrate, int)
         self._t0 = float(1 + bytesize + stopbits) / self.comm_params.baudrate
 
         # Check every 4 bytes / 2 registers if the reading is ready
--- a/pymodbus/client/tcp.py
+++ b/pymodbus/client/tcp.py
@@ -3,7 +3,7 @@ import asyncio
 import select
 import socket
 import time
-from typing import Any, Tuple, Type
+from typing import Any, Optional, Tuple, Type
 
 from pymodbus.client.base import ModbusBaseClient
 from pymodbus.exceptions import ConnectionException
@@ -40,7 +40,7 @@ class AsyncModbusTcpClient(ModbusBaseClient, asyncio.Protocol):
         host: str,
         port: int = 502,
         framer: Type[ModbusFramer] = ModbusSocketFramer,
-        source_address: Tuple[str, int] = None,
+        source_address: Optional[Tuple[str, int]] = None,
         **kwargs: Any,
     ) -> None:
         """Initialize Asyncio Modbus TCP Client."""
@@ -102,7 +102,7 @@ class ModbusTcpClient(ModbusBaseClient):
         host: str,
         port: int = 502,
         framer: Type[ModbusFramer] = ModbusSocketFramer,
-        source_address: Tuple[str, int] = None,
+        source_address: Optional[Tuple[str, int]] = None,
         **kwargs: Any,
     ) -> None:
         """Initialize Modbus TCP Client."""
--- a/pymodbus/client/tls.py
+++ b/pymodbus/client/tls.py
@@ -1,7 +1,7 @@
 """Modbus client async TLS communication."""
 import socket
 import ssl
-from typing import Any, Type
+from typing import Any, Optional, Type
 
 from pymodbus.client.tcp import AsyncModbusTcpClient, ModbusTcpClient
 from pymodbus.framer import ModbusFramer
@@ -44,11 +44,11 @@ class AsyncModbusTlsClient(AsyncModbusTcpClient):
         host: str,
         port: int = 802,
         framer: Type[ModbusFramer] = ModbusTlsFramer,
-        sslctx: ssl.SSLContext = None,
-        certfile: str = None,
-        keyfile: str = None,
-        password: str = None,
-        server_hostname: str = None,
+        sslctx: Optional[ssl.SSLContext] = None,
+        certfile: Optional[str] = None,
+        keyfile: Optional[str] = None,
+        password: Optional[str] = None,
+        server_hostname: Optional[str] = None,
         **kwargs: Any,
     ):
         """Initialize Asyncio Modbus TLS Client."""
@@ -113,11 +113,11 @@ class ModbusTlsClient(ModbusTcpClient):
         host: str,
         port: int = 802,
         framer: Type[ModbusFramer] = ModbusTlsFramer,
-        sslctx: ssl.SSLContext = None,
-        certfile: str = None,
-        keyfile: str = None,
-        password: str = None,
-        server_hostname: str = None,
+        sslctx: Optional[ssl.SSLContext] = None,
+        certfile: Optional[str] = None,
+        keyfile: Optional[str] = None,
+        password: Optional[str] = None,
+        server_hostname: Optional[str] = None,
         **kwargs: Any,
     ):
         """Initialize Modbus TLS Client."""
--- a/pymodbus/server/simulator/http_server.py
+++ b/pymodbus/server/simulator/http_server.py
@@ -5,13 +5,13 @@ import importlib
 import json
 import os
 from time import time
-from typing import List
+from typing import Any, List, Optional, cast
 
 
 try:
     from aiohttp import web
 except ImportError:
-    web = None
+    web = cast(Any, None)
 
 import contextlib
 
@@ -124,7 +124,7 @@ class ModbusSimulatorServer:
         http_port: int = 8080,
         log_file: str = "server.log",
         json_file: str = "setup.json",
-        custom_actions_module: str = None,
+        custom_actions_module: Optional[str] = None,
     ):
         """Initialize http interface."""
         if not web:
@@ -157,7 +157,7 @@ class ModbusSimulatorServer:
             del server["port"]
         device = setup["device_list"][modbus_device]
         self.datastore_context = ModbusSimulatorContext(
-            device, custom_actions_dict or None
+            device, custom_actions_dict or {}
         )
         datastore = ModbusServerContext(slaves=self.datastore_context, single=True)
         comm = comm_class[server.pop("comm")]
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -54,7 +54,7 @@ import ssl
 import sys
 from contextlib import suppress
 from enum import Enum
-from typing import Any, Callable, Coroutine
+from typing import Any, Callable, Coroutine, cast
 
 from pymodbus.logging import Log
 from pymodbus.transport.transport_serial import create_serial_connection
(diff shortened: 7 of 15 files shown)


pymodbus/client/base.py
```diff
@@ -122,10 +122,10 @@ def __init__(  # pylint: disable=too-many-arguments
         self.transaction = DictTransactionManager(
             self, retries=retries, retry_on_empty=retry_on_empty, **kwargs
         )
-        self.reconnect_delay_current = self.params.reconnect_delay
+        self.reconnect_delay_current = cast(float, self.params.reconnect_delay)
```
janiversen, line 125: both delay_current and reconnect_delay are (or should be float) so why the cast.

cast is an ugly beast, that preferable never should be used, it typically hides a fundamental problem, likely maybe in this case.

pymodbus/client/base.py
```diff
@@ -210,15 +210,15 @@ async def async_execute(self, request=None):
...
+    def callback_data(self, data: bytes, addr: tuple | None = None) -> int:
         """Handle received data
 
         returns number of bytes consumed
         """
         self.framer.processIncomingPacket(data, self._handle_response, slave=0)
         return len(data)
 
-    def callback_disconnected(self, _reason: Exception) -> None:
+    def callback_disconnected(self, _reason: Exception | None) -> None:
```
janiversen, line 221: Can that really be None ? where is it called with None ?

pymodbus/client/mixin.py
```diff
@@ -502,7 +502,7 @@ def read_fifo_queue(self, address: int = 0x0000, **kwargs: Any) -> ModbusRespons
     # code 0x2B sub 0x0D: CANopen General Reference Request and Response, NOT IMPLEMENTED
 
     def read_device_information(
-        self, read_code: int = None, object_id: int = 0x00, **kwargs: Any
+        self, read_code: Union[int, None] = None, object_id: int = 0x00, **kwargs: Any
```
janiversen, line 505: Union ?? I think we should either use "Union" or "|" not both.

pymodbus/client/serial.py
```diff
@@ -141,6 +141,7 @@ def __init__(
 
         self.last_frame_end = None
 
+        assert isinstance(self.comm_params.baudrate, int)
```
janiversen, line 144: No assert in production code please. Apart from that this is a strange place to assert on an internal struct...I could understand (but not accept) if you did assert on a parameter.

pymodbus/client/tcp.py
```diff
@@ -40,7 +40,7 @@ def __init__(
         host: str,
         port: int = 502,
         framer: Type[ModbusFramer] = ModbusSocketFramer,
-        source_address: Tuple[str, int] = None,
+        source_address: Optional[Tuple[str, int]] = None,
```
janiversen, line 43: Please be consistent !, this can be expressed with "|" as well.

pymodbus/client/tcp.py
```diff
@@ -102,7 +102,7 @@ def __init__(
...
```
janiversen, line 105: Same here, please do not mix "Union", "Optional" and "|", that makes the code more complicated than needed.

pymodbus/client/tls.py
```diff
@@ -44,11 +44,11 @@ def __init__(
         host: str,
         port: int = 802,
         framer: Type[ModbusFramer] = ModbusTlsFramer,
-        sslctx: ssl.SSLContext = None,
-        certfile: str = None,
-        keyfile: str = None,
-        password: str = None,
-        server_hostname: str = None,
+        sslctx: Optional[ssl.SSLContext] = None,
```
janiversen, line 47: Why not "|" as used in other places. Please be consistent.

pymodbus/server/simulator/http_server.py
```diff
@@ -5,13 +5,13 @@
...
 from time import time
-from typing import List
+from typing import Any, List, Optional, cast
 
 
 try:
     from aiohttp import web
 except ImportError:
-    web = None
+    web = cast(Any, None)
```
janiversen, line 14: This seems very wrong.

pymodbus/transport/transport.py
```diff
@@ -177,16 +177,19 @@ def __init__(
             host = self.comm_params.host
             port = int(self.comm_params.port)
         if self.comm_params.comm_type == CommType.SERIAL:
-            host, port = self.init_setup_serial(host, port)
-            if not host and not port:
+            host_serial, port_serial = self.init_setup_serial(host, port)
+            if host_serial is None or port_serial is None:
                 return
+            host, port = host_serial, port_serial
```
janiversen, line 183: This is just complicating things ! please use host, port directly.

pymodbus/transport/transport.py
```diff
@@ -177,16 +177,19 @@ def __init__(
...
         if host == NULLMODEM_HOST:
             self.call_create = lambda: self.create_nullmodem(port)
             return
         # TCP/TLS/UDP
         self.init_setup_connect_listen(host, port)
 
-    def init_setup_serial(self, host: str, _port: int) -> tuple[str, int]:
+    def init_setup_serial(
+        self, host: str, _port: int
+    ) -> tuple[str, int] | tuple[None, None]:
```
janiversen, line 192: Can this really return None, None ?

pymodbus/transport/transport.py
```diff
@@ -211,19 +214,25 @@ def init_setup_connect_listen(self, host: str, port: int) -> None:
         """Handle connect/listen handler."""
         if self.comm_params.comm_type == CommType.UDP:
             if self.is_server:
-                self.call_create = lambda: self.loop.create_datagram_endpoint(
+                self.call_create = lambda: cast(
```
janiversen, line 217: That is wrong, call_create is NOT loop.

pymodbus/transport/transport.py
```diff
@@ -289,7 +304,7 @@ def connection_made(self, transport: asyncio.BaseTransport):
         self.reset_delay()
         self.callback_connected()
 
-    def connection_lost(self, reason: Exception):
+    def connection_lost(self, reason: Exception | None) -> None:
```
janiversen, line 307: Can that be called with None ?

janiversen (changes requested):
Something to get you going.

A couple of general comments:
- do not mix "Union", "Optional" and "|", decide what to use and use it
- do not use "cast" unless there is a very good reason to do so

Once these parts are done, please ask for a new review.

janiversen:
I wrote "choose", but thinking a second time, the preference is to use "|" because that is the future.

pymodbus/client/base.py
```diff
@@ -22,7 +22,7 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
 
     **Parameters common to all clients**:
 
-    :param framer: (optional) Modbus Framer class.
+    :param framer: Modbus Framer class.
```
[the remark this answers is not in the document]
MAKOMO: Did you read my comment 3? If framer is optional and defaults to None then it should not be called unprotected in `__init__`:

```
self.framer = framer(ClientDecoder(), self)
```

I don't think `None` can be called.

MAKOMO:
> I wrote "choose", but thinking a second time, the preference is to use "|" because that is the future.

Yes

pymodbus/client/base.py
```diff
@@ -53,20 +53,20 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
...
-        source_address: tuple[str, int] = None
+        source_address: tuple[str, int] | None = None
 
-        server_hostname: str = None
+        server_hostname: str | None = None
 
     def __init__(  # pylint: disable=too-many-arguments
         self,
-        framer: type[ModbusFramer] = None,
+        framer: type[ModbusFramer],
```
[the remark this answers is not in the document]
MAKOMO: Well, there is a type error in `__init__` if this stays optional which should then be resolved in another way. How to initialize `self.framer` if the argument to `framer` not given and thus `framer` defaults to None?

pymodbus/client/base.py
MAKOMO: `self.params.reconnect_delay` defaults to `None`. Maybe it shouldn't.

pymodbus/client/base.py
MAKOMO: See my comment 2

pymodbus/server/simulator/http_server.py
MAKOMO: Yes. The variable `web` should not be used as package and as flag if the package is available.

pymodbus/transport/transport.py
MAKOMO: See my comment 9

pymodbus/transport/transport.py
MAKOMO: Yes, also in comment 9.

pymodbus/client/base.py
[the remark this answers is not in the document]
janiversen: That is another problem, which I think is handled at lower levels in the code.

Please do not change API, because this will cause every app to be changed.

pymodbus/transport/transport.py
MAKOMO: No, `self.loop` has to be a loop to call `create_datagram_endpoint`. See comment 10.

pymodbus/client/base.py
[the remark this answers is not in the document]
janiversen: Apart from that the type error is quite easy to solve without changing the parameter itself.

pymodbus/client/base.py
janiversen: Yes it should

pymodbus/transport/transport.py
MAKOMO: See comment 11.

pymodbus/client/base.py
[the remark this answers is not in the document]
janiversen: No I do not read comments, I review code. All optional parameters are described as optional, and framer is and should be optional. 

None is a legal value for framer at API level.

pymodbus/client/base.py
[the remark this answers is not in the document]
MAKOMO: Ok, I did not see how to resolve that easily. Here I need your support.

pymodbus/client/base.py
MAKOMO: Ok, but if `self.params.reconnect_delay` can be `None` it should not be assigned to `self.reconnect_delay_current` which is declared to be a `float` in 

```
self.reconnect_delay_current = self.params.reconnect_delay
```

pymodbus/client/base.py
janiversen: Again I review your code...All call I can see have an exception.

pymodbus/server/simulator/http_server.py
janiversen: I am sorry that is a very traditional way to include optional packages.

If the package is not installed the variable "web" still need to be declared.

pymodbus/transport/transport.py
janiversen: It is a not acceptable complication, that brings nothing.

pymodbus/client/base.py
MAKOMO: In `transport/transport.py:transport_close` there is a `value.callback_disconnected(None)` as I wrote in the PR comment.

pymodbus/transport/transport.py
MAKOMO: Well it makes this type correct

janiversen:
I will wait to comment/review until you have made the requested changes, otherwise we keep commenting on the same things.

Please add me a reviewer when it is ready to be reviewed.

janiversen:
I think the problem is that you try to solve the whole world in one go, that is difficult, even though I know the code quite good, I would have problems doing that.

Instead of giving up, submit a pull request with the simple changes, basically "| None", easy to review and easy to get merged...then you can focus on the more difficult items in smaller chunks.

But of course its all up to you.

(not sure why we started talking about this PR in my PR).

MAKOMO:
> > I wrote "choose", but thinking a second time, the preference is to use "|" because that is the future.
> 
> Yes

I just replaced all Unions in client/mixing.py by | bars. Fine for mypy, but pylint complaints. Not supported under Python 3.8? Strange as transport/transport.py the | bars are accepted. Somehow the CI seems to treat different files in this project differently. I seem not to be the right person to bring this forward. Sorry.

janiversen:
CI are not treating the files differently, but the files causing you problems might miss an import at the very top, to allow future changes.

janiversen:
If I may be blunt, I do not think it's you as a person that's wrong....I actually find it refreshing to see your PR with interesting changes.

It seems you lack som python experience. The difference in the files are this "from __future__ import annotations" and not CI or other things,

A similar problem was with the import variable "web", where it is correct that we want optional import, but there are different ways of doing it. transport_serial.py have a way that do not make mypy complain.

So back to what I started saying do not try to solve the whole world in one pull request, that is extremely difficult, use the "salami tactic", and make smaller easier pull requests. Just a recommendation.

Ps. I still think your goal is correct, and I hope you noted that I have not spoken against your changes just tried to guide you in the direction where we all benefit.

alexrudd2:
This situation is unfortunate.  A well-meaning contributor and a well-meaning maintainer became frustrated at each other over some misunderstandings.  I hope it may be possible to try again.  :)

@janiversen I recommend you follow Marko's advice and read the pull request comments!  He offers good analysis of the problems `mypy` finds, which are _not_ apparent reading the code.  Tthere are times he in fact caught something you missed.  Reading the comments will make his intentions clear and your improvements more accurate.

@MAKAMO I recommend you follow Jan's advice and split this PR into smaller ones, starting with one change per PR.  This makes it much easier to keep the comments near the code and review them. 
I also recommend you setup the CI on your own fork, which will help you understand the different Python versions.

alexrudd2:
As a first example, there was much discussion regarding `Optional`, `Union`, and `I`.  Jan expressed frustration at inconsistency and prefers `I`, and Marko expressed frustration that `I` does not always work in the CI.

Using `|` instead of `Optional[foo]` or `Union[foo, None]` is [PEP-0604](https://peps.python.org/pep-0604/), and was only added in Python3.10.

I presume Marko is using Python > 3.10 and so `|` always works.  The CI tests against many versions, including `3.8` and `3.9`.  So it sometimes fails.
https://github.com/pymodbus-dev/pymodbus/blob/2adff1fe5564526e3fd742de5b05e1c8790356f8/.github/workflows/ci.yml#L31-L36

The answer is to use for the cases where it fails.
```
from __future__ import annotations
```

janiversen:
Yes it is very unfortunate! and I apologize but the pull request was big and I saw the same problems again and again.

what especially concerned me was the inconsistent use of | and Option which was caused by not importing future (which I explained)

I hope he will come back, because his work was not bad, just needed finishing touches...which I hope I explained directly earlier.

alexrudd2:
When passionate Spain and methodical Germany can work peacefully together, the world will be amazed 😄 .  I am going to try a few smaller PRs.

pymodbus/transport/transport.py
alexrudd2: As MAKOMO explained, the default is in fact calling with `None` (via a lambda)

https://github.com/pymodbus-dev/pymodbus/blob/2adff1fe5564526e3fd742de5b05e1c8790356f8/pymodbus/transport/transport.py#L162

pymodbus/client/serial.py
alexrudd2: The intention here was to [narrow the type](https://mypy.readthedocs.io/en/stable/type_narrowing.html) to an `int`.

However, the better solution is to find where `mypy` does not understand the default value (which _is_ an `int`) will be used.  There is probably a small bug with parameters in subclasses.

pymodbus/client/mixin.py
alexrudd2: Use `from __future__ import annotations` to allow `|` to be used.

I will mark the other redundant comments Resolved.

janiversen:
heh, do not forget I am a real Viking (dane), living in Spain...I did live 5 years in Vienna if that counts.

alexrudd2:
See https://github.com/pymodbus-dev/pymodbus/pull/1825 as an example

janiversen:
Yes and see #1827, to see why not just accept everything.

Everything that makes the code more stable or faster are super, but changes that makes life harder for users of the library or complicates the code just to please mypy are at the very least something that should be discussed.

alexrudd2:
For an update on progress:

When the PR was created:
> Found 87 errors in 13 files

Today:
> Found 41 errors in 7 files

janiversen:
Seems this PR is no longer updated, and have been replaced by a number of PR´s by @alexrudd2.

I am closing this, but the base of this is not bad !!! it just needed a bit of hand holding which have happened thanks to @alexrudd2. I am sorry for not being of more help, but mypy is to me working a lot against the nature of python. To be fair I am a C/C++ programmer so I am used to very strict typing etc, and python offered an interesting alternative.

PR #1818, opened 2023-10-14
Import aiohttp in way pleasing mypy.

janiversen:


alexrudd2:
This actually does not do what you intended, if `aiohttp` is not installed.  I will submit a PR

```
>       if not web:
E       NameError: name 'web' is not defined
```

janiversen:
It actually does, we do the same with pyserial.

If you call the methods using it without having it installed it's ok to get an error.

However if you have a better solution, then please.....but also in transport_serial.

PR #1826, opened 2023-10-16
Add 3.12 classifier

alexrudd2:


janiversen:
3.12 is still experimental, there seems to be a few async problems.

alexrudd2:
> 3.12 is still experimental, there seems to be a few async problems.

Interesting, do you have any descriptions of them?

My CI runs fine on `3.12` (but it's not very difficult).  I have not deployed any real devices yet.

janiversen:
no, its hidden in transport and transport_serial....basically it seems the sequencing in asyncio have changed (e.g. connection_lost is called a lot later).

Issue #1128, opened 2022-10-21
Runtime is dependant on pyserial, even when using TCP/IP


alexrudd2:
### Versions

* Python: 3.9 or 3.10
* OS: MacOSX
* Pymodbus: 3.0.0
* Modbus Hardware (if used): N/A

### Pymodbus Specific
* Client: tcp async

### Description

pymodbus `ReconnectingAsyncioModbusTcpClient` fails if `pyserial` is not installed.  This is listed as optional in `requirements.txt`
https://github.com/riptideio/pymodbus/blob/b4f5a558254f8242fd65fb49fe10f4f255610b32/requirements.txt#L31-L33

### Code and Logs

```
pip uninstall pyserial
pip install pymodbus==3.0.0

python3
>    from pymodbus.client import AsyncModbusTcpClient
  File "/opt/homebrew/lib/python3.9/site-packages/pymodbus/client/__init__.py", line 5, in <module>
    from pymodbus.client.serial import AsyncModbusSerialClient, ModbusSerialClient
  File "/opt/homebrew/lib/python3.9/site-packages/pymodbus/client/serial.py", line 7, in <module>
    from serial_asyncio import create_serial_connection
  File "/opt/homebrew/lib/python3.9/site-packages/serial_asyncio/__init__.py", line 21, in <module>
    import serial
ModuleNotFoundError: No module named 'serial'
```

janiversen:
Correct, this was reported earlier and is solved on dev. We expect to release 3.0.1 next month with this and other bugs solved..

https://github.com/riptideio/pymodbus/commit/7e58d51d1adb431f7f4d110815235af69b8e977d

alexrudd2:
Aha, I was in the process of submitting a PR but you are too fast for me :)

This was how I had solved it, but `dev` works fine for me.
```
from pymodbus.client.base import ModbusBaseClient
try:
    from pymodbus.client.serial import AsyncModbusSerialClient, ModbusSerialClient
except ModuleNotFoundError:
    pass
```

janiversen:
Heh, I did that PR a couple of days ago...but please keep submitting PR´s, in open source 1+1 = 3 is really true.

alexrudd2:
I've spent this week updating a number of drivers we ([numat](https://github.com/numat)) maintain.  So far 3.0.x appears to be working well after the port.

* https://github.com/numat/productivity/pull/42
* [numat/clickplc](https://github.com/numat/clickplc)
[...]

PR #1827, opened 2023-10-17
Consistent ImportError messages

alexrudd2:
See https://github.com/pymodbus-dev/pymodbus/pull/1818#issuecomment-1765423835

Thanks @MAKOMO

### DIFF
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -1,7 +1,6 @@
 """Modbus client async serial communication."""
 import asyncio
 import time
-from contextlib import suppress
 from functools import partial
 from typing import Any, Type
 
@@ -14,8 +13,13 @@ from pymodbus.transport import CommType
 from pymodbus.utilities import ModbusTransactionState
 
 
-with suppress(ImportError):
+try:
     import serial
+except ImportError:
+    raise ImportError(  # pylint: disable=raise-missing-from
+        "Serial client requires pyserial "
+        'Please install with "pip install pyserial" and try again.'
+    )
 
 
 class AsyncModbusSerialClient(ModbusBaseClient, asyncio.Protocol):
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -23,10 +23,6 @@ from pymodbus.transaction import (
 from pymodbus.transport import CommParams, CommType, ModbusProtocol
 
 
-with suppress(ImportError):
-    pass
-
-
 # --------------------------------------------------------------------------- #
 # Protocol Handlers
 # --------------------------------------------------------------------------- #
--- a/pymodbus/server/reactive/main.py
+++ b/pymodbus/server/reactive/main.py
@@ -15,11 +15,10 @@ from enum import Enum
 try:
     from aiohttp import web
 except ImportError:
-    print(
+    raise ImportError(  # pylint: disable=raise-missing-from
         "Reactive server requires aiohttp. "
         'Please install with "pip install aiohttp" and try again.'
     )
-    sys.exit(1)
 
 from pymodbus import __version__ as pymodbus_version
 from pymodbus.datastore import ModbusServerContext, ModbusSlaveContext
--- a/pymodbus/server/simulator/http_server.py
+++ b/pymodbus/server/simulator/http_server.py
@@ -9,8 +9,13 @@ from time import time
 from typing import List
 
 
-with contextlib.suppress(ImportError):
+try:
     from aiohttp import web
+except ImportError:
+    raise ImportError(  # pylint: disable=raise-missing-from
+        "Simulator server requires aiohttp. "
+        'Please install with "pip install aiohttp" and try again.'
+    )
 
 from pymodbus.datastore import ModbusServerContext, ModbusSimulatorContext
 from pymodbus.datastore.simulator import Label
@@ -124,8 +129,6 @@ class ModbusSimulatorServer:
         custom_actions_module: str = None,
     ):
         """Initialize http interface."""
-        if not web:
-            raise RuntimeError("aiohttp not installed!")
         with open(json_file, encoding="utf-8") as file:
             setup = json.load(file)
 


pymodbus/client/serial.py
```diff
@@ -14,8 +13,13 @@
 from pymodbus.utilities import ModbusTransactionState
 
 
-with suppress(ImportError):
+try:
     import serial
+except ImportError:
```
janiversen, line 18: This is wrong, because it demands users to have the dependency installed.

The existing version, allows the user not to have the dependency installed, but of course he/she is then not allowed to use these methods.

We do NOT want to demand users to install e.g. pyserial / aiohttp if they are going to use serial communication / the simulator.

janiversen (changes requested):

pymodbus/client/serial.py
janiversen: what I did earlier to avoid followup errors was to have

```
except ImportError:
    serial = None
```

alexrudd2: > We do NOT want to demand users to install e.g. pyserial / aiohttp if they are going to use serial communication / the simulator.

I'm confused.  What use is importing the serial client without `pyserial`, or importing the HTTP simulator without `aiohttp`?  They will not run, and fail with some confusing error message.

Are you concerned about importing `pymodbus` to use another part? (for instance, the TCP client as in https://github.com/pymodbus-dev/pymodbus/issues/1128) I checked and this is not a problem.  
```
>>> import serial
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'serial'
>>> from pymodbus.client import AsyncModbusTcpClient
>>>
```

Last year pymodbus did something like `import *`, but now after your refactors it does not.  So I'm not sure how the ImportError would be triggered when it shouldn't be.

alexrudd2:
> changes that makes life harder for users of the library or complicates the code just to please mypy are at the very least something that should be discussed.

Yes, that is my intention as well!

Depending on how it's called, `dev` produces one of these error messages.
```
>       if not web:
E       NameError: name 'web' is not defined
```
```
>>> from pymodbus.server.simulator import ModbusSimulatorServer
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'ModbusSimulatorServer' from 'pymodbus.server.simulator' (/Users/a.ruddick/Documents/github/pymodbus/pymodbus/server/simulator/__init__.py)
```

With this PR, the error message is below. I find this much more helpful.

```
<traceback>
ImportError: Simulator server requires aiohttp. Please install with "pip install aiohttp" and try again.
```

pymodbus/client/serial.py
janiversen: Because the user does "import pymodbus" and that scans all py files, and it then breaks down.

Try to uninstall pyserial and try for yourself, I had the problem about a month ago. aiohttp was never the problem, but pyserial have caused a couple of issues.

See this:
```
(pymodbus) jan@MacMini examples % pip uninstall pyserial
...
(pymodbus) jan@MacMini examples % ./client_async.py
  File "/Volumes/DiskHomes/home_jan/repos/pymodbus/pymodbus/client/serial.py", line 17, in <module>
    import serial
ModuleNotFoundError: No module named 'serial'

During handling of the above exception, another exception occurred:

  File "/Volumes/DiskHomes/home_jan/repos/pymodbus/examples/./client_async.py", line 35, in <module>
    from pymodbus.client import (
  File "/Volumes/DiskHomes/home_jan/repos/pymodbus/pymodbus/client/__init__.py", line 16, in <module>
    from pymodbus.client.serial import AsyncModbusSerialClient, ModbusSerialClient
  File "/Volumes/DiskHomes/home_jan/repos/pymodbus/pymodbus/client/serial.py", line 19, in <module>
    raise ImportError(  # pylint: disable=raise-missing-from
ImportError: Serial client requires pyserial Please install with "pip install pyserial" and try again.
(pymodbus) jan@MacMini examples % 
```
client_async without parameters uses a tcp client.

janiversen: Done with this PR.

janiversen:
Well the code as it is without this PR, do not have problem, nor did it have problems with the "web".

But the error message you get is exactly what we try to avoid by making a conditional import. In reality, apart from the error message, there are no reason for your try/except.

pymodbus/client/serial.py
alexrudd2: > Because the user does "import pymodbus" and that scans all py files, and it then breaks down. [...]
> Try to uninstall pyserial and try for yourself, I had the problem about a month ago. 

Yes, I did :). It's fine
Screenshot 2023-10-17 at 11 27 37 AM

```
(pymodbus) jan@MacMini examples % ./client_async.py
  File "/Volumes/DiskHomes/home_jan/repos/pymodbus/pymodbus/client/serial.py", line 17, in <module>
    import serial
ModuleNotFoundError: No module named 'serial'
```

OK, now I understand what I missed.  _The examples_ fail, not all of pymodbus.  I'll look to see if that can be changed.

alexrudd2:
> In reality, apart from the error message, there are no reason for your try/except.

Yes, I agree.  The entire point (instead of `contextlib.suppress`) is for a more helpful error message, so people do not try to e.g. `pip install serial` and get confused.

janiversen:
> > In reality, apart from the error message, there are no reason for your try/except.
> 
> Yes, I agree. The entire point (instead of `contextlib.suppress`) is for a more helpful error message, so people do not try to e.g. `pip install serial` and get confused.

You miss the point, it is a conditional import, so all the users that do not use simulator and/or serial do not need to install these libraries...and as you might already have seen from my test above, uninstalling pyserial and running client_async.py caused problems.

janiversen:
The solution we have currently on dev allows:
```
$ pip uninstall pyserial
$ ./client_async.py --comm tcp
```
This PR should not change that.

alexrudd2:
> The solution we have currently on dev allows:
> 
> ```
> $ pip uninstall pyserial
> $ ./client_async.py --comm tcp
> ```
> 
> This PR should not change that.

Yes, understood.  I think the answer is to use conditional importing (and/or error suppression) in the examples as well.

janiversen:
> > The solution we have currently on dev allows:
> > ```
> > $ pip uninstall pyserial
> > $ ./client_async.py --comm tcp
> > ```
> > 
> > 
> > This PR should not change that.
> 
> Yes, understood. I think the answer is to use conditional importing (and/or error suppression) in the examples as well.

NO because the examples are building block of applications, and we do not want to force applications to use conditional imports.

Apart from the fact that I would not know which import should be conditional...because the application import pymodbus, pymodbus.client etc....the application do not even know (in theory) that we import pysierlal and other dependencies.

alexrudd2:
Can I ask you to slow down a bit, please?  😃 

> because the application import pymodbus, pymodbus.client

Yes, both of these are fine.
```
>>> import serial
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'serial'
>>> from pymodbus.client import AsyncModbusTcpClient
>>>
```
image


Only the examples are problematic, because they import everything (even if unneeded).  It is here where I propose to suppress the unused imports.

https://github.com/pymodbus-dev/pymodbus/blob/4ddb9b136838da6a35f755252f317e4a3df7fbe8/examples/client_async.py#L37-L42

janiversen:
And I politely disagree 😃 The application should be allowed to import everything, but not to instantiate the classes.

I am not sure we still have it, but earlier you would get an error message if you did client = AsyncModbusSerialClient() and did not have the library installed.

Please do not forget that many users copy our examples (that is sort of the idea) so if we introduce conditionals in the examples it grows into the applications.

I really do not understand what we are discussing....this works nicely on dev, so why change it to something less handy ?

alexrudd2:
> And I politely disagree 😃 The application should be allowed to import everything, but not to instantiate the classes.

OK, perhaps it's best to check at time of use rather than time of import.

> I am not sure we still have it, but earlier you would get an error message if you did client = AsyncModbusSerialClient() and did not have the library installed.

No, it fails silently now.

> I really do not understand what we are discussing....this works nicely on dev, so why change it to something less handy ?

Well, there are multiple cases that are inconsistent.
* Running the reactive simulator without `aiohttp` gives a clear warning and does a `sys.exit(1)`.
* Running the HTTP simulator with `aiohttp` gives a confusing `NameError`
```
 if not web:
E       NameError: name 'web' is not defined
```
* Running the client(s) or examples without `pyserial` gives a `NameError`, which is OK but could be better (suggesting `pyserial` not `serial`)

I am trying to make them all the same, and helpful.  (Also avoid the wrath of strict `mypy` 😆 )

janiversen:
Sorry if the error message dropped out (aiohttp in simulator probably never had it, I am guilty in not being consistent) ! 

I am all in favour of having a "common" error message if a dependency is missing ... 

I think the easiest (apart from mypy) is something like:

```
try
   import x
except:
  x = None
```

And then in the relevant base class __init__():
```
   if not x:
     nice error message
     exit(-1)
```

I have no idea how that works with mypy, but it would work well with users.

janiversen:
Your suggestion is a good compromise, with a small change, because I agree that sys.exit is ugly, how about
```
__init__():
    if x_missing:
        raíse RuntimeError("nice error message.....")
```

That could be used in REPL as well, there are a number of dependencies there.

janiversen:
Forgot to write using RuntimeError secures the exception is not caught in a wildcard except. I think runtime error is the only exception not catchable.

PR #1829, opened 2023-10-17
Cache

janiversen:
with python in place, we can make our own cache key, and thereby only have:

3 (Windows, Ubuntu, MacOS) x 5 (python 3.8 - 313) = 15 cache.

We need to avoid making new caches for each pull request, that currently is the cause of why some jobs are slow in starting and that the real clock time is too high.

Current timing is at 15 minutes 2 seconds.

1) Change work order since windows takes longer it should be started first.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
       - name: pytest
         run: |
           env
-          pytest -n0  -v --full-trace --timeout=1200
+          pytest -v --full-trace --timeout=1200
 
   analyze:
     name: Analyze Python
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - dev
       - master
+      - dev_*
     tags:
       - v*
   pull_request:
@@ -13,8 +14,6 @@ on:
     # Sunday at 02:10 UTC.
     - cron: '10 2 * * 0'
   workflow_dispatch:
-
-# Test with new caching.
 
 jobs:
   testing:
@@ -48,11 +47,23 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
 
+      - name: Set venv path (NON Windows)
+        if: matrix.os != 'windows-latest'
+        run: |
+          echo "VIRTUAL_ENV=${{ github.workspace }}/venv" >> $GITHUB_ENV
+          echo ${{ github.workspace }}/venv/bin >> $GITHUB_PATH
+
+      - name: Set venv path (Windows)
+        if: matrix.os == 'windows-latest'
+        run: |
+          echo "VIRTUAL_ENV=${{ github.workspace }}\\venv" >> $Env:GITHUB_ENV
+          echo "${{ github.workspace }}\\venv\\Scripts" >> $Env:GITHUB_PATH
+
       - name: Restore base Python virtual environment
         id: cache-venv
         uses: actions/cache@v3.3.2
         with:
-          path: ${{ github.workspace }}/venv
+          path: ${{ env.VIRTUAL_ENV }}
           key: >-
             ${{ runner.os }}-${{ matrix.python }}-venv-${{
               hashFiles('pyproject.toml') }}-${{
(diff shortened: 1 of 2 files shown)


alexrudd2:
The two other big slowdowns:

(1) Switching to using pre-commit means that its environment needs to be initialized
(2) https://github.com/pymodbus-dev/pymodbus/pull/1673 removed the parallel execution with `pytest-xdist`.  (Hopefully the Python3.11 problem is transient and this can be reverted)

janiversen:
Removing xdist did not make a big difference because I controlled something like 20 runs in we más had 2 cores. But of course once I get rid of the spontaneous errors I will activate it again.

I made another change that speed things up, nullmodem ! the tests (apart from a few) no longer uses the network but communicates through a couple of class variables

Pre-commit is actually expensive in CI, but your idea with the env variable could be the solution.

alexrudd2:
> I made another change that speed things up, nullmodem ! the tests (apart from a few) no longer uses the network but communicates through a couple of class variables

o_0 that's great!
 
> Pre-commit is actually expensive in CI, but your idea with the env variable could be the solution.

What if we don't use it in CI?  We can use something like `renovatebot` to keep the versions in `pre-commit` in sync with the versions in `pyproject.toml`.

janiversen:
That's another possibility, however I have no experience in using that bot...it needs to read the precommit yaml and then somehow update py project.toml

I hope that soon precommit will be able to use toml so we can get rid of this double configuration.

alexrudd2:
> That's another possibility, however I have no experience in using that bot...it needs to read the precommit yaml and then somehow update py project.toml

It's a great bot; if you'd like I can set it up. Here's an [example](https://github.com/numat/alicat/pull/86/files) of updating both files at the same time.

> I hope that soon precommit will be able to use toml so we can get rid of this double configuration.

Unfortunately, the maintainer is an arrogant jerk that will never do this: https://github.com/pre-commit/pre-commit/issues/1165. :(

alexrudd2:
Also, we should be able to drop `black` soon in favor of `ruff`
https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_formatter/README.md

janiversen:
> Also, we should be able to drop `black` soon in favor of `ruff` https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_formatter/README.md
> 
> In fact it could be done today, if you are OK with some small formatting changes and using an alpha.

I am all in favor of moving towards ruff, so please do..

If I remember right you were also closing in on pylint?


Regarding the bot, I will have to read a bit more, I am pr nature sceptical with apps that run outside github, but it does look like a good candidate. An alternative is precommit.ci that seems to do a similar thing.

I think github also have a bot that updates, I remember we use (or used) that in Homeassistant.

alexrudd2:
> I am all in favor of moving towards ruff, so please do..

OK, I will test the alpha.  It has ~80 minor formatting changes compared to `black` right now.
 
> If I remember right you were also closing in on pylint?

Many of pylint's rules can be done by `ruff`, but it does more thorough checks and cannot be fully replaced yet (https://github.com/astral-sh/ruff/issues/970)

> Regarding the bot, I will have to read a bit more, I am pr nature sceptical with apps that run outside github, but it does look like a good candidate. An alternative is precommit.ci that seems to do a similar thing.

`precommit.ci` suffers from the problem of the same arrogant author and does not play well with other tools.  I would not recommend it.

> I think github also have a bot that updates, I remember we use (or used) that in Homeassistant.

Yeah, it's `dependabot`.  Not a bad bot, but `renovatebot` is significantly better.

janiversen:
@alexrudd2 I think you are trying to give me a heart attack !! I was sitting here testing caching, it worked find until I suddenly had 20 caches instead of 15....guess what "someone" merged a PR.

Just joking, I think I have solved most of the caching issues, except for pre-commit.

alexrudd2:
> Just joking, I think I have solved most of the caching issues, except for pre-commit.

I propose dropping pre-commit from the CI and running the tools manually.  The versions can be kept in sync between `pyproject.toml` and `pre-commit` by `renovatebot`.  [See here](https://github.com/alexrudd2/pymodbus/pull/15) for a PR where it groups them.

janiversen:
@alexrudd2 just thinking loud!

pre-commit really does not bring a lot to this project, another option is to:
- save the precommit hook in e.g githooks
- change documentation to say "cp githooks/precommit ..." instead of "pre-commit install"

And then suddenly we have all dependencies back in pyproject.toml.

What is the disadvantage I have overlooked ?

I mean for this project, with its "huge" developer population.

alexrudd2:
> What is the disadvantage I have overlooked ?

pre-commit has minor value to me because it's easy to transfer between projects and install/uninstall.  I tried using bare git hooks before and they were a bit fussy, but I'm OK using them if you can get them to work.

PR #1830, opened 2023-10-17
Demonstrate old caching behaviour

alexrudd2:
This reverts commit e275017cdaf8671bfddcbf846d9f9881f0ac0950.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
         id: cache-venv
         uses: syphar/restore-virtualenv@v1
         with:
-          requirement_files: pyproject.toml
+          requirement_files: [ pyproject.toml, pre-commit-config.yaml ]
 
       - name: venv create
         if: steps.cache-venv.outputs.cache-hit != 'true'
@@ -101,7 +101,7 @@ jobs:
         id: cache-venv
         uses: syphar/restore-virtualenv@v1
         with:
-          requirement_files: pyproject.toml
+          requirement_files: [ pyproject.toml, pre-commit-config.yaml ]
 
       - name: venv create
         if: steps.cache-venv.outputs.cache-hit != 'true'


janiversen:
What do you want to demonstrate, I have already said a couple of times that the change to python-checkout is a stepping stone.

janiversen:
Your understand is not totally correct...I want to have a set of caches (15 in total os´es combined with python's). With the old version we had 15 pr PR and the current version is even more bad.

Only 15 should be shared among all pull requests and other CI runs...and of course also within the jobs. That will really lower the clock time used for each pull request.

Now with the experience I can combine the 2. In reality what I need to do is a number of things:
- generate proper cache keys
- ensure the path to venv is stable (e.g. windows have a different path)
- include pre-commit in venv (in the old version it installed everytime)

So you see it is a bit more than "just".

alexrudd2:
>     * generate proper cache keys
>     * [that are shared across pull requests]

Done.  The setup below has 15 different caches (3 OS x 5 Python), and I think they are shared across PRs.
```yaml
      - name: venv restore
        id: cache-venv
        uses: syphar/restore-virtualenv@v1
        with:
          requirement_files: pyproject.toml
```
>     * ensure the path to venv is stable (e.g. windows have a different path)

Argh, damn it Windows!

>     * include pre-commit in venv (in the old version it installed everytime)

Yes, I noticed this and am thinking of solutions.  I think it might be possible to jam pre-commit into the cached venv with `PRE_COMMIT_HOME` envvar.

janiversen:
> ```yaml
>       - name: venv restore
>         id: cache-venv
>         uses: syphar/restore-virtualenv@v1
>         with:
>           requirement_files: pyproject.toml
> ```

That is what we had in the old solution, and that creates 15 new caches with each pull request.

The problem is that it creates a cache if there are no hit in the branch, solution is to define it as readonly "lookup: true", and then save the caches somewhere else.

> Yes, I noticed this and am thinking of solutions. I think it might be possible to jam pre-commit into the cached venv with `PRE_COMMIT_HOME` envvar.

That sounds like a good solution, I will try that.

alexrudd2:
> That is what we had in the old solution, and that creates 15 new caches with each pull request.
> The problem is that it creates a cache if there are no hit in the branch [...]

Where did you see this behaviour?  I have the same setup on one of my other repositories and it shares the cache.  (Sadly it's a private repo so I can't share). 


I am rebasing and force-pushing to hope to get a clean CI run (damn flaky tests)

janiversen:
Hmmm thought I left a response.

I saw it before I changed CI, every PR had its set of cache.

You can follow it here:
https://github.com/pymodbus-dev/pymodbus/actions/caches

Look who is "owning" the cache. Caches owned by "dev" can (depending on how cache is invoked) be shared with other branches...but the default behaviour of cache is to make caches for each branch.

alexrudd2:
> Look who is "owning" the cache. 

On my (sadly private) repo with the cache key being `requirements.txt`, the cache owner is `master` if there are only code changes.  The cache owner is the PR merge branch if there are dependency changes.

I'm not sure what's happening here, since the system I understand (i.e. this PR) has had the caches purged, and the new runs keep failing :S

alexrudd2:
Closing in favor of experimentation in my fork: https://github.com/alexrudd2/pymodbus/pull/14

PR #1831, opened 2023-10-17
Consistent messages if imports fail

alexrudd2:
Supercedes https://github.com/pymodbus-dev/pymodbus/pull/1827

This PR makes consistent error messages when an import (either `aiohttp` or `pyserial`) is missing, warning the user they will need to install the appropriate package.


janiversen (approved):
LGTM. Would be nice if you do REPL later.

PR #1834, opened 2023-10-17
Use other dependency groups for 'all'

alexrudd2:
Don't Repeat Yourself :)


alexrudd2:
Hahaha that's why it's good we both try different things.  (Thanks to https://stackoverflow.com/a/75503961)

I did test locally first; it works fine.

PR #1835, opened 2023-10-17
Try out renovatebot

alexrudd2:

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,23 @@
+{
+  "extends": [
+    "config:recommended"
+  ],
+  "pre-commit": {
+    "enabled": true
+  },
+  "rangeStrategy": "bump",
+  "packageRules": [
+    {
+      "matchManagers": ["github-actions"],
+      "automerge": true
+    },
+    {
+      "matchPackagePatterns": ["mypy", "pre-commit", "ruff"],
+      "automerge": true
+    },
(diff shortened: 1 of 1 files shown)


alexrudd2:
@janiversen If you agree, you'll need to install the bot here: https://github.com/marketplace/renovate

alexrudd2:
you can see it in action on my fork: https://github.com/alexrudd2/pymodbus/pulls

(The amount and type of PRs can be configured)

janiversen:
As far as I can see, this bot is not relevant, with all configurations in pyproject.toml.

alexrudd2:
The bot reads `pyproject.toml` and submits PRs to update dependencies.

https://github.com/alexrudd2/pymodbus/pulls
Screenshot 2023-10-27 at 9 36 02 AM

It also constantly updates an overview of them: https://github.com/alexrudd2/pymodbus/issues/17

janiversen:
the GitHub variant does the same without being depending on an external site.

alexrudd2:
> the GitHub variant does the same without being depending on an external site.

I've used both, and prefer Renovatebot since it's much more powerful and configurable.  That said, I'm OK with using dependabot too.

janiversen:
My preference is clearly dependabot, but we do not really need any of them, we have no reason to always use the newest version of the dependencies.

PR #1836, opened 2023-10-17
Re-enable multiple cores with xdist

alexrudd2:
Although the GitHub hosted runners are only 2-cores, there's still an improvement with `pytest-xdist`.

See comments here: https://github.com/pymodbus-dev/pymodbus/pull/1829#issuecomment-1766988198
There are some recently flaky tests, so those should be fixed before re-enabling `xdist`


janiversen:
This is done in an already merged PR.

PR #1837, opened 2023-10-17
Remove dummy complete job

alexrudd2:
I do not think the dummy `ci_complete` job is necessary.  Github's branch protection rules should automatically disallow merging if any of the jobs fail, without needing this dummy.


janiversen (commented):
Unless something have changed, this is still needed.

The branch rules did not work before I added this job, and I think you were the one who pointed me in the direction.

alexrudd2:
Damn, you're right (since we're using a matrix, and therefore the status names will change)

PR #1838, opened 2023-10-17
Switch from black to ruff-format

alexrudd2:
Note that black has a default `line-length: 88` which differed from the `110` which was used (but ignored) for `ruff`.


janiversen (approved):
LGTM, thanks

alexrudd2:
OK, if it is too noisy we can revert.

PR #1841, opened 2023-10-17
Eliminate implicit optional in simulator

alexrudd2:
This eliminates another 3 implicit optional in the simulator.

Since we're already doing `from __future__ import annotations`, also eliminate

```
from typing import List
List[x]
```
in favor of just `list[x]`


janiversen (approved):
LGTM, thanks

Issue #1821, opened 2023-10-16
ModbusClientMixin type annotations incompatibles with the async client


jvelo:
### Versions

- Python: 3.12
- OS: macos
- Pymodbus: 3.5.4
- Mypy : 1.5.1

### Pymodbus Specific

- Client: async

### Description

Mypy fails to validate standard async client calls due to incompatible type annotations in ModbusClientMixin.

### Code and Logs

Following code

```python
client = AsyncModbusTcpClient(HOST)
await client.connect()
coils = await client.read_coils(address=0, count=1, slave=SLAVE_ID)
```

Makes mypy fails with:

```
__main__.py:18: error: Incompatible types in "await" (actual type "ModbusResponse", expected type "Awaitable[Any]")
```

janiversen:
that because we use the same method for both sync and async, which is actually very normal.

We have configured mypy to the checks we want, you might want to do something similar.

janiversen:
If I may suggest, make a pull request where you have just changed a couple....just to secure that CI accept your changes, and then add a commit with all calls.

jvelo:
Sounds good, I'll have a try at it

jvelo:
After a little back and force, looks like it works for `mypy`, but still there's pytest failing for one version of python (3.10) – not sure what to make of it and if it's caused by the change itself : https://github.com/pymodbus-dev/pymodbus/actions/runs/6576303048/job/17865375534

PR #1842, opened 2023-10-19
Try and see if CI will accept changes proposed in #1821

jvelo:
As per https://github.com/pymodbus-dev/pymodbus/issues/1821#issuecomment-1764903223

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import struct
+from collections.abc import Awaitable
 from enum import Enum
 from typing import Any
 
@@ -47,7 +48,7 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
     def __init__(self):
         """Initialize."""
 
-    def execute(self, request: ModbusRequest) -> ModbusResponse:
+    def execute(self, request: ModbusRequest) -> ModbusResponse | Awaitable[ModbusResponse]:
         """Execute request (code ???).
 
         :param request: Request to send


janiversen:
That is often a fluke, I have restarted it.

janiversen:
#@alexrudd2 I highly disagree with you, we will maintain duplicated code just to please mypy, especially not when we have a valid alternative.

You can find the same approach in many standard libraries. This is totally legal and not against any PEP. Remark the function just passed the future it does not create it. The future is created in a method properly marked with async.

janiversen:
@jvelo thanks for the PR, to me it an excellent compromise, it does not cause duplicate code, it violates no PEP and it silences mypy.

"async' is demanded for functions that use asyncio functionality.

jvelo:
There's still one aspect we might have overlooked:  verification of synchronous code where previously code that would expect `ModbusResponse` might not accept the new union type (while for async code it's OK since the `await` will "unbox" the Awaitable) ...
Not sure how other libraries deal with this, but I can try and look into it

jvelo:
I confirm that the change I've proposed is not such an improvement over the status quo, since now both sync and async clients cannot make assumption over which side of the union will be returned :/ Sorry to have rushed this, I though it would be simpler than it is.
I tried to play with `@overload` too but to no avail, it doesn't seem to support overloading with the same signature just one async the other not.
Right now I can't see an elegant alternative that would avoid code duplication. Maybe the better tradeoff would be to have a `Protocol` for async with the async signatures, so that client code that use an async version can box to this protocol.

alexrudd2:
> There's still one aspect we might have overlooked: verification of synchronous code where previously code that would expect `ModbusResponse` might not accept the new union type 

Yes, you've articulated the problem I wasn't able to express well :)

alexrudd2:
> @jvelo thanks for the PR, to me it an excellent compromise, it does not cause duplicate code, it violates no PEP and it silences mypy.

It does not silence `mypy`.  There are still errors if you use `--check-untyped-defs`.  All this PR has done is move the location :)

janiversen:
I suspect we end up writing a generator that generates all functions as sync/async, that is a solution acceptable because it still have a low maintenance point.

Basically it could be done with "sed" search for "def" and change to "async def", however the problem is to automate this,

alexrudd2:
Ideally it should be possible to write one wrapper (decorator perhaps?) instead of changing all functions.  I tried earlier but could not figure it out.

PR #1846, opened 2023-10-19
Convert type hints to PEP585

alexrudd2:
Use `from __future__ import annotations` for Python3.8 support.

Switch e.g. `List[int] to list[int]` per PEP585
Replace `Union` and `Optional` with `|`


janiversen (approved):
LGTM

PR #1850, opened 2023-10-24
Fixed stopbits docstring typo.

### LINKED ISSUE #1849: Invalid `stopbits=0` for serial connection
In the following [docstring](https://github.com/pymodbus-dev/pymodbus/blob/8d3e36fe78ed5ec8bcf94b024df1125ee2b1f015/pymodbus/client/serial.py#L121) valid values for stop bits is 0-2:
```
:param stopbits: Number of stop bits 0-2.
```
Is this a typo? I don't think you can have a serial connection without stop bits. Also I don't think the underlying serial library [`pyserial`](https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.STOPBITS_ONE) supports no stop bits !
```
Stop bits

serial.STOPBITS_ONE
serial.STOPBITS_ONE_POINT_FIVE
serial.STOPBITS_TWO
```
I tried setting `stopbits=0` and got the following **expected** error:
```
ValueError: Not a valid stop bit size: 0
```

I propose if this needs to be fixed, instead of changing `0-2` to `1-2`, it is better to have explicit valid values like `1, 1.5, 2`.

farzadpanahi:
Fixes #1849 
`stopbits=0` is not valid and should be removed from `doctsring`.

### DIFF
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -34,7 +34,7 @@ class AsyncModbusSerialClient(ModbusBaseClient, asyncio.Protocol):
     :param baudrate: Bits per second.
     :param bytesize: Number of bits per byte 7-8.
     :param parity: 'E'ven, 'O'dd or 'N'one
-    :param stopbits: Number of stop bits 0-2.
+    :param stopbits: Number of stop bits 1, 1.5, 2.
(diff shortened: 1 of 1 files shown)


janiversen (approved):
LGTM, thanks

PR #1852, opened 2023-10-25
Add more return types to transport

alexrudd2:


janiversen (approved):
There was a reason why I did not do most of these, I think it was a problem with mypy and the base class, but it seems gone.

PR #1856, opened 2023-10-26
Eliminate remaining implicit optional

alexrudd2:
In most cases, expanding the type to include `| None` is all that was required.

In others, I needed to add a simple check: `AssertionError` or similar.

I have split the changes into 3 commits, in case some are too much for one PR.

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -181,35 +181,30 @@ class ModbusProtocol(asyncio.BaseProtocol):
             host = self.comm_params.host
             port = int(self.comm_params.port)
         if self.comm_params.comm_type == CommType.SERIAL:
-            host, port = self.init_setup_serial(host, port)
-            if not host and not port:
+            if NULLMODEM_HOST in host:
+                host, port = NULLMODEM_HOST, int(host[9:].split(":")[1])
+            elif self.is_server and host.startswith("socket"):
+                # format is "socket://<host>:port"
+                self.comm_params.comm_type = CommType.TCP
+                parts = host.split(":")
+                host, port = parts[1][2:], int(parts[2])
+            else:
+                self.call_create = lambda: create_serial_connection(
+                    self.loop,
+                    self.handle_new_connection,
+                    host,
+                    baudrate=self.comm_params.baudrate,
+                    bytesize=self.comm_params.bytesize,
+                    parity=self.comm_params.parity,
+                    stopbits=self.comm_params.stopbits,
+                    timeout=self.comm_params.timeout_connect,
+                )
                 return
         if host == NULLMODEM_HOST:
             self.call_create = lambda: self.create_nullmodem(port)
             return
         # TCP/TLS/UDP
         self.init_setup_connect_listen(host, port)
-
-    def init_setup_serial(self, host: str, _port: int) -> tuple[str, int]:
-        """Split host for serial if needed."""
-        if NULLMODEM_HOST in host:
-            return NULLMODEM_HOST, int(host[9:].split(":")[1])
-        if self.is_server and host.startswith("socket"):
-            # format is "socket://<host>:port"
-            self.comm_params.comm_type = CommType.TCP
-            parts = host.split(":")
-            return parts[1][2:], int(parts[2])
-        self.call_create = lambda: create_serial_connection(
-            self.loop,
-            self.handle_new_connection,
-            host,
-            baudrate=self.comm_params.baudrate,
-            bytesize=self.comm_params.bytesize,
-            parity=self.comm_params.parity,
-            stopbits=self.comm_params.stopbits,
-            timeout=self.comm_params.timeout_connect,
-        )
-        return None, None
 
     def init_setup_connect_listen(self, host: str, port: int) -> None:
         """Handle connect/listen handler."""
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -266,9 +266,11 @@ class ModbusProtocol(asyncio.BaseProtocol):
             self.loop = asyncio.get_running_loop()
         self.is_closing = False
         try:
-            self.transport = await self.call_create()
-            if isinstance(self.transport, tuple):
-                self.transport = self.transport[0]
+            _transport = await self.call_create()
+            if isinstance(_transport, tuple):
+                self.transport = _transport[0]
+            else:
+                self.transport = _transport
         except OSError as exc:
             Log.warning("Failed to start server {}", exc)
             # self.transport_close(intern=True)
@@ -435,7 +437,7 @@ class ModbusProtocol(asyncio.BaseProtocol):
 
     def reset_delay(self) -> None:
         """Reset wait time before next reconnect to minimal period."""
-        self.reconnect_delay_current = self.comm_params.reconnect_delay
+        self.reconnect_delay_current = self.comm_params.reconnect_delay or 0
 
     def is_active(self) -> bool:
         """Return true if connected/listening."""
@@ -469,6 +471,12 @@ class ModbusProtocol(asyncio.BaseProtocol):
 
     async def do_reconnect(self) -> None:
         """Handle reconnect as a task."""
+        if not (
+            self.comm_params.reconnect_delay and self.comm_params.reconnect_delay_max
+        ):
+            raise AssertionError(
+                "do_reconnect should not be called if reconnect_delay is None"
+            )
         try:
             self.reconnect_delay_current = self.comm_params.reconnect_delay
             while True:
@@ -519,7 +527,7 @@ class NullModem(asyncio.DatagramTransport, asyncio.Transport):
         asyncio.DatagramTransport.__init__(self)
         asyncio.Transport.__init__(self)
         self.protocol: ModbusProtocol = protocol
-        self.other_modem: NullModem = None
+        self.other_modem: NullModem | None = None
         self.listen = listen
         self.manipulator: Callable[[bytes], list[bytes]] | None = None
         self._is_closing = False
@@ -605,6 +613,8 @@ class NullModem(asyncio.DatagramTransport, asyncio.Transport):
 
     def write(self, data: bytes) -> None:
         """Send data."""
+        if not self.other_modem:
+            raise AssertionError("Missing other_modem")
         if not self.manipulator:
             self.other_modem.protocol.data_received(data)
             return
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -156,10 +156,14 @@ class ModbusProtocol(asyncio.BaseProtocol):
         self.is_server = is_server
         self.is_closing = False
 
-        self.transport: asyncio.BaseTransport = None
-        self.loop: asyncio.AbstractEventLoop = None
+        self.loop: asyncio.AbstractEventLoop = None  # type: ignore[assignment]
+        self.transport: asyncio.BaseTransport | None = None
         self.recv_buffer: bytes = b""
-        self.call_create: Callable[[], Coroutine[Any, Any, Any]] = lambda: None
+
+        async def _noop():
+            ...
+
+        self.call_create: Callable[[], Coroutine[Any, Any, Any]] = _noop
         if self.is_server:
             self.active_connections: dict[str, ModbusProtocol] = {}
         else:
@@ -398,11 +402,11 @@ class ModbusProtocol(asyncio.BaseProtocol):
             self.sent_buffer += data
         if self.comm_params.comm_type == CommType.UDP:
             if addr:
-                self.transport.sendto(data, addr=addr)  # type: ignore[attr-defined]
+                self.transport.sendto(data, addr=addr)  # type: ignore[union-attr]
             else:
-                self.transport.sendto(data)  # type: ignore[attr-defined]
+                self.transport.sendto(data)  # type: ignore[union-attr]
         else:
-            self.transport.write(data)  # type: ignore[attr-defined]
+            self.transport.write(data)  # type: ignore[union-attr]
 
     def transport_close(self, intern: bool = False, reconnect: bool = False) -> None:
         """Close connection.
--- a/pymodbus/transport/transport_serial.py
+++ b/pymodbus/transport/transport_serial.py
@@ -20,7 +20,7 @@ class SerialTransport(asyncio.Transport):
         super().__init__()
         self.async_loop = loop
         self._protocol: asyncio.BaseProtocol = protocol
-        self.sync_serial = serial.serial_for_url(*args, **kwargs)
+        self.sync_serial: serial.Serial | None = serial.serial_for_url(*args, **kwargs)
         self._write_buffer: list[bytes] = []
         self.poll_task: asyncio.Task | None = None
         self._poll_wait_time = 0.0005
@@ -59,12 +59,12 @@ class SerialTransport(asyncio.Transport):
     def write(self, data) -> None:
         """Write some data to the transport."""
         self._write_buffer.append(data)
-        if not self.poll_task:
+        if not self.poll_task and self.sync_serial:
             self.async_loop.add_writer(self.sync_serial.fileno(), self._write_ready)
 
     def flush(self) -> None:
         """Clear output buffer and stops any more data being written."""
-        if not self.poll_task:
+        if not self.poll_task and self.sync_serial:
             self.async_loop.remove_writer(self.sync_serial.fileno())
         self._write_buffer.clear()
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -187,7 +187,7 @@ overgeneral-exceptions = "builtins.Exception"
 bad-functions = "map,input"
 
 [tool.mypy]
-strict_optional = false
+strict_optional = true
 show_error_codes = true
 local_partial_types = true
 strict_equality = true


pymodbus/transport/transport.py
```diff
@@ -177,36 +181,32 @@ def __init__(
...
             port = int(self.comm_params.port)
         if self.comm_params.comm_type == CommType.SERIAL:
-            host, port = self.init_setup_serial(host, port)
-            if not host and not port:
+            if NULLMODEM_HOST in host:
+                host, port = NULLMODEM_HOST, int(host[9:].split(":")[1])
+            elif self.is_server and host.startswith("socket"):
+                # format is "socket://<host>:port"
+                self.comm_params.comm_type = CommType.TCP
+                ## FIXME: does this make sense??  # pylint: disable=fixme
```
alexrudd2, line 189: This doesn't make sense to me.  Why would `self.comm_params.comm_type == CommType.Serial` and then be changed to `CommType.TCP`?!

alexrudd2: https://github.com/pymodbus-dev/pymodbus/pull/1710/files#diff-232ea64eea656684ba1d20bc100847b5fabad3d4627d4b8a51c49041ad84897dR30-R31

```
Pyserial allow the comm_port to be a socket e.g. "socket://localhost:502",
this allows serial clients to connect to a tcp server with RTU framer.
```

janiversen: Yes for clients not for servers.

pymodbus/client/serial.py
```diff
@@ -184,6 +183,8 @@ def __init__(
 
         self.last_frame_end = None
 
+        if not self.comm_params.baudrate:
```
janiversen, line 186: Actually it does not if you use "socket:". 

Apart from that if we catch baud rate, then why not stop bits....all these have a default value.

pymodbus/transport/transport.py
```diff
@@ -156,17 +156,21 @@ def __init__(
         self.is_server = is_server
         self.is_closing = False
 
-        self.transport: asyncio.BaseTransport = None
-        self.loop: asyncio.AbstractEventLoop = None
+        self.loop: asyncio.AbstractEventLoop = None  # type: ignore[assignment]
```
janiversen, line 159: This is why I do not like the "| None", sometimes you use it, other times not...that its not really intuitive !

In my mind mypy is a bit stupid in this respect !

pymodbus/transport/transport.py
```diff
@@ -156,17 +156,21 @@ def __init__(
...
+        self.transport: asyncio.BaseTransport | None = None
```
janiversen, line 160: And here you use it....any maintainer will think why on earth is self.loop without None and this one is with None...both are later used in the same way.

pymodbus/transport/transport.py
```diff
@@ -156,17 +156,21 @@ def __init__(
...
         if self.is_server:
             self.active_connections: dict[str, ModbusProtocol] = {}
         else:
-            self.listener: ModbusProtocol = None
+            self.listener: ModbusProtocol | None = None
             self.unique_id: str = str(id(self))
-            self.reconnect_task: asyncio.Task = None
-            self.reconnect_delay_current: float = 0.0
+            self.reconnect_task: asyncio.Task | None = None
+            self.reconnect_delay_current: float | None = 0.0
```
janiversen, line 173: this cannot be None as far as I remember the code... 0.0 is not None.

pymodbus/transport/transport.py
```diff
@@ -177,36 +181,32 @@ def __init__(
             host = self.comm_params.host
             port = int(self.comm_params.port)
         if self.comm_params.comm_type == CommType.SERIAL:
-            host, port = self.init_setup_serial(host, port)
-            if not host and not port:
+            if NULLMODEM_HOST in host:
```
janiversen, line 184: This doesn't make sense to me...you split something that will be split later ?

pymodbus/transport/transport.py
```diff
@@ -397,11 +399,11 @@ def transport_send(self, data: bytes, addr: tuple = None) -> None:
             self.sent_buffer += data
         if self.comm_params.comm_type == CommType.UDP:
             if addr:
-                self.transport.sendto(data, addr=addr)  # type: ignore[attr-defined]
+                self.transport.sendto(data, addr=addr)  # type: ignore[union-attr]
```
janiversen, line 402: This looks like something that can be avoided !

pymodbus/transport/transport.py
```diff
@@ -469,6 +471,12 @@ def handle_new_connection(self) -> ModbusProtocol:
 
     async def do_reconnect(self) -> None:
         """Handle reconnect as a task."""
+        if not (
```
janiversen, line 474: We try, in our internal functions, not to have
- exceptions due to bad calling.
- I believe we always call do_reconnect, and it then decides what actually to do.

pymodbus/transport/transport_serial.py
```diff
@@ -58,12 +58,12 @@ def close(self, exc=None):
...
         """Write some data to the transport."""
         self._write_buffer.append(data)
-        if not self.poll_task:
+        if not self.poll_task and self.sync_serial:
             self.async_loop.add_writer(self.sync_serial.fileno(), self._write_ready)
 
     def flush(self) -> None:
         """Clear output buffer and stops any more data being written"""
-        if not self.poll_task:
+        if not self.poll_task and self.sync_serial:
```
janiversen, line 66: same here.

janiversen (changes requested):
Quite a number of questions/comments.

I think a lot of it comes because you change code and add typing at the same time.

pymodbus/transport/transport.py
alexrudd2: Not if `transport = None`.

Perhaps a better answer here is some sort of dummy transport instead of `None`

pymodbus/transport/transport.py
```diff
@@ -435,7 +437,7 @@ def transport_close(self, intern: bool = False, reconnect: bool = False) -> None
 
     def reset_delay(self) -> None:
         """Reset wait time before next reconnect to minimal period."""
-        self.reconnect_delay_current = self.comm_params.reconnect_delay
+        self.reconnect_delay_current = self.comm_params.reconnect_delay or 0
```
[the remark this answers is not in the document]
alexrudd2: My understanding of the API is that `reconnect_delay = None` disables reconnects.

pymodbus/transport/transport.py
alexrudd2: > We try, in our internal functions, not to have exceptions due to bad calling.

Yes, this is why I split all these `AssertionError` into a separate commit.

> I believe we always call do_reconnect, and it then decides what actually to do.

Unfortunately with the various conditionals it's not possible to prove this is always called correctly.  (`mypy` would not complain in that case).  If called incorrectly, it will raise an unhandled `ValueError`

pymodbus/transport/transport.py
```diff
@@ -605,6 +613,8 @@ def sendto(self, data: bytes, _addr: Any = None) -> None:
 
     def write(self, data: bytes) -> None:
         """Send data"""
+        if not self.other_modem:
```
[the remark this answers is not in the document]
alexrudd2: In that case, I'll need to use `type: ignore` instead of runtime checks.

pymodbus/transport/transport.py
alexrudd2: This one was quite difficult because I could not figure out the intended usage of `get_running_loop()`.

pymodbus/transport/transport.py
alexrudd2: The later callers are different.  Calling code checks `self.transport != None` in all places.  But the `loop()` code is more complicated.

pymodbus/transport/transport.py
janiversen: That is a variant, which is acceptable.

pymodbus/transport/transport.py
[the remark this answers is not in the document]
janiversen: Actually at this level it should be 0. this is not a user API.

pymodbus/transport/transport.py
alexrudd2: It's unclear what the intended value of `self.reconnect_delay_current` is when `self.reconnect_delay == None`.

I could not easily decide between `0.0` and `None` in that case, so I went with `None` for fewer code changes.

pymodbus/transport/transport.py
alexrudd2: It didn't make sense to me either, but that's how the existing code works.

I suspect there is further improvement possible here.

pymodbus/transport/transport.py
janiversen: into a separate commit does not help, I do not review your individual commits, that would potentially mean adding review comments you have solved in a later commit.

Apart from that all commits in one PR is merged, so later it is all mixed.

pymodbus/transport/transport.py
```diff
@@ -177,36 +181,32 @@ def __init__(
...
+                    timeout=self.comm_params.timeout_connect,
+                )
                 return
         if host == NULLMODEM_HOST:
             self.call_create = lambda: self.create_nullmodem(port)
             return
         # TCP/TLS/UDP
         self.init_setup_connect_listen(host, port)
 
-    def init_setup_serial(self, host: str, _port: int) -> tuple[str, int]:
```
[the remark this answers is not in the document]
alexrudd2: I think it's possible we can simplify the code further.  Having a separate function that returned either `host, port` or `None, None` made no sense to me.

pymodbus/transport/transport.py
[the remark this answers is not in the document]
janiversen: ??? self.other_modem is a reference to the other part. The object is created as part of the connect process.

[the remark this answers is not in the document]
alexrudd2: I understand what `self.other_modem` is.  However, there is currently no hard guarantee that this function is not called with `self.other_modem == None`.

Options:
(1) add this check
(2) `type: ignore` and trust all calling code to be correct (such a "soft" guarantee will not please mypy)
(3) refactor the object/methods so `write()` can only be called after connect (?!)

alexrudd2:
> I think a lot of it comes because you change code and add typing at the same time.

The final changes are of course the most difficult.  :)  It's not possible to prove to `mypy` all callers are correct without such code changes.

pymodbus/transport/transport_serial.py
```diff
@@ -58,12 +58,12 @@ def close(self, exc=None):
     def write(self, data) -> None:
         """Write some data to the transport."""
         self._write_buffer.append(data)
-        if not self.poll_task:
+        if not self.poll_task and self.sync_serial:
```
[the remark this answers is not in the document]
alexrudd2: "cannot" != "do not"

To satisfy `mypy` it must be totally impossible to call incorrectly.

pymodbus/transport/transport.py
janiversen: I know I can do that...but could we please keep 1 PR with the type changes that are really needed, which I am most likely to approve without a lot of comments, and a second PR with the code changes, where I am likely to have a lot more opinions.

I might be old-fashioned, but I see a PR as a unit, I have never cared about the single commits....the reason being that when I make comments the author add commits etc....and the only thing that counts is the sum of all the commits,  because that is what will be merged.

pymodbus/transport/transport.py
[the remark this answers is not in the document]
janiversen: The hard guarantee is that it is our own code !

(1) is not really the way we want to go !
(2) this is why I do not like mypy !
(3) but that is already the fact !, that class is instantiated by the connect call....the problem is that mypy sees this as an independent class which it is NOT.

Mypy needs to learn (and I have absolutely no idea how) that some class can only be instantiated in a context, and not randomly. The solution might be to secure that other_modem can never be null (like you suggested for transport).

I do believe that mypy helps a lot in making the code more stable, but I refuse to make "bad" code just because mypy does not understand e.g. that some classes cannot be used randomly.

To put it in a positive light, I approve any type change you make without a second thought, including small non essential code changes !! but when it affects the running code like asserting due to a programming error on our side, I am strongly opposed. 

I think you do a very good and important job, so please do not see my comments as solely negative !

pymodbus/transport/transport_serial.py
```diff
@@ -35,7 +35,7 @@ def setup(self):
             self.async_loop.add_reader(self.sync_serial.fileno(), self._read_ready)
         self.async_loop.call_soon(self._protocol.connection_made, self)
 
-    def close(self, exc=None):
+    def close(self, exc=None) -> None:
```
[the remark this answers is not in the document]
janiversen: exc is not typed.

pymodbus/transport/transport_serial.py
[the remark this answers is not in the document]
janiversen: this is a basic problem of mypy, it assumes all classes can be used randomly, which is NOT correct...and in reality one of my basic problems with mypy.

janiversen:
> > I think a lot of it comes because you change code and add typing at the same time.
> 
> The final changes are of course the most difficult. :) It's not possible to prove to `mypy` all callers are correct without such code changes.

Well that is your challenge, because we will not start to add code, that ensures that we call the methods correctly...that would and endless amendment of code that needs to be maintained and basically only slows down the library.

janiversen:
May I politely suggest you isolate the type changes in another PR which would be approved fast, and leave the rest in this PR:

janiversen:
Please accept, that I do not want to be negative !!! you do a lot good for the library, for which I am grateful, but mypy have some severe limitations (or assumptions) which do not work well with real life code, and together we need to find a compromise that do not add extra maintenance just to please mypy.

alexrudd2:
> May I politely suggest you isolate the type changes in another PR which would be approved fast, and leave the rest in this PR:

Yep, I'll do that.

> together we need to find a compromise that do not add extra maintenance just to please mypy.

Yes, of course.  Partly I opened this PR so we can discuss and find those compromises.  (The easy ones already got merged)

pymodbus/client/serial.py
alexrudd2: Solved by https://github.com/pymodbus-dev/pymodbus/pull/1872

alexrudd2:
Rebased + force-pushed to reduce the diff and show the remaining problems.

pymodbus/transport/transport.py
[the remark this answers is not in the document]
alexrudd2: Split out into https://github.com/pymodbus-dev/pymodbus/pull/1866

pymodbus/transport/transport.py
alexrudd2: Split out into https://github.com/pymodbus-dev/pymodbus/pull/1866

pymodbus/transport/transport.py
alexrudd2: I tried the dummy transport; it didn't work well.

pymodbus/transport/transport.py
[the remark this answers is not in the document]
alexrudd2: https://github.com/pymodbus-dev/pymodbus/pull/1874

pymodbus/transport/transport.py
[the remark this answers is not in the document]
alexrudd2: https://github.com/pymodbus-dev/pymodbus/pull/1873

pymodbus/transport/transport.py
janiversen: self.reconnect_delay_current is a value between self.reconnect_delay and self.reconnect_delay_max. so setting it to self.reconnect_delay is probably the most generic way.

janiversen:
In I think we need to have a relaxed view on mypy and not the 100% which it seems you support. Because we know how our code works, there are a lot of situations where mypy tells this is an error, but due to the way the code is build this only happens if someone misuse the code.

I am all in favour of using mypy where the code actually benefit from it, but e.g. adding code, which we have to maintain or deleting variables are not examples which I consider makes the code better.

alexrudd2:
> In I think we need to have a relaxed view on mypy and not the 100% which it seems you support. Because we know how our code works, there are a lot of situations where mypy tells this is an error, but due to the way the code is build this only happens if someone misuse the code.

The stricter the checks, the more it finds.  [Here's an example](https://github.com/pymodbus-dev/pymodbus/discussions/1803) where `mypy` is currently too lenient and did not find this bug: https://github.com/pymodbus-dev/pymodbus/pull/1804

Mypy is currently at about 50% (because untyped defs are not checked, and other stricter options are not enabled).

janiversen:
I disagree, mypy make one false assumption especially about our internal code, it assumes that all methods can be called anytime....but the internal code have sequences which we know and do not violate......like e.g. self.sync_serial = None in __init__ is just to get it declared when the object is created, no calls will see that later.

While I agree an API must be robust to invalid calls, our internals do not need to be just as robust, just like we in general do not validate parameters in internal calls.

alexrudd2:
> Your example is actually a good exammple to not make mypy stricter, the code works nice in our library but in the example it was used outside it the library.

I think you are missing the point.  OK, the bug was _first_ found by external calling code.  However, there was _also_ an incorrect caller hiding in [pymodbus code](https://github.com/pymodbus-dev/pymodbus/pull/1804/files#diff-5fbeb0dd3ac9ac917954cf4c42aa4edccd8329ae13ee56c56f8fb9af931fbba9R98).  It was not caught by the tests because it's a rarely-used error handler, and not the "happy path".

```python
        except serial.SerialException as exc:
       ---  self._protocol.loop.call_soon(self._call_connection_lost, exc)  # error
       +++  self.async_loop.call_soon(self._call_connection_lost, exc)  # fix
            self.close()
```

janiversen:
No I did not miss the point, but I think you missed the point that this is imported code and currently work in progress to be converted fully to our needs.

alexrudd2:
Closing in favor of https://github.com/pymodbus-dev/pymodbus/pull/1888

PR #1857, opened 2023-10-26
Ignore files downloaded by pytest

alexrudd2:


janiversen:
This is something special in your setup, my different setups do not have that. Do you overwite the pip cache path environment variable?

janiversen:
hmmm....strange as I said it is not happening here.

maybe recreate your venv ?

PR #1859, opened 2023-10-27
More elegant noop

janiversen:
Changed the _noop introduced, with a more elegant one line solution.


janiversen:
@alexrudd2 you were the one who suggested to use "lambda: None" :-) 

so lets see if mypy in CI accepts it.

alexrudd2:
mypy in CI accepts the `lambda: None` because it is not in strict mode.
`mypy` in strict mode will not.  🤣

alexrudd2:
Aha, but we are close to `strict-optional = True`

PR #1860, opened 2023-10-27
Remove pre-commit

janiversen:
There are no real advantage of using pre-commit in our setup.

It is being replaced by a manual git hook (located in githooks)

As a first step .pre-commit-config.yaml is reduced by removing all the duplicate checks.


alexrudd2 (approved):
The `exec`s is unnecessary in the hook script, but this works for me.

PR #1863, opened 2023-10-27
Fix broken server examples

ammarjai:
The current examples for server scripts are not able to communicate back due to bad framer arguments. 

The examples through the helper function that parse the argument only produce a string and is directly passed through the class arguments causing the framer attribute of the server to be a string instead of Framer object. 

This commit fix that problem.

### DIFF
--- a/examples/server_async.py
+++ b/examples/server_async.py
@@ -38,6 +38,7 @@ import logging
 import helper
 
 from pymodbus import __version__ as pymodbus_version
+from pymodbus import Framer
 from pymodbus.datastore import (
     ModbusSequentialDataBlock,
     ModbusServerContext,
@@ -71,7 +72,7 @@ def setup_server(description=None, context=None, cmdline=None):
         # This is because many devices exhibit this kind of behavior (but not all)
         if args.store == "sequential":
             # Continuing, use a sequential block without gaps.
-            datablock = ModbusSequentialDataBlock(0x00, [17] * 100)
+            datablock = ModbusSequentialDataBlock(0x00, [1, 0 , 0] * 100)
         elif args.store == "sparse":
             # Continuing, or use a sparse DataBlock which can have gaps
             datablock = ModbusSparseDataBlock({0x00: 0, 0x05: 1})
@@ -155,7 +156,7 @@ async def run_async_server(args):
             # TBD port=
             address=address,  # listen address
             # custom_functions=[],  # allow custom handling
-            framer=args.framer,  # The framer strategy to use
+            framer=Framer(args.framer),  # The framer strategy to use
             # ignore_missing_slaves=True,  # ignore request to a missing slave
             # broadcast_enable=False,  # treat slave_id 0 as broadcast address,
             # timeout=1,  # waiting time for request to complete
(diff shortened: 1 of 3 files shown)


janiversen:
With what version do you have this problem?

dev is changed on purpose to accept an enum (new way) as well as a framer object (old way).

Could it be you use examples from dev, with e,g. v3.5.4 ? that would produce the problem.

janiversen (changes requested):
This is an incorrect change, framer is allowed to be an enum.

Please verify you use examples from the same version as the pymodbus library.

ammarjai:
It is with pymodbus version 3.5.4 since that is the version I'm using when trying to test my modbus client.

For that one my modification was to import each Framer type to make it work.

ammarjai:
> This is an incorrect change, framer is allowed to be an enum.
> 
> Please verify you use examples from the same version as the pymodbus library.

Yes it is an enum. The original example pass the framer argument as a string literal instead of enum causing error when client tries to communicate with the server. This commit convert the string to the proper enum.

janiversen:
You are using examples from dev, together with pymodbus version 3.5.4

In version 3.5.4, get_commandline returns a framer object, see line 126:

Captura de pantalla 2023-10-27 a las 18 25 36

If you run examples from dev, with the library build in dev, the enum works.

janiversen:
if you look at the examples in dev:

Captura de pantalla 2023-10-27 a las 18 29 30

Line 107-108 show that framer is returned as an enum, which works with the code on dev.

So in short, make sure the version of examples match the version of the library.

examples/server_async.py
```diff
@@ -72,7 +72,7 @@ def setup_server(description=None, context=None, cmdline=None):
         # This is because many devices exhibit this kind of behavior (but not all)
         if args.store == "sequential":
             # Continuing, use a sequential block without gaps.
-            datablock = ModbusSequentialDataBlock(0x00, [1, 0 , 0] * 100)
+            datablock = ModbusSequentialDataBlock(0x00, [17] * 100)
```
janiversen, line 75: Why this ? I am pretty sure you cannot change this without making changes in the test suite as well.

janiversen (changes requested):

examples/server_async.py
janiversen: hmmm seems I was too quick. This change makes sense.

janiversen, line 75: you really got me confused with this change:

v3.5.4 have
´´´
if args.store == "sequential":
            # Continuing, use a sequential block without gaps.
            datablock = ModbusSequentialDataBlock(0x00, [17] * 100)
´´´

current dev have:
´´´
        if args.store == "sequential":
            # Continuing, use a sequential block without gaps.
            datablock = ModbusSequentialDataBlock(0x00, [17] * 100)
´´´

so I am not really sure from where you got the "[1, 0, 0]" ?

ammarjai:
> if you look at the examples in dev:
> 
> Captura de pantalla 2023-10-27 a las 18 29 30
> Line 107-108 show that framer is returned as an enum, which works with the code on dev.
> 
> So in short, make sure the version of examples match the version of the library.

Ah no wonder I can't get it to work. I was reading directly from the 3.5.4 docs and the source provide links directly to dev instead of a tags or branch. Also didn't see the zip file provided in the docs as well where I can get this code.

[3.5.4 docs](https://pymodbus.readthedocs.io/en/v3.5.4/source/examples.html#server-asynchronous)

Yeah retested the server examples with correct version (this time) and has everything working without the need for this commit.

Sorry for wasting a bit of time.

janiversen (changes requested):

examples/server_async.py
ammarjai: Yeah that one is a wrong commit that should be squash with the next fixup. I don't think this commit is necessary anymore after retesting everything with the correct version.

PR #1866, opened 2023-10-27
Rework host/port and listener setup

alexrudd2:
The refactored code is less nested, so I hope it is easier to understand.

It also solves a `mypy` error.

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -176,9 +176,9 @@ class ModbusProtocol(asyncio.BaseProtocol):
         else:
             host = self.comm_params.host
             port = int(self.comm_params.port)
-        if NULLMODEM_HOST in host:
-            if self.comm_params.comm_type == CommType.SERIAL:
-                host, port = NULLMODEM_HOST, int(host[9:].split(":")[1])
+        if self.comm_params.comm_type == CommType.SERIAL and NULLMODEM_HOST in host:
+            host, port = NULLMODEM_HOST, int(host[9:].split(":")[1])
+        if host == NULLMODEM_HOST:
             self.call_create = lambda: self.create_nullmodem(port)
             return
         if (
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -188,8 +188,8 @@ class ModbusProtocol(asyncio.BaseProtocol):
         ):
             # format is "socket://<host>:port"
             self.comm_params.comm_type = CommType.TCP
-            parts = host.strip("socket://").split(":")
-            host, port = parts[0], int(parts[1])
+            parts = host.split(":")
+            host, port = parts[1][2:], int(parts[2])
         self.init_setup_connect_listen(host, port)
 
     def init_setup_connect_listen(self, host: str, port: int) -> None:
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -188,8 +188,7 @@ class ModbusProtocol(asyncio.BaseProtocol):
         ):
             # format is "socket://<host>:port"
             self.comm_params.comm_type = CommType.TCP
-            host = host.strip("socket://")
-            parts = host.split(":")
+            parts = host.strip("socket://").split(":")
             host, port = parts[0], int(parts[1])
         self.init_setup_connect_listen(host, port)
 
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -176,39 +176,37 @@ class ModbusProtocol(asyncio.BaseProtocol):
         else:
             host = self.comm_params.host
             port = int(self.comm_params.port)
-        if self.comm_params.comm_type == CommType.SERIAL:
-            host, port = self.init_setup_serial(host, port)
-            if not host and not port:
-                return
-        if host == NULLMODEM_HOST:
+        if NULLMODEM_HOST in host:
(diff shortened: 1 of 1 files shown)


alexrudd2:
See https://github.com/pymodbus-dev/pymodbus/discussions/1865.

If the test that creates `host == socket://__pymodbus_nullmodem:7636` is not valid, this can be simplified further.

janiversen:
> See #1865.
> 
> If the test that creates `host == socket://__pymodbus_nullmodem:7636` is not valid, this can be simplified further.

That is a valid specification, the definition is "socket://<host>:<port>" for serial port as sockets.

alexrudd2:
> You are touching a very delicate part of the code.
> 
> Some of your does not seem correct at a first look,but might very well be correct.

It will remain a secret, how many test failures I had in my attempts 🤣

pymodbus/transport/transport.py
```diff
@@ -176,39 +176,37 @@ def __init__(
         else:
             host = self.comm_params.host
             port = int(self.comm_params.port)
-        if self.comm_params.comm_type == CommType.SERIAL:
-            host, port = self.init_setup_serial(host, port)
-            if not host and not port:
-                return
-        if host == NULLMODEM_HOST:
+        if NULLMODEM_HOST in host:
```
janiversen, line 179: this allows "NULLMODEM_HOST:7109" for all commtype...and it is only defined for serial.

Not sure it is a big problem, but at least a concern.

pymodbus/transport/transport.py
```diff
@@ -176,39 +176,37 @@ def __init__(
...
-            return NULLMODEM_HOST, int(host[9:].split(":")[1])
-        if self.is_server and host.startswith("socket"):
+        if (
+            self.comm_params.comm_type == CommType.SERIAL
+            and self.is_server
+            and host.startswith("socket")
+        ):
             # format is "socket://<host>:port"
             self.comm_params.comm_type = CommType.TCP
+            host = host.strip("socket://")
```
janiversen, line 191: the old code was more compact, and did the job, so why change it ?

janiversen (changes requested):

pymodbus/transport/transport.py
alexrudd2: clarity

pymodbus/transport/transport.py
alexrudd2: This is functionally possible in the old/existing code as well:
https://github.com/pymodbus-dev/pymodbus/blob/6b1ee62b34c50f4e8f3f8c9a9b42e02ff2f8ffe4/pymodbus/transport/transport.py#L183-L185

It's certainly possible that was unintended, in which case the refactor exposed a bug :)

pymodbus/transport/transport.py
janiversen: well that is not more clear to me...now I have to make sure nothing references host after it is changed.

alexrudd2: Would you prefer below?
```python
parts = host.strip("socket://").split(":")
```

pymodbus/transport/transport.py
janiversen: no the old code checked for exact match, those lines would not accept "NULLMODEM:9657"

alexrudd2: d'oh, you're right

pymodbus/transport/transport.py
janiversen: That is at least still a one liner, but I fail to understand what is wrong with the original code, that only indexed the string.

I am not in favor of making changes just for the sake of changing... months from now when bisecting you will wonder if there was a reason why it was changed and if the old code was faulty.

janiversen (approved):
LGTM.

alexrudd2:
The reason I tried changing things here to was to avoid this inelegant return of `None, None`.

```python
        host, port = self.init_setup_serial(host, port)
        if not host and not port:
            return

def init_setup_serial(self, host: str, _port: int) -> tuple[str, int]:
...
return None, None
```

Perhaps you have a cleaner way.

janiversen:
> The reason I tried changing things here to was to avoid this inelegant return of `None, None`.
> 
> ```python
> if self.comm_params.comm_type == CommType.SERIAL:
>         host, port = self.init_setup_serial(host, port)
>         if not host and not port:
>             return
> 
> def init_setup_serial(self, host: str, _port: int) -> tuple[str, int]:
> ...
> return None, None
> ```
> 
> Perhaps you have a cleaner way.

Your last change looks ok....but you still have the change with strip and split, which I overlooked when I first approved the PR.

pymodbus/transport/transport.py
```diff
@@ -176,39 +176,36 @@ def __init__(
...
-            self.handle_new_connection,
-            host,
-            baudrate=self.comm_params.baudrate,
-            bytesize=self.comm_params.bytesize,
-            parity=self.comm_params.parity,
-            stopbits=self.comm_params.stopbits,
-            timeout=self.comm_params.timeout_connect,
-        )
-        return None, None
+            parts = host.strip("socket://").split(":")
```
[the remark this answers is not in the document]
alexrudd2:
```python
# Alex prefers being explicit
parts = host.strip("socket://").split(":")
host, port = parts[0], int(parts[1])

# jan prefers being compact
parts = host.split(":")
host, port = parts[1][2:], int(parts[2])
```

janiversen (approved):
LGTM, thanks.

PR #1867, opened 2023-10-28
Relax test task/thread checker.

janiversen:


janiversen:
@alexrudd2 I think I got it...but please do me a favour if you get more of these errors, please look in the log (pylint) and copy the last lines (the actual error and 3-4 lines before that) and send it to me.

PR #1870, opened 2023-10-30
Fix ModbusSerialServer super().__init__ Parameter Order

chandlerwr:
In the constructor for `ModbusSerialServer`, the call to `super().__init__` flipped the `"request_tracer"` and `"response_manipulator"` parameters. Passing a `request_tracer` as normal resulted in a `"cannot unpack non-iterable NoneType object"` error when a request was recieved.

To reproduce the same bug, replace the TCP server with a serial server in the `server_hook.py` example.

Fixed by reversing the two parameters in the function call. Also added labels to try and prevent further ordering issues if parameters are changed in the future.

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -522,7 +522,7 @@ class ModbusSerialServer(ModbusBaseServer):
         :param request_tracer: Callback method for tracing
         """
         super().__init__(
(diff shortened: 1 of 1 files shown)


janiversen (approved):
LGTM, thanks.

PR #1871, opened 2023-10-30
Eliminate more implicit optional (client transport)

alexrudd2:
There is no particular reason to set `self.transport = None`, so just ... remove those lines :)

This solves another 4 `mypy` errors.


janiversen (approved):
LGTM (actually I thought I had removed those lines already).

alexrudd2:
> LGTM (actually I thought I had removed those lines already).

Hahaha, I spent a while trying to figure out their purpose :)

PR #1872, opened 2023-10-30
Use baudrate directly to solve mypy errors

alexrudd2:
This solves another 2 `mypy` implicit optional errors

### DIFF
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -183,14 +183,14 @@ class ModbusSerialClient(ModbusBaseClient):
 
         self.last_frame_end = None
 
-        self._t0 = float(1 + bytesize + stopbits) / self.comm_params.baudrate
+        self._t0 = float(1 + bytesize + stopbits) / baudrate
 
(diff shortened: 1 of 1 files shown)


janiversen (commented):
This change is technically correct....but it confuses me as to why it is needed.

Could it be that self.comm_params.baudrate is defined wrongly ?

The reason for the question, is that the code you change will be moved as soon as I start integrating transport into the sync clients.

alexrudd2:
> Could it be that self.comm_params.baudrate is defined wrongly ?

I could not figure out the problem with `self.comm_params.baudrate`, either.

My guess is that mypy is confused with `kwargs.get()`.
https://github.com/pymodbus-dev/pymodbus/blob/8b63dd20918364f197b3b34ded8a49cbde827d76/pymodbus/client/base.py#L94

> the code you change will be moved as soon as I start integrating transport into the sync clients

Well, at least now it is consistent with nearby code.  (e.g. `stopbits` instead of `self.comm_params.stopbits`), so the refactor should be easier?

janiversen (approved):
OK, I have no strong feelings for either one.

PR #1873, opened 2023-10-31
Eliminate more implicit optionals in transport

alexrudd2:
Instead of 
```python
if foo [is not None]:
  foo = None
```

use 
```python
if hasattr(self, 'foo'):
  del foo
```

Also, revert https://github.com/pymodbus-dev/pymodbus/pull/1859

`pymodbus/transport/transport.py:255: error: "None" not callable  [misc]`

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -159,7 +159,11 @@ class ModbusProtocol(asyncio.BaseProtocol):
         self.transport: asyncio.BaseTransport = None
         self.loop: asyncio.AbstractEventLoop = None
         self.recv_buffer: bytes = b""
-        self.call_create: Callable[[], Coroutine[Any, Any, Any]] = lambda: None
+
+        async def _noop():
+            ...
+
+        self.call_create: Callable[[], Coroutine[Any, Any, Any]] = _noop
         if self.is_server:
             self.active_connections: dict[str, ModbusProtocol] = {}
         else:
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -159,7 +159,7 @@ class ModbusProtocol(asyncio.BaseProtocol):
         self.transport: asyncio.BaseTransport = None
         self.loop: asyncio.AbstractEventLoop = None
         self.recv_buffer: bytes = b""
-        self.call_create: Callable[[], Coroutine[Any, Any, Any]] = lambda: None
+        self.call_create: Callable[[], Coroutine[Any, Any, Any]] = lambda: None  # type: ignore[assignment]
         if self.is_server:
             self.active_connections: dict[str, ModbusProtocol] = {}
         else:
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -159,11 +159,7 @@ class ModbusProtocol(asyncio.BaseProtocol):
         self.transport: asyncio.BaseTransport = None
         self.loop: asyncio.AbstractEventLoop = None
         self.recv_buffer: bytes = b""
-
-        async def _noop():
-            ...
-
-        self.call_create: Callable[[], Coroutine[Any, Any, Any]] = _noop
+        self.call_create: Callable[[], Coroutine[Any, Any, Any]] = lambda: None
         if self.is_server:
             self.active_connections: dict[str, ModbusProtocol] = {}
         else:
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -524,7 +524,7 @@ class NullModem(asyncio.DatagramTransport, asyncio.Transport):
         asyncio.DatagramTransport.__init__(self)
         asyncio.Transport.__init__(self)
         self.protocol: ModbusProtocol = protocol
-        self.other_modem: NullModem = None
+        self.other_modem: NullModem
         self.listen = listen
         self.manipulator: Callable[[bytes], list[bytes]] | None = None
         self._is_closing = False
@@ -597,10 +597,10 @@ class NullModem(asyncio.DatagramTransport, asyncio.Transport):
         if self.connections:
             with suppress(KeyError):
                 del self.connections[self]
-        if self.other_modem:
-            self.other_modem.other_modem = None
+        if hasattr(self, 'other_modem'):
+            del self.other_modem.other_modem
             self.other_modem.close()
-            self.other_modem = None
+            del self.other_modem
         if self.protocol:
             self.protocol.connection_lost(None)
 
--- a/pymodbus/transport/transport_serial.py
+++ b/pymodbus/transport/transport_serial.py
@@ -38,7 +38,7 @@ class SerialTransport(asyncio.Transport):
 
     def close(self, exc: Exception | None = None) -> None:
         """Close the transport gracefully."""
-        if not self.sync_serial:
+        if not hasattr(self, 'sync_serial'):
             return
         with contextlib.suppress(Exception):
             self.sync_serial.flush()
@@ -51,7 +51,7 @@ class SerialTransport(asyncio.Transport):
         else:
             self.async_loop.remove_reader(self.sync_serial.fileno())
         self.sync_serial.close()
-        self.sync_serial = None
+        del self.sync_serial
         if exc:
             with contextlib.suppress(Exception):
                 self._protocol.connection_lost(exc)


pymodbus/transport/transport_serial.py
```diff
@@ -51,7 +51,7 @@ def close(self, exc: Exception | None = None) -> None:
         else:
             self.async_loop.remove_reader(self.sync_serial.fileno())
         self.sync_serial.close()
-        self.sync_serial = None
+        del self.sync_serial
```
janiversen, line 54: This works only in theory as far as I can see, for a couple of reasons:
- the variables set in __init__ in sync_serial is also removed
- it prohibits reusing the object, because sync_serial no longer exist
- It makes debugging quite difficult because you can no longer see the (faulty) sync_serial.

pymodbus/transport/transport_serial.py
```diff
@@ -38,7 +38,7 @@ def setup(self):
 
     def close(self, exc: Exception | None = None) -> None:
         """Close the transport gracefully."""
-        if not self.sync_serial:
+        if not hasattr(self, 'sync_serial'):
```
janiversen, line 41: see below please.

pymodbus/transport/transport.py
```diff
@@ -593,10 +597,10 @@ def close(self) -> None:
         if self.connections:
             with suppress(KeyError):
                 del self.connections[self]
-        if self.other_modem:
-            self.other_modem.other_modem = None
+        if hasattr(self, 'other_modem'):
+            del self.other_modem.other_modem
```
janiversen, line 601: This is an example of different coding styles...removing variables is a no-go to me, but in this case your code is just as good as mine...so I have no problem approving this part.

janiversen (changes requested):

pymodbus/transport/transport.py
```diff
@@ -159,7 +159,11 @@ def __init__(
         self.transport: asyncio.BaseTransport = None
         self.loop: asyncio.AbstractEventLoop = None
         self.recv_buffer: bytes = b""
-        self.call_create: Callable[[], Coroutine[Any, Any, Any]] = lambda: None
+
+        async def _noop():
```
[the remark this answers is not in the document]
alexrudd2: This is the only one!

(`dev` is down to 9 `mypy` errors with `strict_optional = true`)

pymodbus/transport/transport_serial.py
alexrudd2: > the variables set in init in sync_serial is also removed

What variables do you expect to persist after `close()` is called, and why would they need to persist?

> it prohibits reusing the object, because sync_serial no longer exist

The existing code doesn't allow re-use, since you can't any method after it's been set to `None`
`AttributeError: 'NoneType' object has no attribute 'foo'`
Any re-use will have to recreate it anyways.

> It makes debugging quite difficult because you can no longer see the (faulty) sync_serial.

I'd rather deal with an explicitly deleted object rather than one that may be garbage collected at any time!

pymodbus/transport/transport_serial.py
alexrudd2: Yes, I understand your concern.  Permitting `sync_serial` to be `None` means introducing more code to guard against calling it when `None`, so I found it cleanest to simply `del` the whole object.

pymodbus/transport/transport.py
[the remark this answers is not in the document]
janiversen: It's the principle, we d9 not want to have extra code just to please mypy....of course assuming the original is correct and in accordance with the PEP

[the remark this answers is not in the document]
alexrudd2:
```python
self.call_create: Callable[[], Coroutine[Any, Any, Any]] = lambda: None
```
is not in accordance with the [revised PEP484](url).

```python
self.call_create: Callable[[], Coroutine[Any, Any, Any]] | None = lambda: None
```
would be, but of course it introduces other type errors with calling code.

I tried to come up with an `async lambda: return None` or similar, but could not find one.

pymodbus/transport/transport_serial.py
janiversen: Or telling mypy to keep silent !!! we do not add code, just to please mypy.

pymodbus/transport/transport_serial.py
janiversen: after calling close() you can call connect().

I politely disagree, deleting objects is not really standard python.

alexrudd2: > after calling close() you can call connect().

```python
   self.async_loop.remove_reader(self.sync_serial.fileno())
        self.sync_serial.close()
        self.sync_serial = None
>       self.sync_serial.connect()  # not in dev, for demonstration only
E       AttributeError: 'NoneType' object has no attribute 'connect'

pymodbus/transport/transport_serial.py:55: AttributeError
======================================= short test summary info =======================================
FAILED test/sub_transport/test_basic.py::TestBasicSerial::test_external_methods - AttributeError: 'NoneType' object has no attribute 'connect'
FAILED test/sub_transport/test_comm.py::TestCommModbusProtocol::test_connected[CommType.SERIAL-socket://localhost:5020] - AttributeError: 'NoneType' object has no attribute 'connect'
FAILED test/sub_transport/test_comm.py::TestCommModbusProtocol::test_split_serial_packet[CommType.SERIAL-socket://localhost:5020] - assert False
FAILED test/sub_transport/test_comm.py::TestCommModbusProtocol::test_serial_poll[CommType.SERIAL-socket://localhost:5020] - AttributeError: 'NoneType' object has no attribute 'connect'
```

janiversen: ahh that's correct, that is still not cleaned up on dev.

But anyhow my comment still stands, deleting variables is not the standard python way (even though it also works).

As I just wrote in another place, we want to have object variables declared in __init__ thus the None, but in reality the way we use the code, it will always. be set....and honestly if somebody takes it and use the class elsewhere it is not our problem.

alexrudd2: > But anyhow my comment still stands, deleting variables is not the standard python way (even though it also works).

Agreed.  However, I've been unable to find a better way.  🤷🏼‍♂️ 

> As I just wrote in another place, we want to have object variables declared in init thus the None, but in reality the way we use the code, it will always. be set....and honestly if somebody takes it and use the class elsewhere it is not our problem.

There already _is_ a runtime check in `dev`!  This is internal to `pymodbus`, not some external caller
https://github.com/pymodbus-dev/pymodbus/blob/7a3ce95aa73b9e88e4ef5f6d62c83791f3342f1c/pymodbus/transport/transport_serial.py#L41-L42

Go ahead and remove it and watch the tests fail...
```
FAILED test/sub_transport/test_comm.py::TestCommModbusProtocol::test_connected[CommType.SERIAL-socket://localhost:5020] - AttributeError: 'NoneType' object has no attribute 'fileno'
FAILED test/sub_transport/test_comm.py::TestCommModbusProtocol::test_split_serial_packet[CommType.SERIAL-socket://localhost:5020] - assert False
FAILED test/sub_transport/test_comm.py::TestCommModbusProtocol::test_serial_poll[CommType.SERIAL-socket://localhost:5020] - assert False
```

janiversen: I wrote about the theme in general....you know very well transport_serial is imported code that I slowly change to make it fit our needs.

There is a better way, ask mypy to stop complaining or for that sake exclude that file from mypy.

alexrudd2: > I wrote about the theme in general....

Yes, I understand your point in general. 😃  It's why I'm refactoring the suggestions from @MAKOMO to, for instance, not use `cast()` or `assert()`.  (He's lurking in this thread to see if I have success, I think)

> you know very well transport_serial is imported code that I slowly change to make it fit our needs.

Haha, actually no I don't know that.  😄 You have the entire context for the code in your head, but I only understand bits and pieces.  It's why my contributions are stricter checks that do not rely on understanding the whole codebase.

Another way of thinking about the question...  If code can be explained to `mypy` then it is easier to explain to junior contributors.

pymodbus/transport/transport_serial.py
janiversen: My experience is actually different a lot of the juniors do not understand nor agree with type checking at all....

There are many python developers who see the lack of type checking as a big advantage...being a C/C++/C# I am very used to strict type checking, but have also seen how it can make code a lot more complicated.

Personally I think the bridge (free use / type checking) python tries to build might end being a split into 2 products.

I like the freedom python allows, but see that mypy catches many problems....so for me a balance is the correct choice, meaning think if what mypy suggest mamés the code better if not ask mypy to stop complaining.

pymodbus/transport/transport_serial.py
MAKOMO: Yes, I am "lurking" and enjoy the somewhat funny discussion on the use of explicit type specification and automatic validators which seem not to be too important for "internal code" as somehow the developers have some internal (documented?) knowledge that they know (also potential contributors?) and do not violate (as they do not introduce bugs). My experience is different especially with those implicit optionals. They helped me quite a bit to find "null pointer" bugs in my large code base automatically that I would otherwise never have found and it makes me feel comfortable that those checks run on any change I further apply to it. Anyhow, I am following your discussion to learn. Keep going!

alexrudd2:
Closing in favor of https://github.com/pymodbus-dev/pymodbus/pull/1888

PR #1874, opened 2023-10-31
Eliminiate implicit optional in reconnect_delay*

alexrudd2:
See https://github.com/pymodbus-dev/pymodbus/pull/1856#discussion_r1373590566 and https://github.com/pymodbus-dev/pymodbus/pull/1856#discussion_r1373590329

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -88,7 +88,7 @@ class CommParams:
     comm_name: str | None = None
     comm_type: CommType | None = None
     reconnect_delay: float | None = None
-    reconnect_delay_max: float | None = None
+    reconnect_delay_max: float = 0.0
     timeout_connect: float | None = None
     host: str = "127.0.0.1"
     port: int = 0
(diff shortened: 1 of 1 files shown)


pymodbus/transport/transport.py
```diff
@@ -88,7 +88,7 @@ class CommParams:
     comm_name: str | None = None
     comm_type: CommType | None = None
     reconnect_delay: float | None = None
-    reconnect_delay_max: float | None = None
+    reconnect_delay_max: float = 0.0
```
janiversen, line 91: Actually there is a hidden problem here, "reconnect_delay_max" is defined a milliseconds and is an integer.

It seems asyncio.py does still use seconds, so I have to look at code before I approve this one.

janiversen (changes requested):

janiversen:
your PR´s are starting to get very small.

janiversen (approved):
LGTM (reconnect_delay is consistent float).

Issue #1883, opened 2023-11-09
aiohttp==3.9.0b1  breaks pymodbus tests


alexrudd2:
There is some sort of change in `aiohttp==3.9.0b1` that breaks 2 of the tests.

```python
python3 -m pytest -k simulator_server_tcp .
[...] snip
E       AssertionError: ERROR tasks/threads hanging after 9 retries:
E
E       assert False
```

`aiohttp==3.9.0b0` is fine

alexrudd2:
Breaking change bisected to https://github.com/aio-libs/aiohttp/pull/7780

Dreamsorcerer:
Can you link to the actual tests failing. The above comment doesn't show the test, filename, stacktrace or anything. ;)

alexrudd2:
https://github.com/pymodbus-dev/pymodbus/actions/runs/6805365296/job/18505192239

```
 =========================== short test summary info ============================
ERROR test/sub_examples/test_examples.py::TestExamples::test_simulator - AssertionError: ERROR tasks/threads hanging after 9 retries:
  
assert False
ERROR test/sub_server/test_simulator.py::TestSimulator::test_simulator_server_tcp - AssertionError: ERROR tasks/threads hanging after 9 retries:
  
assert False
```

Here's one of the two tests:
https://github.com/pymodbus-dev/pymodbus/blob/1db29affa46279f7eed8e572a528842a11c0e748/test/sub_server/test_simulator.py#L576-L581

I think the actual `assert` is here:
https://github.com/pymodbus-dev/pymodbus/blob/dev/test/conftest.py#L76
It's a fixture to make sure teardown/cleanup works, but you'll have to ask @janiversen  for how it actually works.

Dreamsorcerer:
Ah, you're starting/stopping the site directly and not the runner. Which atleast means that the app will never run any cleanup steps (and with my recent change, won't run any shutdown steps either). Let me take a closer look and see if any of this makes sense.

alexrudd2:
Thanks for taking a look.  If you need anything from me, please let me know.  (The CI runs in a fork too, if you'd prefer).  That said, I don't know these tests and cannot comment on whether they're actually the right way to do things.   You'll have to take it up with Jan.

Dreamsorcerer:
So, I think if you just replace `site.stop()` with `runner.shutdown()`, as shown in the documentation: https://docs.aiohttp.org/en/stable/web_advanced.html#application-runners
[...]

PR #1886, opened 2023-11-09
Support aiohttp 3.9.0b1

alexrudd2:
Thanks to @Dreamsorcerer in https://github.com/pymodbus-dev/pymodbus/issues/1883#issuecomment-1804196681

### DIFF
--- a/pymodbus/server/simulator/http_server.py
+++ b/pymodbus/server/simulator/http_server.py
@@ -161,6 +161,7 @@ class ModbusSimulatorServer:
         self.serving: asyncio.Future = asyncio.Future()
         self.log_file = log_file
         self.site = None
+        self.runner: web.AppRunner
         self.http_host = http_host
         self.http_port = http_port
         self.web_path = os.path.join(os.path.dirname(__file__), "web")
@@ -236,9 +237,9 @@ class ModbusSimulatorServer:
     async def run_forever(self, only_start=False):
         """Start modbus and http servers."""
         try:
-            runner = web.AppRunner(self.web_app)
-            await runner.setup()
-            self.site = web.TCPSite(runner, self.http_host, self.http_port)
+            self.runner = web.AppRunner(self.web_app)
+            await self.runner.setup()
+            self.site = web.TCPSite(self.runner, self.http_host, self.http_port)
             await self.site.start()
         except Exception as exc:
             Log.error("Error starting http server, reason: {}", exc)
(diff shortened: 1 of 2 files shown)


janiversen:
I think we should wait !!

HA uses:
aiohttp==3.9.0b0;python_version>='3.12'
aiohttp==3.8.5;python_version<'3.12'

and that works fine, you pinned 3.9.0b0 so lets stay with that until they get a stable version.

Dreamsorcerer:
HA is affected by a different regression (client-side), which we already have a fix for. The change in this PR is correct, for both current and previous versions of aiohttp (as shown in the docs).

pymodbus/server/simulator/http_server.py
```diff
@@ -161,6 +161,7 @@ def __init__(
         self.serving: asyncio.Future = asyncio.Future()
         self.log_file = log_file
         self.site = None
+        self.runner: web.AppRunner
```
janiversen, line 164: This change should cause a mypy problem, because self.runner is now defined outside __init__

Dreamsorcerer:
Even in previous versions of aiohttp, you were never running the cleanup steps for the web app.

pyproject.toml
```diff
@@ -56,7 +56,7 @@ repl = [
     "click>=8.0.0"
 ]
 simulator = [
-    "aiohttp==3.9.0b0"
+    "aiohttp>=3.9.0b0"
```
janiversen, line 59: Let please not start hunting beta versions !!!!

I added the b0, but forgot to limit it to 3.12...see my preference as comment in some of the other PR´s

janiversen (changes requested):

pyproject.toml
alexrudd2: I'm fine with switching to 
```aiohttp==3.9.0b0;python_version>='3.12'
aiohttp==3.8.5;python_version<'3.12'
```

However, it seems orthogonal to this PR, which DreamSorcerer has stated is an improvement regardless of `aiohttp` version.
> The change in this PR is correct, for both current and previous versions of aiohttp (as shown in the docs).
> Even in previous versions of aiohttp, you were never running the cleanup steps for the web app.

pymodbus/server/simulator/http_server.py
alexrudd2: `pylint` did complain... until I added this line with the type, which satisfies both `pylint` and `mypy`

pyproject.toml
janiversen: I have no problem approving the code changes!

alexrudd2: This PR was not a version bump; in fact the opposite (expanding a pin).

Nevertheless, I have expanded the version strategy so `3.8.6` is permitted on Python <3.12.

janiversen (approved):
LGTM, thanks

PR #1875, opened 2023-10-31
Eliminate final (!) implicit optional

alexrudd2:
See discussions in https://github.com/pymodbus-dev/pymodbus/pull/1856, which this supercedes

Note: CI will fail unless/until all the other PRs are merged.

EDIT: I rebased and fixed a pylint error.

PR #1888, opened 2023-11-12
Ignore all remaining implicit optional

alexrudd2:
See #1875, #1873, #1856

Since there's not an agreed way to eliminate the final implicit optionals, ignore them instead.

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -517,7 +517,7 @@ class NullModem(asyncio.DatagramTransport, asyncio.Transport):
         asyncio.DatagramTransport.__init__(self)
         asyncio.Transport.__init__(self)
         self.protocol: ModbusProtocol = protocol
-        self.other_modem: NullModem = None
+        self.other_modem: NullModem
         self.listen = listen
         self.manipulator: Callable[[bytes], list[bytes]] | None = None
         self._is_closing = False
@@ -590,10 +590,10 @@ class NullModem(asyncio.DatagramTransport, asyncio.Transport):
         if self.connections:
             with suppress(KeyError):
                 del self.connections[self]
-        if self.other_modem:
-            self.other_modem.other_modem = None
+        if hasattr(self, "other_modem"):
+            del self.other_modem.other_modem
             self.other_modem.close()
-            self.other_modem = None
+            del self.other_modem
         if self.protocol:
             self.protocol.connection_lost(None)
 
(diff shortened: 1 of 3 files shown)


pymodbus/transport/transport.py
```diff
@@ -517,7 +517,7 @@ def __init__(self, protocol: ModbusProtocol, listen: int | None = None) -> None:
         asyncio.DatagramTransport.__init__(self)
         asyncio.Transport.__init__(self)
         self.protocol: ModbusProtocol = protocol
-        self.other_modem: NullModem = None
+        self.other_modem: NullModem
```
janiversen, line 520: This looks wrong, declaring a variable with an assignment, what value does it have and is that a defined behavior? very strange that mypy accept an unknown value.

That could potentially be used in quite a number of cases..

pymodbus/transport/transport.py
```diff
@@ -590,10 +590,10 @@ def close(self) -> None:
         if self.connections:
             with suppress(KeyError):
                 del self.connections[self]
-        if self.other_modem:
-            self.other_modem.other_modem = None
+        if hasattr(self, "other_modem"):
+            del self.other_modem.other_modem
```
janiversen, line 594: I thought we agreed not to delete variables.

janiversen (changes requested):

pymodbus/transport/transport.py
alexrudd2: > This is an example of different coding styles...removing variables is a no-go to me, but in this case your code is just as good as mine...so I have no problem approving this part.

_Originally posted by @janiversen in https://github.com/pymodbus-dev/pymodbus/pull/1873#discussion_r1377223124_

You've said different things at different times (re: deleting), so I'm not sure now which you prefer.  Anyway, I can revert.

pymodbus/transport/transport.py
alexrudd2: It's just [adding to](https://stackoverflow.com/questions/74308012/type-hints-without-value-assignment-in-python) `_annotations_` to help the type checker; actual definition is done later.

There's no use of `other_modem` being `Null` anyways; it always get instanced before use.

pymodbus/transport/transport.py
janiversen: My preference is clearly not to delete, as this is not standard python, however if it's unavoidable I can live with it,

In this case it does not seem unavoidable.

pymodbus/transport/transport.py
janiversen: Pylint does not complain about it being declared outside init, so for all purposes it is declared and thus must have a value.

I am just curious what value that would, this is not blocking an approval.

janiversen: Having read the link to stack overflow I am really confused why pylint accept this.....but for some reason it does.

alexrudd2: My understanding is the definition/annotation here doesn't affect the runtime code, but `pylint` is satisfied because it shows the programmer all of the instance variables in `_init_`.

janiversen (approved):
Thanks.

PR #1889, opened 2023-11-15
Update transport.py

garvan2021:
transport.abort() is already contains transport.close(), and it would cause error when transport.close() occurs after transport.abort()

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -414,7 +414,8 @@ class ModbusProtocol(asyncio.BaseProtocol):
         if self.transport:
             if hasattr(self.transport, "abort"):
                 self.transport.abort()
-            self.transport.close()
+            else:
+                self.transport.close()
             self.transport = None
         self.recv_buffer = b""
         if self.is_server:


garvan2021:
Error:
> AttributeError: 'NoneType' object has no attribute 'close'

Machine: ARM 64

How it occurs:
> AsyncModbusSerialClient().async_execute(...)

janiversen (changes requested):
Looks correct, but it broke all tests so something is not right.

janiversen:
I think your problem is different, because the error states the self.transport is None !!

alexrudd2:
> I think your problem is different, because the error states the self.transport is None !!


If only there were some [type checking tool](https://github.com/pymodbus-dev/pymodbus/pull/1888/files#diff-232ea64eea656684ba1d20bc100847b5fabad3d4627d4b8a51c49041ad84897dR418) that could find such problems with implicit optionals and `None`!
:laughing::laughing:

alexrudd2:
> Error:
> 
> > AttributeError: 'NoneType' object has no attribute 'close'
> 
> Machine: ARM 64
> 
> How it occurs:
> 
> > AsyncModbusSerialClient().async_execute(...)

Does this still occur for you if you use the most recent `dev`?

garvan2021:
I should be more carefully with my code, may there is another error I missed. So I close the commit temperary and do apologize for my mistake

alexrudd2:
> I should be more carefully with my code, may there is another error I missed. So I close the commit temperary and do apologize for my mistake

It was OK.  The test failure only said the code had changed, not that the code was wrong.

I tried to solve your problem a different way. https://github.com/pymodbus-dev/pymodbus/pull/1890.  

Can you try installing from `dev` and seeing if this works for you?

PR #1890, opened 2023-11-15
Remove unnecessary abort() call

alexrudd2:
Supercedes #1889

See rationale in https://github.com/pymodbus-dev/pymodbus/pull/1889#issuecomment-1813056191

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -412,8 +412,6 @@ class ModbusProtocol(asyncio.BaseProtocol):
         if not intern:
             self.is_closing = True
         if self.transport:
-            if hasattr(self.transport, "abort"):
-                self.transport.abort()
             self.transport.close()
             self.transport = None
         self.recv_buffer = b""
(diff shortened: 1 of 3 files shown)


alexrudd2:
Ooops, this commit is the wrong one.

pymodbus/transport/transport.py
```diff
@@ -614,7 +612,7 @@ def write(self, data: bytes) -> None:
     # Dummy methods #
     # ------------- #
     def abort(self) -> None:
-        """Abort connection."""
+        """Old alias for closing the connection."""
```
janiversen, line 615: That is not a "old" ... it is needed because the parent class declares it virtual.

pymodbus/transport/transport_serial.py
```diff
@@ -117,7 +117,7 @@ def is_closing(self):
         return False
 
     def abort(self) -> None:
-        """Close the transport immediately."""
+        """Old alias for closing the connection."""
```
janiversen, line 120: Same here.

janiversen (changes requested):

janiversen (approved):
Thanks

PR #1908, opened 2023-12-08
Missing Copyright in License file

andresrinivasan:


janiversen:
Help is always welcome, which is why I started by thanking for the PR.

I did not say that we would not merge your PR, only that I think it makes no legal difference.

However just marking 2023 indicates there was no copyright in 2022 and earlier years, which plain simple is wrong, which is why I asked for a change in the PR.

andresrinivasan:
Ah, I didn't understand. You're asking that I update the PR to 2008-2023. Correct?

janiversen:
yes

PR #1913, opened 2023-12-12
ModbusTCP: identify slave context by peer ip and unit id

ickedude:
The slave context is now identified by a tuple of the peer      IP address and unit identifier of the request.
It allows to have multiple master sending an unit identifier of 0xFF and also to implement a bridge with routing based on both information.

minimal working example:
```python
#!/usr/bin/env python3

import asyncio

from pymodbus.datastore.store import ModbusSequentialDataBlock
from pymodbus.datastore import ModbusSlaveContext
from pymodbus.datastore import ModbusServerContext
from pymodbus.server import StartAsyncTcpServer
from pymodbus.framer.socket_framer import ModbusSocketFramer


async def run_modbus_server():
    ds1 = ModbusSequentialDataBlock(0, [0]*100)
    ds2 = ModbusSequentialDataBlock(0, [0]*64)
    ds3 = ModbusSequentialDataBlock(0, [0]*64)
    ds4 = ModbusSequentialDataBlock(0, [0]*144)
    slaves = {
        ('192.168.10.22', 0xFF): ModbusSlaveContext(di=ds1, co=ds1, hr=ds1, ir=ds1, zero_mode=True),
        ('192.168.10.20', 0x01): ModbusSlaveContext(di=ds2, co=ds2, hr=ds2, ir=ds2),
        ('192.168.10.20', 0x02): ModbusSlaveContext(di=ds3, co=ds3, hr=ds3, ir=ds3),
        ('192.168.10.21', 0xFF): ModbusSlaveContext(di=ds4, co=ds4, hr=ds4, ir=ds4),
    }
    server_context = ModbusServerContext(slaves=slaves, single=False)
    server = await StartAsyncTcpServer(
        context=server_context,
        address=('192.168.10.10', 5020),
        framer=ModbusSocketFramer,
    )
    return server


async def main() -> None:
    server = asyncio.create_task(run_modbus_server())
    await server


if __name__ == "__main__":
    asyncio.run(main(), debug=True)
```


ickedude:
I guess I cannot resolve the failing check, because CI / windows-latest - 3.12 reports
No Python at '"C:\hostedtoolcache\windows\Python\3.12.0\x64\python.exe'

pymodbus/client/base.py
```diff
@@ -387,7 +387,7 @@ def __init__(  # pylint: disable=too-many-arguments
         # Common variables.
         self.framer = FRAMER_NAME_TO_CLASS.get(
             framer, cast(Type[ModbusFramer], framer)
-        )(ClientDecoder(), self)
+        )(ClientDecoder(), client=self)
```
janiversen, line 390: client is NOT optional.

pymodbus/framer/ascii_framer.py
```diff
@@ -32,12 +32,12 @@ class ModbusAsciiFramer(ModbusFramer):
 
     method = "ascii"
 
-    def __init__(self, decoder, client=None):
+    def __init__(self, decoder, **kwargs):
```
janiversen, line 35: kwargs indicates it takes any number of parameters, but it only takes client=, the change make it harder to understand the code.

pymodbus/framer/base.py
```diff
@@ -27,14 +27,14 @@ class ModbusFramer:
     def __init__(
         self,
         decoder: ClientDecoder | ServerDecoder,
-        client=None,
+        **kwargs
```
janiversen, line 30: client= is a named parameter, why make the anonymous....also the class do not allow other parameters.

pymodbus/framer/base.py
```diff
@@ -27,14 +27,14 @@ class ModbusFramer:
...
     ) -> None:
         """Initialize a new instance of the framer.
 
         :param decoder: The decoder implementation to use
         """
         self.decoder = decoder
-        self.client = client
+        self.client = kwargs.get('client')
```
janiversen, line 37: using kwargs complicates the code.

pymodbus/framer/binary_framer.py
```diff
@@ -41,12 +41,12 @@ class ModbusBinaryFramer(ModbusFramer):
 
     method = "binary"
 
-    def __init__(self, decoder, client=None):
+    def __init__(self, decoder, **kwargs):
```
janiversen, line 44: it does not take any parameter, only client=

pymodbus/framer/rtu_framer.py
```diff
@@ -53,12 +53,12 @@ class ModbusRtuFramer(ModbusFramer):
 
     method = "rtu"
 
-    def __init__(self, decoder, client=None):
+    def __init__(self, decoder, **kwargs):
```
janiversen, line 56: It does not take any parameter, just client=

it seems you are preparing the code to somehow be overwritten from the outside, but this code is internal.

pymodbus/framer/socket_framer.py
```diff
@@ -36,13 +37,32 @@ class ModbusSocketFramer(ModbusFramer):
 
     method = "socket"
 
-    def __init__(self, decoder, client=None):
+    def __init__(self, decoder, **kwargs):
```
janiversen, line 40: same here.

pymodbus/framer/socket_framer.py
```diff
@@ -36,13 +37,32 @@ class ModbusSocketFramer(ModbusFramer):
...
         """Initialize a new instance of the framer.
 
         :param decoder: The decoder factory implementation to use
         """
-        super().__init__(decoder, client)
+        super().__init__(decoder, **kwargs)
         self._hsize = 0x07
+        self.transport = kwargs.get('transport')
+
+    def _validate_slave_id(self, slaves: list, single: bool) -> bool:
```
janiversen, line 49: This looks like a function that should be in base.

tcp connections allow others framers as well !

pymodbus/framer/socket_framer.py
```diff
@@ -36,13 +37,32 @@ class ModbusSocketFramer(ModbusFramer):
...
-    def __init__(self, decoder, client=None):
+    def __init__(self, decoder, **kwargs):
         """Initialize a new instance of the framer.
 
         :param decoder: The decoder factory implementation to use
         """
-        super().__init__(decoder, client)
+        super().__init__(decoder, **kwargs)
         self._hsize = 0x07
+        self.transport = kwargs.get('transport')
```
janiversen, line 47: If transport is need it should be an optional parameter, and not only in the socket framer, since most framers can be used with tcp.

pymodbus/framer/tls_framer.py
```diff
@@ -28,12 +28,12 @@ class ModbusTlsFramer(ModbusFramer):
 
     method = "tls"
 
-    def __init__(self, decoder, client=None):
+    def __init__(self, decoder, **kwargs):
```
janiversen, line 31: it does not take any parameter only client=

pymodbus/framer/socket_framer.py
```diff
@@ -36,13 +37,32 @@ class ModbusSocketFramer(ModbusFramer):
...
+
+    def _validate_slave_id(self, slaves: list, single: bool) -> bool:
+        """Validate if the received data is valid for the client.
```
janiversen, line 50: slaves is a list of int, not just any list.

pymodbus/framer/socket_framer.py
```diff
@@ -36,13 +37,32 @@ class ModbusSocketFramer(ModbusFramer):
...
+
```
janiversen, line 51: having _ means it´s an internal function, but I cannot see that it is called anywhere.

pymodbus/server/async_io.py
```diff
@@ -180,19 +180,23 @@ def execute(self, request, *addr):
         if self.server.request_tracer:
             self.server.request_tracer(request, *addr)
 
-        broadcast = False
+        if self.server.broadcast_enable and not request.slave_id:
```
janiversen, line 183: Moving this out from the try/except, will probably cause problems with the remote datastore.

pymodbus/server/async_io.py
```diff
@@ -203,14 +207,12 @@ def execute(self, request, *addr):
...
             response = request.doException(merror.SlaveFailure)
-        # no response when broadcasting
-        if not broadcast:
-            response.transaction_id = request.transaction_id
-            response.slave_id = request.slave_id
-            skip_encoding = False
-            if self.server.response_manipulator:
-                response, skip_encoding = self.server.response_manipulator(response)
-            self.send(response, *addr, skip_encoding=skip_encoding)
+        response.transaction_id = request.transaction_id
```
janiversen, line 210: This changes functionality, I am pretty sure that when broadcasting the devices do not respond (Have to check the spec on that).

janiversen (changes requested):
Part of this PR is ok (the peer name etc), but the parameter changes are not making the code better and not needed.

janiversen:
In general you need to document this feature, so that users know how to use it.

An example (or extending an existing example) would also be nice, since most users look more at the examples than at the documentation.


Have you considered, what happens to a server running tcp, with framer=socket and the slaves defined as slave id´s not uid...we need to secure that those still work unchanged.

pymodbus/framer/ascii_framer.py
ickedude: There was a misunderstanding on my site. After changing the code to allow missing peer information I forgot to revert the changes to the interface. Will correct that.

pymodbus/framer/base.py
ickedude: same here

pymodbus/framer/base.py
ickedude: see comment above

pymodbus/framer/binary_framer.py
ickedude: see comment above

pymodbus/framer/rtu_framer.py
ickedude: see comment above

pymodbus/framer/socket_framer.py
ickedude: see comment above

pymodbus/server/async_io.py
```diff
@@ -180,19 +180,23 @@ def execute(self, request, *addr):
...
-            if self.server.broadcast_enable and not request.slave_id:
-                broadcast = True
-                # if broadcasting then execute on all slave contexts,
-                # note response will be ignored
-                for slave_id in self.server.context.slaves():
-                    response = request.execute(self.server.context[slave_id])
-            else:
-                context = self.server.context[request.slave_id]
-                response = request.execute(context)
+            if self.server.is_server and self.comm_params.comm_type in (CommType.TCP,):
```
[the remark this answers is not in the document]
ickedude: Had the exact same question regarding is_server. It is [set to true by ModbusBaseServer](https://github.com/pymodbus-dev/pymodbus/blob/131d30391efd9c64f1fec857ae0372746374e6d2/pymodbus/server/async_io.py#L263C1-L266C10).
I only had time to elaborate TCP. Since UDP is also using a socket, it should be supported.
TLS I am not quite sure, because [the doc](https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport.get_extra_info) is not specific enough if the information is provided. I need to look into the tests to be sure.

pymodbus/server/async_io.py
ickedude: case broadcast before:
1. a response is created for the last slave_context
2. a NoSuchSlaveException could not happen, because of the iterator
3. a ServerContext, SlaveContext, datastore could rise an exception within the loop
4. data integrity is corrupted, because only slave contexts till the exception were updated
5. a misleading log messages and a response is created
6. the created responses are never send

case broadcast pull request:
1. responses are ignored
2. raised exceptions in ServerContext, SlaveContext, datastore within the loop bubble up
3. server will probably terminate or restart

I think exceptions should be handled in the server context.
Note the misleading error messages is still there for non broadcast use cases

pymodbus/server/async_io.py
ickedude: Note the [return](https://github.com/pymodbus-dev/pymodbus/blob/41b2d3a601c1a2da081649b16a3510c6dd57ae53/pymodbus/server/async_io.py#L189) after the loop.

pymodbus/framer/socket_framer.py
ickedude: It is overwriting the method from [ModbusFramer](https://github.com/pymodbus-dev/pymodbus/blob/131d30391efd9c64f1fec857ae0372746374e6d2/pymodbus/framer/base.py#L48C1-L61C45).

pymodbus/framer/socket_framer.py
ickedude: Just copied the signature from [ModbusFramer](https://github.com/pymodbus-dev/pymodbus/blob/131d30391efd9c64f1fec857ae0372746374e6d2/pymodbus/framer/base.py#L48), didn't want to change the signture, but yes it can be more explicit, because with this pull request it is not `int` anymore.
I would like to change to:
```python
def _validate_slave_id(self, slaves: Sequence[Union[int, Tuple[str, int]]], single: bool) -> bool:
```
, but because ModbusFramer is requesting a list, we need to change ModbusFramer or use list instead of sequence.
`int` is still possible, because of  `0` used for the single option (see [ModbusServerContext](https://github.com/pymodbus-dev/pymodbus/blob/131d30391efd9c64f1fec857ae0372746374e6d2/pymodbus/datastore/context.py#L133)) and if the transport is not capable of delivering the peer information, the uid is still used.

ickedude:
> In general you need to document this feature, so that users know how to use it.

You are totally right. I created the pull request to start a conversation.

> Have you considered, what happens to a server running tcp, with framer=socket and the slaves defined as slave id´s not uid...we need to secure that those still work unchanged.

I don't get this part. Do you mean the RemoteSlaveContext and related forwarder examples? If you mean that, then you are right. Did not know about RemoteSlaveContext till today.

pymodbus/client/base.py
```diff
@@ -91,7 +91,7 @@ def __init__(
...
```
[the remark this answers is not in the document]
ickedude: See below comment

pymodbus/client/base.py
ickedude: See below comment

pymodbus/framer/rtu_framer.py
ickedude: Now, I remember my problem: [ModbusServerRequestHandler is instantiating the framer](https://github.com/pymodbus-dev/pymodbus/blob/131d30391efd9c64f1fec857ae0372746374e6d2/pymodbus/server/async_io.py#L64C1-L67C14), which means that I have to add transport to each framer constructor even if the framer does not care about transports. How should I handle that?

janiversen: Problem is that framer is supposed to be independent of the physical transport, but you want to interconnect framer and transport.

One way could be to let the transport layer set a uid parameter, that way we would avoid tcp specific code in the framers.

pymodbus/framer/socket_framer.py
janiversen: This is the beauty of mypy, you need to think carefully about how the parameters are used....but the advantage is that it get clearly documented.

pymodbus/framer/socket_framer.py
janiversen: But that is wrong, what happens when I use the rtu framer on tcp.

Or the tls framer for that matter.

pymodbus/server/async_io.py
janiversen: ? you removed the if which secured broadcast is not handled.

janiversen:
Having read your updated description, I think there is a fundamental flaw/misunterstanding this this Pull Request!

slaves are NOT clients!
slave id == unit id (its just a confusion of names unit=socket, slave=serial, but they mean the same).

slave id / unit id == 0xff is just another slave nothing special.
slave id / unit id == 0x00 means broadcast.

one server emulates n slaves/units, and when multiple client read the same register in the same slave they should pr definition get the same content.

Example:

Server contains slaves/units 1,2 and 3, slave 3 register 10 have a value of 17.


ClientA connect to server, and can address slave 1,2 and when reading slave 3 register 10 17 is returned

ClientB connect to server, and can address slave 1,2 and when reading slave 3 register 10 17 is returned


Your configuration would allow clientA and clientB to read different values for the same register and that would be wrong.

janiversen:
Sorry for not catching this when you started, but I thought you were talking about ModbusDeviceIdentification.

Alex0176:
Thats exacly what i searching for - is it possible to modify the datastore for eatch instance/client ip with diffrent content in runtime?

Alex0176:
> Having read your updated description, I think there is a fundamental flaw/misunterstanding this this Pull Request!
> 
> slaves are NOT clients! slave id == unit id (its just a confusion of names unit=socket, slave=serial, but they mean the same).
> 
> slave id / unit id == 0xff is just another slave nothing special. slave id / unit id == 0x00 means broadcast.
> 
> one server emulates n slaves/units, and when multiple client read the same register in the same slave they should pr definition get the same content.
> 
> Example:
> 
> Server contains slaves/units 1,2 and 3, slave 3 register 10 have a value of 17.
> 
> ClientA connect to server, and can address slave 1,2 and when reading slave 3 register 10 17 is returned
> 
> ClientB connect to server, and can address slave 1,2 and when reading slave 3 register 10 17 is returned
> 
> Your configuration would allow clientA and clientB to read different values for the same register and that would be wrong.

In my case i use pymodbus to simulate a smartmeter for more solar inverters and this inverters need diffrent serial numbers on the modbus. So i have to modify the serial number for every ,,fake" smartmeter. Otherwise i get troubles with the visualisation from the vendor. 
i step up for every inverter and if the identifierer is 0,1,2,3 i add to the original one, this value.

janiversen:
That is easy to do, give each device its own slave id and define a storage for each (NOT single) slave, then you can modify it as you like, that way reading register X can have different values for each slave.

ModbusDeviceIdentification is a singleton, and cannot be different for each slave...but it seems to me you are only looking to have different registers.

janiversen:
And just to be clear your PR does not change ModbusDeviceInformation at all.

janiversen:
Closing, as it seems to be stale.

If/when you start working on it again, feel free to reopen the Pr.

PR #1914, opened 2023-12-15
Bugfix Fix: default source_address invalid for IPv6

### LINKED ISSUE #1881: Default source_address breaks IPv6 and dual stack client connections - breaks localhost
### Versions

- Python: 3.10
- OS: MacOS / Linux
- Pymodbus: (3.4.0?) - 3.5.2
- Modbus Hardware (if used): N/A

### Pymodbus Specific

- Client: AsyncModbusTcpClient

### Description

PyModbus uses an explicit `source_address` default `("0.0.0.0", 0).  When connecting to dual-stack and IPv6 endpoints, including `localhost` this will result in failed connections with an error.

```
Failed to connect no matching local address with family=<AddressFamily.AF_INET6: 10> found
```

The source address `None` can and should be used for `source_address` for dual stack endpoints and will work with both IPv4 and IPv6.

### Code and Logs

```python
import pymodbus
import pymodbus.client

# Connects to IPv6 localhost
client = pymodbus.client.AsyncModbusTcpClient("::1", 9876)
    await client.connect()

# On some systems the same can and will happen for:
client = pymodbus.client.AsyncModbusTcpClient("localhost", 9876)
    await client.connect()
```

couling:
The source_address `0.0.0.0` is only valid for IPv4. For IPv6 this would need to be `::` but neither of these is valid for both IPv4 and IPv6 at the same time.  In python, `None` indicates any source address on either ip stack.

This PR fixes #1881

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -66,7 +66,7 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
             CommParams(
                 comm_type=kwargs.get("CommType"),
                 comm_name="comm",
-                source_address=kwargs.get("source_address", ("0.0.0.0", 0)),
+                source_address=kwargs.get("source_address", None),
                 reconnect_delay=reconnect_delay,
                 reconnect_delay_max=reconnect_delay_max,
                 timeout_connect=timeout,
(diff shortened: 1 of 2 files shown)


janiversen:
Please do not try to remake all of pymodbus in one pull request, that makes it hard to bisect if we later have a problem.

Simulator and http would be better in seperate pull requests

couling:
Rest assured I'm not, just trying to remove the same problem from multiple places, I'll trim this down a bit because I don't want to include anything for the HTTP server. That was over zealous of me.

Theres a couple of knock on issues I'm having with the modbus server because it doesn't interpret `None` the same way the client does for `source_address` and the same config class is used for both with the `source_address` reinterpreted as listening address.

I think I know the right way to get the modbus server to behave when it sees `None` but it needs a lot more testing.

I'll set the PR to "ready for review" when done.

couling:
It looks like the 3.12 tests are broken in the CI.

janiversen:
Seems the python action have a problem, investigating.

janiversen (approved):
LGTM, thanks.

PR #1918, opened 2023-12-26
Reset frame for serial connections

janiversen:
This might have a negative effect, when using the server on a rs485 line with multiple devices.


janiversen:
As I thought, this is not as easy as suggested.

PR #1936, opened 2024-01-14
contrib example: TCP drainage simulator with two devices

serpent213:
Simulates two Modbus TCP slave servers:

Digital IO (DIO) with 8 discrete inputs and 8 coils. The first two coils each control a simulated pump. Inputs are not used.

Water level meter (WLM) returning the current water level in the input register. It increases chronologically and decreases rapidly when one or two pumps are active.

Feels pretty straight forward, but I was not able to find a good example like this.


janiversen (approved):
LGTM, thanks.

PR #1964, opened 2024-02-03
Always set exclusive serial port access.

jameshilliard:
Should prevent two applications from trying to access the same serial port simultaneously.

Alternative to #1961.


janiversen (approved):
LGTM, thanks.

PR #1972, opened 2024-02-06
Support close_comm_on_error parameter in more places.

### LINKED ISSUE #1594: Connection closed after AsyncModbusSerialClient timeout
### Versions

- Python: 3.10
- OS: Ubuntu 22.04
- Pymodbus: 3.3.1
- Modbus Hardware (if used): GPIO Modbus

### Pymodbus Specific

- Client: rtu - async

### Description

When sending a read input registers and a timeout is triggered (device is not connected) the .connected property is False, even with close_comm_on_error=False

### Code and Logs
ModbusBaseClient._params(host=None, port='/dev/ttyUSB0', framer=<class 'pymodbus.framer.rtu_framer.ModbusRtuFramer'>, timeout=1.5, retries=3, retry_on_empty=False, close_comm_on_error=False, strict=True, broadcast_enable=False, kwargs={}, reconnect_delay=0, baudrate=9600, bytesize=8, parity='N', stopbits=1, handle_local_echo=False, source_address=None, sslctx=None, certfile=None, keyfile=None, password=None, server_hostname=None)

2023-06-12 02:25:50.345 DEBUG: Resetting frame - Current Frame in buffer -
2023-06-12 02:25:50.345 DEBUG: send: 0x1 0x4 0x0 0x1 0x0 0x2 0x20 0xb
2023-06-12 02:25:50.345 DEBUG: Adding transaction 1
2023-06-12 02:25:51.847 ERROR: TEMP] Connected? False

jameshilliard:
It looks like this is missing from a few places.

Should fix #1594, @bsense-rius you might want to give this a try.


janiversen (changes requested):
A couple of notes:

1) any api change must be documented in API_CHANGES
2) the documentation is not updated

I do not understand the purpose of your PR, we removed that parameter quite a while ago and ensured it was implemented pr default.

Why do you want NOT to close a communication channel on communication errors ? that in many cases leaves the socket in a limbo state, which prohibits further communication.

Remark #1594 is about SERIAL communication, where a device error should not close the usb port….this was solved later.

jameshilliard:
> 1. any api change must be documented in API_CHANGES
> 2. the documentation is not updated

Oh, I thought this was an existing feature that just wasn't really working right.

> I do not understand the purpose of your PR, we removed that parameter quite a while ago and ensured it was implemented pr default.

So that we can have the ability to handle various communication errors in ways other than by closing the channel. How was this parameter previously used?

> Why do you want NOT to close a communication channel on communication errors ?

Because not all communication errors necessitate closing the channel, for example we're using commands like `read_holding_registers` and catching the errors for auto-detection of devices where a communication error may just mean a device is not present. There isn't always a real error which requires a communication channel reset, forced resets on error are problematic as they mess up the auto-detection process by making it more complex and slow due to unnecessary resets.

> that in many cases leaves the socket in a limbo state, which prohibits further communication.

Sure, but there are cases where I think it makes more sense to trigger resetting the communication channel application side(due to having more specific knowledge of what specific error conditions require a reset).

> Remark https://github.com/pymodbus-dev/pymodbus/issues/1594 is about SERIAL communication, where a device error should not close the usb port….this was solved later.

Well in our case it's physical serial ports on industrial PC's, isn't that basically the same issue though, just going through USB adapters instead or something?

janiversen:
The easiest way to implement autodetect is probably in rtu_framer.py and then add a calling method in the serial client, at framer level you have the control you need.

jameshilliard:
> hmm, when you add a parameter, it`s hard to see why not do document it.

Well it's documented in docstrings in a number of places, where does it still need documenting? I'll try and add more docs.

> Your example is valid and just about the only reason for not closing a serial port, however your change is in base.py and thus affects sockets as well.

I think it would still be needed for autodetection logic when using say a serial port over TCP as well so I don't see why it would make sense to restrict this flag to just serial ports.

> I still cannot see what cases that might be ?? Apart from auto detect. In general we try to simplify the API, and transport errors is deep down and should not be exposed on the API. If you want a special transport layer, then simply inherit and make your own transport implementation

So I could see cases involving buggy slave device firmware that under some conditions may say have timeout issues or other communication issues where a transport reset may not be required or helpful. Transport errors triggering connection resets seem to leak through the API already however as they can break other pending asyncio tasks using the same connection.

> There is a world of difference between pymodbus controlling a serial connection (through USB or direct com( and pymodbus connecting to a tcp connected device!

So if I were to run say ser2net on my industrial PC and connect to that from python over tcp then I would still need the parameter to prevent closing the tcp connection on timeout for autodetection I think.

> In a running system, communication errors should always restart the communication, so the parameter is not needed for that.

I'm somewhat skeptical that protocol errors should *always* reset the underlying transport, I mean protocol errors can happen for many reasons that may be entirely unrelated to the underlying transport(ie slave device firmware bugs).

> Adding a autodetect method to the serial client makes a of sense, and is something that have been in our plans for a long time, so please think about changing the PR in that direction.

I think adding that makes sense, but it doesn't really seem like it would be a complete substitution for this particular flag which may have other use cases.

> The easiest way to implement autodetect is probably in rtu_framer.py and then add a calling method in the serial client, at framer level you have the control you need.

Well it should be supported in clients other than the serial client right?

janiversen:
> I think it would still be needed for autodetection logic when using say a serial port over TCP as well so I don't see why it would make sense to restrict this flag to just serial ports.

No in that case you have a socket connection to the repeater, the repeater is responsible for reporting slave failures properly,


> 

> So I could see cases involving buggy slave device firmware that under some conditions may say have timeout issues or other communication issues where a transport reset may not be required or helpful. Transport errors triggering connection resets seem to leak through the API already however as they can break other pending asyncio tasks using the same connection.

You have to ask differently, why do the communication reset disturb, opening a serial port is a matter of less than a couple of milliseconds and a socket with connection is typically between 10 and 100ms. So for production I really do not see the disturbance.

> So if I were to run say ser2net on my industrial PC and connect to that from python over tcp then I would still need the parameter to prevent closing the tcp connection on timeout for autodetection I think.

If that is the case then ser2net do not handle the slaves in a proper way.

> 

> I'm somewhat skeptical that protocol errors should *always* reset the underlying transport, I mean protocol errors can happen for many reasons that may be entirely unrelated to the underlying transport(ie slave device firmware bugs).

It is a tradeoff between keeping the API simple and providing full flexibility. You can always make your own framer or transport to support special cases.

The nullmodem is such an example.

> I think adding that makes sense, but it doesn't really seem like it would be a complete substitution for this particular flag which may have other use cases.

I disagree, in a running system the client app, should not bother about all kind of low level failures, which is why we e.g. have automatic retry and reconnect.

> Well it should be supported in clients other than the serial client right?

I don't think so. Autodetect really only makes sense on a multidrop line, and the only multidrop line supported is serial. In case of a repeater, I would not trust autodetect of slaves, because you have no idea how the repeater works, I know at least one repeater that reboots if you send a request to an unknown (as in not configured) slave...do not forget the repeaters are configured, autodetect is very very rare, mainly due to the security risk. 

Modbus is the prof. automation world, where security is valued, so normal procedure is fixed configurations and not autodetect.

janiversen:
And just to be clear, Adding close_comm_on_error as a parameter again will take a lot of convincing !

janiversen:
But adding an autodetect() in the serial client, or in all clients would be accepted without much discussion.

jameshilliard:
> No in that case you have a socket connection to the repeater, the repeater is responsible for reporting slave failures properly,

Hmm, what about a TCP socket connection to a simple serial port bridge that doesn't act as a repeater or implement any higher level protocol logic?

> You have to ask differently, why do the communication reset disturb, opening a serial port is a matter of less than a couple of milliseconds and a socket with connection is typically between 10 and 100ms. So for production I really do not see the disturbance.

Well depending on how many devices are being autodetected...that can add up. But I think the bigger issue is more how the closing the transport can mess up concurrent tasks using the same client.

> If that is the case then ser2net do not handle the slaves in a proper way.

AFAIU ser2net is mostly just taking a serial socket and providing a simple tcp socket server on top of it, wouldn't handling slaves effectively be the responsibility of the pymodbus in that case since that's part of the higher level protocol?

> It is a tradeoff between keeping the API simple and providing full flexibility. You can always make your own framer or transport to support special cases.

But the connection close is triggered by the base client and transaction manager...so I'm not sure how a custom framer or transport would really help there?

> The nullmodem is such an example.

I mean that has a close method, but I don't think it would make sense to override that AFAIU since I may still want to actually close the transport when say shutting down the application. It seems to me that one would need to override the callers.

> I disagree, in a running system the client app, should not bother about all kind of low level failures, which is why we e.g. have automatic retry and reconnect.

We have to handle these failures either way, for example by catching and appropriately handling the `ModbusIOException` thrown on a timeout error. From my testing it's much easier to handle a `ModbusIOException` in the client app without a spurious transport reset being triggered than a `ModbusIOException` with a spurious transport reset being triggered.

> I don't think so. Autodetect really only makes sense on a multidrop line, and the only multidrop line supported is serial. In case of a repeater, I would not trust autodetect of slaves, because you have no idea how the repeater works, I know at least one repeater that reboots if you send a request to an unknown (as in not configured) slave...do not forget the repeaters are configured, autodetect is very very rare, mainly due to the security risk.

But serial can be trivially bridged to other transports(ie ser2net), I'm not very familiar with repeaters but presumably there would be cases where a transport reset due to say a timeout is unnecessary. Presumably there are also repeater protocol errors where a transport reset may be undesirable.

> Modbus is the prof. automation world, where security is valued, so normal procedure is fixed configurations and not autodetect.

I mean, in my experience prof. automation world security practices are pretty inconsistent, and I'm not sure how auto-detection of slave devices would meaningfully increase attack surface in most real world cases. Also pymodbus is increasingly being used in home automation like [home-assistant](https://www.home-assistant.io/integrations/modbus/) where auto-detection can be especially handy.

> But adding an autodetect() in the serial client, or in all clients would be accepted without much discussion.

Wouldn't slave device auto-detection not be transport specific since `read_holding_registers` and such isn't transport specific?

janiversen:
> Hmm, what about a TCP socket connection to a simple serial port bridge that doesn't act as a repeater or implement any higher level protocol logic?

In the prof. ones you normally configure the slaves, to avoid someone adds a device undetected.

> Well depending on how many devices are being autodetected...that can add up. But I think the bigger issue is more how closing the transport can mess up concurrent tasks using the same client.

We do not support concurrent usage of the client !

> But the connection close is triggered by the base client and transaction manager...so I'm not sure how a custom framer or transport would really help there?

Because they are below base and both have full control over the transport.

> We have to handle these failures either way, for example by catching and appropriately handling the `ModbusIOException` thrown on a timeout error. From my testing it's much easier to handle a `ModbusIOException` in the client app without a spurious transport reset being triggered than a `ModbusIOException` with a spurious transport reset being triggered.
> 
The spurious transport reset is actually handled internally in transport, with automatic reconnect.


> I mean, in my experience prof. automation world security practices are pretty inconsistent, and I'm not sure how auto-detection of slave devices would meaningfully increase attack surface in most real world cases. Also pymodbus is increasingly being used in home automation like [home-assistant](https://www.home-assistant.io/integrations/modbus/) where auto-detection can be especially handy.

I know that, I am the active maintainer there...HA is not considered prof. like when you automate e.g. a power plant. Actually HA is the reason we have been talking about autodetect for a while.

And just for reference, the modbus integration in HA do NOT depend on close_comm_on_error.

> Wouldn't slave device auto-detection not be transport specific since `read_holding_registers` and such isn't transport specific?

It cannot be at transport level do to the nature of the modbus protocol, it needs to be at framer level.

b-rowan:
Hey @janiversen, I would like to clarify slightly, as it may help with understanding what the actual issue is.

We are using pymodbus to control some (very poorly designed) Chinese made devices, which upon sending a write command (`0x06`) do not return a response, and we have no ability to change this.  There is also a requirement to send many write commands to many devices all on the same loop, maxing out at approximately 240 requests.  Generally I am of the opinion that opening and closing a connection for each one of those connections is a bad idea.  In this case I end up catching the timeout that comes back from the devices, and ignoring it completely, because I already know that they will return no response.

I suppose a better question to ask would be:

Is there a possibility of allowing handling of timeout errors separately from other protocol related errors that would require closing the connection, such as implicitly changing this functionality, or better in my opinion, adding a new flag, `close_connection_on_timeout` that defaults to `True`, but would allow us to solve the problem in this specific case?

If neither of these suggestions seem reasonable, do you have a suggestion for what our best option is here?  You suggested a custom transport above, and I haven't looked into that idea, but is that the ideal way to handle this as a third party?

Awesome library in any case BTW, quite enjoyable to write with.

janiversen:
> I mean, from what it looks like calling async methods in concurrent tasks effectively queues up the requests in the transaction manager or something along those lines and processes them in the order of request sequentially, or am I reading that wrong. In general when I've written custom serial protocol clients I allow async request methods which in the background queue up and process those requests sequentially. For example [this somewhat messy code](https://github.com/jameshilliard/hlk-sw16/blob/0.0.9/hlk_sw16/protocol.py) talks over tcp to a serial protocol bridged via ser2net and will queue the `turn_on` `turn_off` and `status` calls when called via concurrent async tasks and process them sequentially in the background.

That might all be correct, but we still do not support concurrent usage, because there are some side effects that we do not handle...one being that transactions are not supported on serial lines nor in all framer types.

> But how would they know if a close request is due to say a timeout vs say the application actually issuing a close on shutdown for example.

There would be no close call, my suggestion is to add autodetect() at framer level, because that way the autodetect have full control over the transport, and the upper layers do not interfere...of course and autodetect() that calls the framer autodetect() should be added too so that it is available in the client API.

> I'm aware, I also [maintain](https://github.com/home-assistant/core/blob/2024.2.0/homeassistant/components/hlk_sw16/manifest.json#L4) integrations there, I was just using it as an example for where auto-detection may be desirable for pymodbus outside of prof automation contexts.

I have repeatedly written that autodetect is a feature we would like to see, just not how you propose making it (by adding a parameter), we would like to see it as a method in the client api.


> Hmm, so it seems the framer is using synchronous IO in `sendPacket`, is the async client wrapping that I guess? Some things are a bit confusing here to me since it seems there's mixing of async functions with sync IO socket handling when using the async client.

Well that life because we support both sync and async, and try to avoid duplicating code.. and in case of HA you would need to support the sync version for now (it is being converted to async but slowly), for nearly all other cases I would only support async, as we are currently changing framer to be only async, and then replace integrate that info the sync client. The basic idea is that the sync client becomes just a call interface similar to the server.

janiversen:
Hi 

Thanks for the clarification, it seems to me as there is a mixture of 2 issues:
the chinese devices and autodetect.

We should really threat this a 2 different issues, otherwise it will be hard to find a solution.

240 requests,that is surely on multiple lines if serial, so the problem is more distributed, secondly I suppose you mean some of the request do not receive a response. Those 2 facts reduces the overhead of opening and closing heavily.

Be aware that on a single serial line you cannot send multiple requests, but have to wait for a response with timeout from each request (it is a master-slave protocol). Using a repeater does NOT change that fact, and it is sadly a programming error I see often, since you do now know how the slaves are distributed (the repeater can have multiple serial lines) you have to wait after each request.

Another frequent error is to do multiple requests, especially writes, to the same device very fast..a lot of devices, not only chinese ones cannot cope with that, because they need time to update their internal structures.

This is just to give you some background, why your problem, could be forced by the way you handle the devices.


> Is there a possibility of allowing handling of timeout errors separately from other protocol related errors that would require closing the connection, such as implicitly changing this functionality, or better in my opinion, adding a new flag, close_connection_on_timeout that defaults to True, but would allow us to solve the problem in this specific case?

this is currently not possible, and a timeout flag would be a very large PR. because it affect client/server/transaction and transport code, so it is not something I think is viable.

I think there is a general misunderstanding here, you need to wait for the timeout to see if the device is responding...that timeout cannot be less than 300ms, now why is close/open on top of that a problem...on my Raspberry pi, closing/open/connect takes less than 2ms.

The fact that you must wait the timeout but find it unacceptable to add a few milliseconds on top of that is what I do not understand! If (as you write) you know the device will not respond, then do not send the request and save the timeout.

b-rowan:
> 240 requests,that is surely on multiple lines if serial, so the problem is more distributed, secondly I suppose you mean some of the request do not receive a response. Those 2 facts reduces the overhead of opening and closing heavily.

Generally this is across 2 serial lines, so 120 requests per line.

> Be aware that on a single serial line you cannot send multiple requests, but have to wait for a response with timeout from each request (it is a master-slave protocol). Using a repeater does NOT change that fact, and it is sadly a programming error I see often, since you do now know how the slaves are distributed (the repeater can have multiple serial lines) you have to wait after each request.

I am well aware of how the modbus protocol works, that is not the issue, we have a computer with 3 rs485 ports, each of which are being used independently, and 2 of those hold these Chinese devices.  It it not a repeater situation.

> Another frequent error is to do multiple requests, especially writes, to the same device very fast..a lot of devices, not only chinese ones cannot cope with that, because they need time to update their internal structures.

This is another thing we have run into with other devices, the switching time with these is roughly 0.5s, not the 3 second timeout we are providing.

> I think there is a general misunderstanding here, you need to wait for the timeout to see if the device is responding...that timeout cannot be less than 300ms, now why is close/open on top of that a problem...on my Raspberry pi, closing/open/connect takes less than 2ms.
>
> The fact that you must wait the timeout but find it unacceptable to add a few milliseconds on top of that is what I do not understand! If (as you write) you know the device will not respond, then do not send the request and save the timeout.

I suppose in any case this is not planned, that's fine, but it seems like this could easily add a lot of overhead if the sockets need to be closed and reopened every time a timeout occurs.  I digress, @jameshilliard maybe just close?

jameshilliard:
> That might all be correct, but we still do not support concurrent usage

Maybe we need an [`asyncio.Lock`](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock) to wrap the code in [`async_execute`](https://github.com/pymodbus-dev/pymodbus/blob/0ffa7c05b728c22f5d2643be9cd1043985a1bf51/pymodbus/client/base.py#L153-L179) to ensure that concurrent calls to the async client functions serializes the low level request properly?

> because there are some side effects that we do not handle...one being that transactions are not supported on serial lines nor in all framer types.

Hmm, would it be difficult to make transactions work correctly on serial lines and all framers or is there some issue that makes that fundamentally problematic?

> There would be no close call, my suggestion is to add autodetect() at framer level, because that way the autodetect have full control over the transport, and the upper layers do not interfere...of course and autodetect() that calls the framer autodetect() should be added too so that it is available in the client API.

So if I'm understanding your suggestion correctly the `autodetect()` method in the framer() should be an async method that checks if a specific slave device is present similar to `read_holding_registers`?

> I have repeatedly written that autodetect is a feature we would like to see, just not how you propose making it (by adding a parameter), we would like to see it as a method in the client api.

Yeah, I've been trying to get a better understanding of the pymodbus internals(there's a lot going on there) so that I can implement autodetect.

> for nearly all other cases I would only support async, as we are currently changing framer to be only async, and then replace integrate that info the sync client. The basic idea is that the sync client becomes just a call interface similar to the server.

So does that mean the plan is for sync methods in the framers like `sendPacket` will be converted to async methods?

> I think there is a general misunderstanding here, you need to wait for the timeout to see if the device is responding...that timeout cannot be less than 300ms, now why is close/open on top of that a problem...on my Raspberry pi, closing/open/connect takes less than 2ms.
> 
> The fact that you must wait the timeout but find it unacceptable to add a few milliseconds on top of that is what I do not understand! If (as you write) you know the device will not respond, then do not send the request and save the timeout.

Maybe we need a way to issue a one-shot write command that doesn't expect a response and thus doesn't trigger a timeout separate from autodetect?

janiversen:
> Maybe we need an [`asyncio.Lock`](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock) to wrap the code in [`async_execute`](https://github.com/pymodbus-dev/pymodbus/blob/0ffa7c05b728c22f5d2643be9cd1043985a1bf51/pymodbus/client/base.py#L153-L179) to ensure that concurrent calls to the async client functions serializes the low level request properly?

that is something we have left to the app, which are better suited to make the locking it needs.

> Hmm, would it be difficult to make transactions work correctly on serial lines and all framers or is there some issue that makes that fundamentally problematic?

Next to impossible, because the modbus request in a RTU frame do not have a transaction id so there are no way to match a response (In case of a repeater with several serial lines).

The modbus serial protocol do not allow multiple outstanding requests (not even to different slaves) on a single serial line.

This would mean that our transaction handler would have to stack the requests, execute them one by one, and return them,
that is a fairly complicated transaction handler.

> So if I'm understanding your suggestion correctly the `autodetect()` method in the framer() should be an async method that checks if a specific slave device is present similar to `read_holding_registers`?

exactly, because at that level you have the transport object available directly, and wait_for etc.


> Yeah, I've been trying to get a better understanding of the pymodbus internals(there's a lot going on there) so that I can implement autodetect.

If I may suggest, concentrate on rtu_framer.py add autodetect() here and in framer/base.py (dummy) and finally in client/base.py add autodetect() that calls something like framer.autodect().

With that "all" you need is to make what you want in framer/autodetect(). framer/base.py have a callback that is called from transport when data is received, and a send method to send data. Tranport provides other callbacks for connected/disconnected but I do not think you need those.


> So does that mean the plan is for sync methods in the framers like `sendPacket` will be converted to async methods?
>
no




> > I think there is a general misunderstanding here, you need to wait for the timeout to see if the device is responding...that timeout cannot be less than 300ms, now why is close/open on top of that a problem...on my Raspberry pi, closing/open/connect takes less than 2ms.
> > The fact that you must wait the timeout but find it unacceptable to add a few milliseconds on top of that is what I do not understand! If (as you write) you know the device will not respond, then do not send the request and save the timeout.
> 
> Maybe we need a way to issue a one-shot write command that doesn't expect a response and thus doesn't trigger a timeout separate from autodetect?

b-rowan:
> that is broadcast, and all devices on a line are supposed to respond, but I supposed you knew that.

Yeah...  I've seen some of the code in these devices, they are literally rewriting the protocol in very bad C, so there is no broadcast handler...  They will only handle requests sent directly "at" them, if that makes sense. 

For reference -
```C
                if (rs485recvbuf[0] == rs485_add && rs485recvbuf[1] == 0x03 && rs485recvbuf[6] == crc_l && rs485recvbuf[7] == crc_h) {
```

janiversen:
Can we close this PR ?

jameshilliard:
> Can we close this PR ?

Hmm, do we need something like this to do one-shot write commands where no response is expected while not closing the socket or is there a better approach? Maybe a variation of broadcast that lets one set the slave_id?

janiversen:
A variation of broadcast sounds like a good idea, change the parameter to be either a bool (slave id 0) or an int (slave id).

I think it should be quite easy to implement that. Remark that I have an issue open with broadcast that still return bytes.

janiversen:
I have been thinking about your comment “we do not expect a response”… when would that happen ??

The modbus protocol defines that every request must be responded to, even broadcast. So it might be your app do not expect a response but apart from failures there will be a response. Not waiting for a response will create problems in pymodbus. Lets assume this sequence:


send X, no wait for response
Device responds
Send Y; wait for response
App receives response X.

Even in the case the device do not respond:

Send X, do not wait for response
Pymodbus responds with modbusIOException (as a return, not raise)
Send Y; wait for response
Depending on timing APP will receive ModbusIOException, or the response from the device.

jameshilliard:
> I have been thinking about your comment “we do not expect a response”… when would that happen ??

I guess when it's known that a slave will not send a response.

> The modbus protocol defines that every request must be responded to, even broadcast.

Do you mean respond as in take some action or respond as in send a reply? From what I'm reading devices that receive a broadcast should take some action but should not send a reply(since that would mean multiple devices would be attempting to reply all at the same time).

> send X, no wait for response
> Device responds
> Send Y; wait for response
> App receives response X.

I think here then what should happen in this sequence is:
send X, no wait for response
Device responds
Send Y; wait for response
pymodbus ignores response X due to validation failure (I think due to `_validate_response` in transaction.py)
pymodbus waits for next response
App receives response Y and associates it with Send Y.

> Send X, do not wait for response
> Pymodbus responds with modbusIOException (as a return, not raise)

If we tell pymodbus to not wait for a response then why we have it return an error? If no response is expected then I don't see how an error response would make sense.

> Send Y; wait for response
> Depending on timing APP will receive ModbusIOException, or the response from the device.

Why would it receive a `ModbusIOException` unless Send Y times out?

janiversen:
> > I have been thinking about your comment “we do not expect a response”… when would that happen ??
> 
> I guess when it's known that a slave will not send a response.

But that is clear violation of the modbus protocol and pymodbus support the standard, where it is a clear error case if a device do not respond.

Broadcast is not different...the modbus protocol defines that when a device receives a request it must send a response...however broadcast on a multidrop line, can lead to corruptions because multiple devices send at the same time.

In order to avoid sending at the same time (in broadcast) a device should listen if there are traffic on the rs485 before sending....that reduces but do not avoid traffic jam.

> > send X, no wait for response
> > Device responds
> > Send Y; wait for response
> > App receives response X.
> 
> I think here then what should happen in this sequence is: send X, no wait for response Device responds Send Y; wait for response pymodbus ignores response X due to validation failure (I think due to `_validate_response` in transaction.py) pymodbus waits for next response App receives response Y and associates it with Send Y.

Your assumption depends a lot on how the transport / framer is configured !! Modbus is a request/response protocol pr definition.

If the app do not want to see the response,  (which seems very illogical to me....even if you send a write, you would like a confirmation that it worked), then the application should throw it away. In the example you gave for implementing a queue, the application would always wait for a response.

> If we tell pymodbus to not wait for a response then why we have it return an error? If no response is expected then I don't see how an error response would make sense.

Because it is against the protocol, and pymodbus implement the standard protocol.

> Why would it receive a `ModbusIOException` unless Send Y times out?
Beacuse X timed out....and again modbus protocol is a request/response protocol, so every request will either have a timeout or a response pr definition.

janiversen:
Let me just a bit of history.

Broadcast was requested many years ago, by schneider, a company that produced a lot of modbus test equipment. The idea was that you could just connect a modbus device to a tester, and the tester would automatically configure the slave. It actually worked quite well for that purpose.

It was never intended for multidrop lines, that would require changes to the basic nature of the protocol. And with all my years of experience, I have never seen broadcast been used in production environments (but surely somebody have been foolish enough to put it production, with all the negative consequences).

jameshilliard:
> But that is clear violation of the modbus protocol and pymodbus support the standard, where it is a clear error case if a device do not respond.

Yeah, seems to be clearly a firmware bug, looking at ways we might handle this sort of thing without introducing a bunch of special case functions.

> Broadcast is not different...the modbus protocol defines that when a device receives a request it must send a response...however broadcast on a multidrop line, can lead to corruptions because multiple devices send at the same time.

Hmm, based on [these docs](https://www.modbus.org/docs/PI_MBUS_300.pdf) it says "Except for broadcast messages, when a master device sends a query to a slave device it expects a normal response".

> In order to avoid sending at the same time (in broadcast) a device should listen if there are traffic on the rs485 before sending....that reduces but do not avoid traffic jam.

Shouldn't the device just not respond to a broadcast?

> If the app do not want to see the response, (which seems very illogical to me....even if you send a write, you would like a confirmation that it worked), then the application should throw it away. In the example you gave for implementing a queue, the application would always wait for a response.

Going to experiment a bit, but I think the requests in the queue should probably be tagged so that when the request queue is consumed the consumer will check if the request expects a response and if not then have the consumer write the request, complete the future in the queue without waiting for a response and immediately move on to the next queue item for processing. I'm thinking maybe an instance variable attached to `ModbusRequest` for this may be one way of handling this.

> Because it is against the protocol, and pymodbus implement the standard protocol.

Yeah, I'm thinking of a way to refactor things so that this can be feasibly implemented as an override of some sort.

janiversen:
> Hmm, based on [these docs](https://www.modbus.org/docs/PI_MBUS_300.pdf) it says "Except for broadcast messages, when a master device sends a query to a slave device it expects a normal response".

That is the server POW and that does not state that the server receives NO response, but due to traffic jam it can happen.

I wrote about the device POW, which must respond when it receives a request.

> Shouldn't the device just not respond to a broadcast?

No, then autodetection for test devices would not work.

> Going to experiment a bit, but I think the requests in the queue should probably be tagged so that when the request queue is consumed the consumer will check if the request expects a response and if not then have the consumer write the request, complete the future in the queue without waiting for a response and immediately move on to the next queue item for processing. I'm thinking maybe an instance variable attached to `ModbusRequest` for this may be one way of handling this.

Far too complicated for my liking....you could do that a lot easier by added a layer in your app.


> > Because it is against the protocol, and pymodbus implement the standard protocol.
> 
> Yeah, I'm thinking of a way to refactor things so that this can be feasibly implemented as an override of some sort.

Good luck, I suspect that is going to be very hard !! a lot easier to handle in the APP.

Your idea of a queue is something that could make the API better....but breaking the protocol but allowing no responses is really a different ball, and very much against the intentions of pymodbus.

jameshilliard:
> Far too complicated for my liking....you could do that a lot easier by added a layer in your app.

I think I came up with something that does this and simplifies the broadcast handling code a little, see #2016 where I attach an `expect_response` variable to `ModbusRequest` and set it to false for broadcasts. This seems to work better in general even for normal broadcast requests as it gives you per request control over whether or not a response is expected vs setting it at the client level. It also lets us simplify some functions by eliminating the broadcast arg. I don't see how it would really be possible to handle this app side alone as the request/reply handling code is internal to pymodbus.

> Your idea of a queue is something that could make the API better....but breaking the protocol but allowing no responses is really a different ball, and very much against the intentions of pymodbus.

The code for handling no-response requests is basically already in pymodbus, we just need to tweak it a little so that no-response request handling can be applied per-request like in #2016 I think which is beneficial even for fully compliant modbus devices as it gives more control.

Issue #1973, opened 2024-02-07
Improve sync and async type checks instead of Union[sync | async]


laundmo:
### Versions

- Python: 3.10
- OS: irrelevant
- Pymodbus: 3.6.4
- Modbus Hardware (if used): irrelevant

### Pymodbus Specific

irrelevant

### Description

The PR #1842 prematurely merged a set of typhints in a attempt to allow type-checking async code, effectively breaking both sync and async usecases.

The issue with this was discussed inside the PRs later messages, but as it was locked (why? why not just leave it open if theres need to follow up?) i am making this issue to request reverting that PR, while considering a better set of typehints.

## Code and Logs

irrelevant

laundmo:
it seems that since #1878 a possiblity would be to type-hint in the concrete sync and async clients, instead of the generic mixin class

alexrudd2:
> it was locked (why? why not just leave it open if there's need to follow up?)

It wasn't manually locked.  Rather an automatic GitHub Action to prevent comments on old issues, in an effort to push people to using newer code and the Discussions feature.  It's a tradeoff. 🤷

> a possibility would be to type-hint in the concrete sync and async clients, instead of the generic mixin class

I may not know how to attempt that myself, but if you give it a try I may be able to help.

janiversen:
If I may correct #1842, was not prematurely merged ! We knew it was not a perfect solution, but only a step in the direction and I for one, saw it as the better option to merge it instead of continuing waiting for the perfect solution.

I am all ears for a solution (like one in the clients) that do NOT duplicate code ! It is correct that we now have 2 base clients, but copying mixin into the 2 base classes, will double the tests and maintenance, and are sure to provide problems down the road (someone changes e.g. the sync call but not the async call).

With my experience from other projects, the most robust solution would be to have a mixin_generator.py, that generates mixin_async.py and mixin_sync.py.

Seen from my life as maintainer I can live with the very old solution, no types at all, but I highly respect the work of @alexrudd2 and others to make the library type checked and agree it makes the library better.

janiversen:
Just checked to be sure, mypy controls all our examples as well as all our test files, so it cannot be totally broken, otherwise we would have had mypy issues.

What I have learned is that both mypy and ruff are very sensitive to the options used, so please be sure you use the options as we do...we know for a fact that mypy will complain heavily with some options active.

janiversen:
What can I say:

```
    print("connect to server")
    await client.connect()
[...]

PR #1976, opened 2024-02-07
Revert "Make client type annotations compatible with async client usage (#1842)"

alexrudd2:
This reverts commit 01ad131d360cc0c01c2a28300f4a58e046058fc9 (with 2 resolved conflicts)

See discussion in #1842 and issue #1973.

This gets us no close to full type hinting, but does at least fix a bit of breakage :)

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -2,7 +2,6 @@
 from __future__ import annotations
 
 import struct
-from collections.abc import Awaitable
 from enum import Enum
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


janiversen (approved):
OK for me, you can merge if you like

But please consider if this is really the way you want to go, what we have is what is actually being returned.

alexrudd2:
> But please consider if this is really the way you want to go, what we have is what is actually being returned.

Both signatures (with or without `Union`) are wrong, sadly.  At least the revert will show in git history for future attempts to get it right.  🤷🏼‍♂️

alexrudd2:
@laundmo @MAKOMO  any thoughts?

janiversen:
As I wrote in the approval, if you feel it´s correct, then just merge it...and lets move on.

alexrudd2:
Closing in favor of https://github.com/pymodbus-dev/pymodbus/pull/1980

PR #1980, opened 2024-02-08
Make ModbusClientMixin Generic to fix type issues for sync and async

laundmo:
this way, the concrete client can specify the generic response return type.

closes: #1973

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -175,7 +175,7 @@ class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]], ModbusProto
                 f"ERROR: No response received after {self.retries} retries"
             )
 
-        return resp # type: ignore
+        return resp # type: ignore[return-value]
 
     def callback_data(self, data: bytes, addr: tuple | None = None) -> int:
         """Handle received data.
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import asyncio
 import socket
 from dataclasses import dataclass
-from typing import Any, Callable, Type, cast
+from typing import Any, Awaitable, Callable, Type, cast, TypeVar, Generic
 
 from pymodbus.client.mixin import ModbusClientMixin
 from pymodbus.exceptions import ConnectionException, ModbusIOException
@@ -17,7 +17,7 @@ from pymodbus.transport import CommParams, ModbusProtocol
 from pymodbus.utilities import ModbusTransactionState
 
 
-class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
+class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]], ModbusProtocol):
     """**ModbusBaseClient**.
 
     Fixed parameters:
@@ -136,7 +136,7 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
             return 0
         return self.last_frame_end + self.silent_interval
 
-    def execute(self, request: ModbusRequest | None = None) -> ModbusResponse:
+    def execute(self, request: ModbusRequest | None = None):
         """Execute request and get response (call **sync/async**).
 
         :param request: The request to process
@@ -150,7 +150,7 @@ class ModbusBaseClient(ModbusClientMixin, ModbusProtocol):
     # ----------------------------------------------------------------------- #
     # Merged client methods
     # ----------------------------------------------------------------------- #
-    async def async_execute(self, request=None):
+    async def async_execute(self, request=None) -> ModbusResponse:
         """Execute requests asynchronously."""
         request.transaction_id = self.transaction.getNextTID()
         packet = self.framer.buildPacket(request)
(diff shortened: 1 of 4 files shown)


pymodbus/client/base.py
```diff
@@ -60,7 +60,6 @@ def __init__(
         **kwargs: Any,
     ) -> None:
         """Initialize a client instance."""
-        ModbusClientMixin.__init__(self)
```
janiversen, line 63: This is no good, we use the init method to inject special tests. Is there a special reason why mixin no longer can have a init() ?

pymodbus/client/base.py
```diff
@@ -136,7 +135,7 @@ def idle_time(self) -> float:
             return 0
         return self.last_frame_end + self.silent_interval
 
-    def execute(self, request: ModbusRequest | None = None) -> ModbusResponse:
+    def execute(self, request: ModbusRequest | None = None):
```
janiversen, line 138: This looks wrong, why can async_execute have a typed return and execute not ?

janiversen (changes requested):
A very elegant PR, I like it !!

I see that CI is failing with mypy complaining, please fix that, along with the couple of review comments I made.

janiversen:
@alexrudd2 Can you please have a look at this PR, from the typing perspective ? I see it as a very elegant solution to the problem at hand, but you know I am by far not the expert on mypy.

pymodbus/client/mixin.py
```diff
@@ -44,12 +46,7 @@ class ModbusClientMixin:  # pylint: disable=too-many-public-methods
         with await <method> (asynchronous) depending on the client used.
     """
 
-    def __init__(self):
```
janiversen, line 47: Please do not delete init, it is used quite frequently for special testing.

janiversen (changes requested):

janiversen:
Once you feel its ready to go, please re-request review.

pymodbus/client/base.py
laundmo: "self" is the wrong type for that init, now. the alternative would be type: ignore so i'll have to switch to that.

pymodbus/client/base.py
laundmo: async_execute needed a explicitly typed return because it also returns bytes in some cases (a, in my opinion, smelly piece of code, but this isnt the PR for that)

the return of execute is not typed explicitly, as type checkers ca infer that from either the mixin or async_execute trivially.

pymodbus/client/base.py
janiversen: "self" is of course a wrong type, it should be "ModbusBaseClient | ModbusAsyncBaseClient",

pymodbus/client/base.py
janiversen: that does not make sense !

first of all async_execute should never return bytes but always a ModbusResponse
secondly, you have already declared async_execute as only returning a ModbusResponse 

My preference is clearly if a method is typed, then the whole line ! 

Btw async_execute and execute returns exactly the same...so please point at the place where you think bytes are returned.

laundmo: line 163-165 ihn base.py
```py
if self.broadcast_enable and not request.slave_id:
    resp = b"Broadcast write sent - no response expected"
    break
```[

these are the lines in async_execute which cause there to be a bytes return
i declared async_execute as returning ModbusResponse to allow `type: ignore` to work on the return type. Otherwise mypy would complain a lot about this. I do not know why this code was accepted in the first place

Comment (laundmo) [reply pymodbus/client/base.py:139]: > Btw async_execute and execute returns exactly the same

just to be clear: they do not. while async_execute returns a ModbusResponse in most cases, execute returns a Awaitable[ModbusResponse]

in fact, execute returns a Coroutine with the return type ModbusResponse, which is compatible with the Awaitable protocol

Comment (janiversen) [reply pymodbus/client/base.py:139]: I am aware that the async returns an awaitable !!! what I meant is that they return the same data, which should always be 
modbusResponse. It still do not explain why execute() is not return typed.

In case of a broadcast there are no response, but I can see we forgot to update base.py to return None, or I suppose it would be better to return an empty ModbusResponse. Please consider that done, I will make a PR doing that later.

File: pymodbus/client/mixin.py [hunk 4]
```diff
@@ -59,11 +59,13 @@ def execute(
         .. tip::
             Response is not interpreted.
         """
-        return ModbusResponse()
+        raise NotImplementedError(
```
alexrudd2, line 62: Have you considered using Abstract Base Classes?  https://docs.python.org/3/library/abc.html

alexrudd2: Never mind, it annoys `pylint` and I don't feel like suppressing those warnings.

alexrudd2 (approved):
You clearly know more than me about Python types, and this is a good solution to a problem Jan and I wasted a lot of cycles on.  Thank you.

alexrudd2:
Oh, if I had to guess you didn't find `check_ci.sh` or the git hook.  (That should be described in `CONTRIBUTING.md`)  Also, you may want to consider installing the ruff extension into VSCode.

janiversen:
It is in README:

```
Development Instructions
------------------------
The current code base is compatible with python >= 3.8.

Here are some of the common commands to perform a range of activities::

   source .venv/bin/activate   <-- Activate the virtual environment
   ./check_ci.sh               <-- run the same checks as CI runs on a pull request.
```

janiversen:
Let get this merged...

@laundmo I look forward to see followup PR, you really did magic here !

janiversen (approved):
LGTM, thanks a lot.

The open comments are not urgent.

janiversen (approved):
need to please ruff

PR #1981, opened 2024-02-08
Lock async_execute operations for concurrent task safety.

jameshilliard:
I think this should make calling methods operations like `read_holding_registers` from concurrent tasks on the same client safe.


pymodbus/client/base.py
```diff
@@ -152,31 +153,34 @@ def execute(self, request: ModbusRequest | None = None) -> ModbusResponse:
...
-            req = self.build_response(request.transaction_id)
-            if not count or not self.no_resend_on_retry:
-                self.transport_send(packet)
-            if self.broadcast_enable and not request.slave_id:
-                resp = b"Broadcast write sent - no response expected"
-                break
-            try:
-                resp = await asyncio.wait_for(
-                    req, timeout=self.comm_params.timeout_connect
+        async with self.execute_lock:
```
janiversen, line 156: This will work perfectly for serial communication, but you effectively prohibit multiple transactions on tcp/tls/udp.

I might not have expressed myself clearly, we do not support concurrent requests, but that does not mean it does not work, it does on tcp/tls/udp.

Our main reason for not supporting concurrent requests are that we cannot allow it on serial, and the API should be independent of the underlying transport. There are also a couple of corner problems, with e.g. file transfer where we might assemble the file wrongly if the transactions are received in a wrong order.

janiversen (changes requested):

pymodbus/client/base.py
jameshilliard: > This will work perfectly for serial communication, but you effectively prohibit multiple transactions on tcp/tls/udp.

Hmm, I'm thinking there should be a way to restrict this locking to serial transports.

> Our main reason for not supporting concurrent requests are that we cannot allow it on serial, and the API should be independent of the underlying transport.

I mean, if it works on everything except serial then isn't the API going to differ based on the transport with some supporting concurrent requests and others not?

> There are also a couple of corner problems, with e.g. file transfer where we might assemble the file wrongly if the transactions are received in a wrong order.

Is that an issue that applies to all transports?

janiversen: > Hmm, I'm thinking there should be a way to restrict this locking to serial transports.

That is a possibility, as I wrote earlier until now we have left it to the app.

> I mean, if it works on everything except serial then isn't the API going to differ based on the transport with some supporting concurrent requests and others not?

Currently the API works identical...and once again we do NOT support concurrent request, but if someone uses concurrent requests and it works for them, then its ok with us.

Not supporting means:
- We do not accept bug reports on this theme.
- We do not document how to (even though we have a test case that actually does it)

> Is that an issue that applies to all transports?

No only to concurrent calls.

jameshilliard: > That is a possibility, as I wrote earlier until now we have left it to the app.

Leaving critical locking to the app seems like a major footgun to me, I would expect a high level client API like the ones exposed by pymodbus to not have unexpected app side locking requirements to use safely.

> We do not document how to (even though we have a test case that actually does it)

I think if the client doesn't natively implement concurrency protections then it should probably be documented what exactly the application side locking requirements are, but I think it would be simpler and less error prone to just try and implement the protections within pymodbus as it may be difficult to validate that an app implements concurrency protections in a way that actually works correctly under various edge cases.

Part of the issues seems to be that there's no obvious indication that an error occurs due to an unexpected concurrency violation.

alexrudd2: > We have never had bug reports on this issue, so it does not seem to a common problem !
> 
> Using async seems foreign to most of our users, and using advanced async (like asyncio.gather) is rare (so far I have only heard of one user, doing advanced stuff.

FWIW, this was one of two problems that led me into pymodbus development in the first place!  The big one was reconnects failing, but also lack of locking.  I ended up putting locks in all my applications/drivers.  I used exclusively `async`, but only `asyncio.gather()` across multiple drivers.

From the Modbus [spec](https://www.modbus.org/docs/Modbus_Messaging_Implementation_Guide_V1_0b.pdf):

> Normally, on MODBUS serial line a client must send one request at a time. This means
> that the client must wait for the answer to the first request before sending a second
> request. On TCP/MODBUS, several requests can be sent without waiting for a
> confirmation to the same server. The MODBUS/TCP to MODBUS serial line gateway is
> in charge of ensuring compatibility between these two behaviors.
> 
> The number of requests accepted by a server depends on its capacity in term of
> number of resources and size of the TCP windows. In the same way the number of
> transactions initialized simultaneously by a client depends also on its resource capacity.
> This implementation parameter is called "NumberMaxOfClientTransaction" and must
> be described as one of the MODBUS client features. Depending of the device type this
> parameter can take a value from 1 to 16.

alexrudd2:
I lean _slightly_ toward implementing locking on serial code.  The slight difference between TCP/serial usage would not bother me; arguably having a different API is per the spec, anyhow.  

In any case, docs on this issue would be good - I think most users will want to implement application-level locking even for TCP because of the misbehaving servers.

pymodbus/client/base.py
alexrudd2: (Note that James' latest messaged was actually accidentally _edited_ by Jan instead of a reply, so it's now of a mix of both of them - but you can still figure out who's who)

jameshilliard: > FWIW, this was one of two problems that led me into pymodbus development in the first place! The big one was reconnects failing, but also lack of locking.

Yeah, pretty much hitting the exact same two issues for us got me looking at things here. I think pymodbus needs some major refactoring to make the API safer for applications. Application side locking requirements especially are a massive footgun as one may often not be able to identify the issue as a locking issue easily in the first place. IMO having to implement stuff like low level protocol locking application side is a pretty big layering violation and if those requirements can't be eliminated there really should be big red warnings all over the documentation.

jameshilliard:
> I lean _slightly_ toward implementing locking on serial code. The slight difference between TCP/serial usage would not bother me; arguably having a different API is per the spec, anyhow.

I think a queuing approach may be more universal, been experimenting with that sort of things in #1982 a little which seemed to work reasonably well for serial connections at least.

> In any case, docs on this issue would be good - I think most users will want to implement application-level locking even for TCP because of the misbehaving servers. 

I'm pretty skeptical application devs *want* to implement application side locking, I mean there may be cases where that's necessary but I think it should be possible to make the API much more safe so that locking wouldn't be required in most cases.

alexrudd2:
> I'm pretty skeptical application devs _want_ to implement application side locking, I mean there may be cases where that's necessary but I think it should be possible to make the API much more safe so that locking wouldn't be required in most cases.

Fair enough.  I'm hesitating to state a strong opinion because I don't feel qualified enough to make architecture decisions.  :)

pymodbus/client/base.py
jameshilliard: > It actually does not unless you tweak async, because each call is a request with the corresponding response.

I mean it's def possible to make it to handle concurrent requests to the API without application side locking, [here's a WIP protocol library I wrote for a protocol that's kinda similar to modbus](https://github.com/jameshilliard/upb/) in some ways(multiple addressed devices with register access requests and such) that's designed to have an API that's fully safe for concurrent task calls by using internal request queues which I was writing a few years back.

> Using async seems foreign to most of our users, and using advanced async (like asyncio.gather) is rare (so far I have only heard of one user, doing advanced stuff.

I'd hardly call `asyncio.gather` rare, sure it's more advanced but it's pretty commonly used for performance optimization and one would typically expect async python libraries to work as expected with it.

> I have been thinking about where a lock could be implemented inside pymodbus and there really are only 2 places, that does not break a lot of architecture.

I'm thinking a queue technique of some sort may work better since you can then use it to say limit the amount of in flight transactions using the socket framer or something I think. Maybe doing that in transaction makes sense? It's not entirely obvious to me where the best spot to do a queue would be is so I've been kinda experimenting.

janiversen:
I think adding queuing for the framers that do not support transaction ID:
rtu_framer,
ascii_framer,
binary_framer (remark this is about to die, because it´s non standard

A queue would allow to simulate transactions for those framers, and would make the API consistent in respect of concurrency.

There are a lot of misbehaving slaves out there, BUT the library cannot and should not try to solve those problems. Pymodbus implement the modbus standard not deviations, that is a major blocker !

We have in the past rejected pull requests, trying to e.g. implement registers as being 32 bits (A special Dutch wish).

We should one day write a architecture document, but it is actually quite simple:

Client classes
Mixin
transaction
framers
transport

Server classes
Datastore
transaction
framers
transport

Server is solely async, client is sync/async. Note that midterm the internals of the client will be only async.

jameshilliard:
> We have in the past rejected pull requests, trying to e.g. implement registers as being 32 bits (A special Dutch wish).

I wonder...would someone be able to implement something like this in their application by providing a custom framer to pymodbus?

janiversen:
> > We have in the past rejected pull requests, trying to e.g. implement registers as being 32 bits (A special Dutch wish).
> 
> I wonder...would someone be able to implement something like this in their application by providing a custom framer to pymodbus?

Possible is everything...but I will not start guessing what e.g. the assumption that a register is 32 bits, will do to the internals of pymodbus.

But really it is not a problem, because modbus defines a register to be 16 bit, and therefore pymodbus implements 16 bits. Pymodbus is not intended for all the variants out there ! pymodbus implement the standard not everything else.

janiversen:
Still working on this one, or is it being replaced by your queue idea ?

I think adding a lock on async serial makes a lot of sense, and is a simple way to avoid problems.

jameshilliard:
closing in favor of #2030

PR #1983, opened 2024-02-09
Fix incorrect bytearray type hint in diagnostics query

alexrudd2:
(I'm trying `pyright` and it seems to be stricter than `mypy`)

Since #1638 it has not been possible to pass a `bytearray` to `diag_query_data`.

Note:
```
>>> msg = bytearray()
>>> type(msg)
<class 'bytearray'>
>>> isinstance(msg, bytes)
False
```

Also, re-enable DiagnasticStatusResponse enconding test, although it doesn't actually test much, ha.


janiversen (approved):
You are right it's not much of a test, not my focus point at the moment.

Thanks for the fix.

alexrudd2:
Yeah, it's clear nobody has used those functions in years (ever?), so it's a safe starting point to experiment (again) with type checks,

PR #1984, opened 2024-02-09
Remove None default for registering a function

alexrudd2:
Who would possibly try to register `None`? :P

﻿


janiversen (approved):
I know who, and it's not me....but nice to have it removed.

Thanks.

PR #1989, opened 2024-02-09
Upgrade pylint and ruff

alexrudd2:
- Bump pylint
- remove unecessary pylint suppressions
- bump ruff
- typo


janiversen (approved):
LGTM, thanks

janiversen:
The 3.12 CI error, is needed in 3.8. I have so far not found a way around that.

alexrudd2:
> The 3.12 CI error, is needed in 3.8. I have so far not found a way around that.

Interesting, it appears to be a bug in pylint [overlapping exceptions plugin](https://github.com/pylint-dev/pylint/blob/main/pylint/extensions/overlapping_exceptions.py).

PR #1996, opened 2024-02-10
Fix setblocking call

alexrudd2:
`socket.setblocking()` takes `bool` not `0`

https://docs.python.org/3/library/socket.html#socket.socket.setblocking
﻿


janiversen:
Ups, good catch !

PR #1997, opened 2024-02-10
Improve client types

alexrudd2:
The base `connect` method is an empty body that is intended to be over-ridden.*  It implicitly has no return value.  The concrete classes have a return value of `bool`.

Also, there is no reason to call `async_execute` without `request`, so it should not be set to `None`.


alexrudd2:
*it's probably best to use an Abstract Base Class for this, but some of the tests run directly against the base class instead of concrete ones.

janiversen (approved):
LGTM, Thanks.

PR #1999, opened 2024-02-11
Avoid pylint complain about lambda

janiversen:

### DIFF
--- a/pymodbus/datastore/remote.py
+++ b/pymodbus/datastore/remote.py
@@ -73,42 +73,42 @@ class RemoteSlaveContext(ModbusBaseSlaveContext):
         if self.slave:
             kwargs["slave"] = self.slave
         self.__get_callbacks = {
-            "d": lambda a, c: self._client.read_discrete_inputs(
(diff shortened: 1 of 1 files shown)


alexrudd2:
Is your local `pylint` perhaps older?  On CI it's complaining again - about these new `ignores` 😆

janiversen:
I did update my venv yesterday, so that should be as new as it gets.

I did notice that CI complains, but if I remember right only on 3.8.

However I would be more than happy to remove the ignores, when I did so in this PR CI gave me the red light.

janiversen:
```
(pymodbus) jan@piso_macmini pymodbus % pylint --version
pylint 3.0.2
astroid 3.0.1
Python 3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 15.0.0 (clang-1500.1.0.2.5)]
```

```
(pymodbus) jan@piso_macmini pymodbus % python --version
Python 3.11.7
```

```
(pymodbus) jan@piso_macmini pymodbus % ruff --version
ruff 0.2.1
```

PR #2010, opened 2024-02-13
solve more mypy issues with datastore

alexrudd2:
This solves another 15 (counting with `--check-untyped-defs`)

### DIFF
--- a/pymodbus/datastore/simulator.py
+++ b/pymodbus/datastore/simulator.py
@@ -37,7 +37,7 @@ class Cell:
     count_write: int = 0
 
 
-class TextCell:
+class TextCell:  # pylint: disable=too-few-public-methods
     """A textual representation of a single cell."""
 
     type: str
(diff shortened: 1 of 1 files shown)


janiversen:
Now I see where you saw the problems....can you try to remove the ignore, and see if you get a problem with python 3.11 - 3.12...thanks in advance.

janiversen:
I just checked your CI run, it looks a lot as if your dev is not updated (I changed CI a couple of days ago).

janiversen (approved):
The changes are OK. I like the idea of a TextCell.

janiversen:
Just to be sure, I have cleaned out all cache, and CI are running on dev as I write this....if CI turns green, I think the problem is somewhere else...agree ?

janiversen:
dev is green after a new CI run, so I think the problem with "ignore" is an not updated dev branch ?

alexrudd2:
I think there was a cache problem.  Some of the CI runners failed to even have Python installed

PR #2011, opened 2024-02-13
Test if removal of pylint ignore works with CI.

janiversen:
If this turns green, then I have a problem with my venv, which complains heavily.

### DIFF
--- a/pymodbus/datastore/remote.py
+++ b/pymodbus/datastore/remote.py
@@ -73,42 +73,42 @@ class RemoteSlaveContext(ModbusBaseSlaveContext):
         if self.slave:
             kwargs["slave"] = self.slave
         self.__get_callbacks = {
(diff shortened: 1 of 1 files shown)


janiversen:
@alexrudd2 my apologies ! my venv is somehow corrupted, sorry for the inconvenience.

What I do not understand, is how all CI dev runs have been successful since I added the ignores...I think there is a problem in what pylint or ruff actually controls, could that be ??

alexrudd2:
> @alexrudd2 my apologies ! my venv is somehow corrupted, sorry for the inconvenience.
> 
> What I do not understand, is how all CI dev runs have been successful since I added the ignores...I think there is a problem in what pylint or ruff actually controls, could that be ??

This particular one is just a pylint warning, so it does not fail the run.

PR #2013, opened 2024-02-14
Add more client type hints

alexrudd2:
`mypy` is sometimes a little dumb, and can't figure out types that are defined in `__init__()`.  So help it out by explicitly defining them.

126 -> 107 `mypy` problems
﻿


janiversen (approved):
I agree nearly on your statement....I would say it's more than "dumb", many times it do not accept totally legal constructs

PR #2029, opened 2024-02-19
Improve HTTP server type hints

alexrudd2:
This solves another 15 errors with `mypy --checkuntyped defs pymodbus` (and a few for `pyright` that `mypy` does not find).

It also looks like it fixes a real bug with  `time.sleep()` not being called correctly.  (The existing code attempts to call `time.time()` instead.)

### DIFF
--- a/pymodbus/server/simulator/http_server.py
+++ b/pymodbus/server/simulator/http_server.py
@@ -8,6 +8,7 @@ import importlib
 import json
 import os
 from time import time
+from typing import TYPE_CHECKING
 
 
 try:
@@ -16,6 +17,8 @@ try:
     AIOHTTP_MISSING = False
 except ImportError:
     AIOHTTP_MISSING = True
+if TYPE_CHECKING:  # type checkers do not understand the Raise RuntimeError in __init__()
+    from aiohttp import web
 
 from pymodbus.datastore import ModbusServerContext, ModbusSimulatorContext
 from pymodbus.datastore.simulator import Label
--- a/pymodbus/server/simulator/http_server.py
+++ b/pymodbus/server/simulator/http_server.py
@@ -7,7 +7,7 @@ import dataclasses
 import importlib
 import json
 import os
-from time import time
+from time import sleep
 from typing import TYPE_CHECKING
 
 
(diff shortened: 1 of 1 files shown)


pymodbus/server/simulator/http_server.py
```diff
@@ -31,6 +32,9 @@
 )
 
 
+if TYPE_CHECKING:  # type checkers do not understand the Raise RuntimeError in __init__()
```
janiversen, line 35: You invented a different method for importing conditionally (pyserial), we should use the same strategy here.

janiversen (changes requested):

pymodbus/server/simulator/http_server.py
alexrudd2: No, look at the whole import area instead of just the changes.

You remember correctly that I wrapped the conditional imports (for both `pyserial` and `aiohttp` so they can `raise RuntimeError` and warn the user.    This remained unchanged.

The type checkers are not smart enough to realize that the `RuntimeError` will prevent any further code execution.  So they throw "warnings" that later usages of `web` (or `serial`) may be unbound (in the case the import failed).
This new code only serves to "convince" type checkers the import succeeded.   It will never execute at runtime. (`TYPE_CHECKING == False`).

janiversen: Now I see you have done the same in serial...I know why I do not like mypy and companions.

Anyhow it is confusing to have the same import in 2 places can't you do "if type checking else" and have the it as one block ?

alexrudd2: > Anyhow it is confusing to have the same import in 2 places can't you do "if type checking else" and have the it as one block ?

How about this?
```python
try:
    from aiohttp import web

    AIOHTTP_MISSING = False
except ImportError:
    AIOHTTP_MISSING = True
    if TYPE_CHECKING:  # always False at runtime
        # type checkers do not understand the Raise RuntimeError in __init__()
        from aiohttp import web
```

alexrudd2:
Closing in favor of https://github.com/pymodbus-dev/pymodbus/pull/2035

PR #2030, opened 2024-02-19
Add RTU Frame request queue

jameshilliard:
This adds a request queue to the RTU framer as the queue logic requirements from my understanding are largely framer specific.

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -131,6 +131,18 @@ class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]], ModbusProto
         else:
             super().close()
 
+    async def drain(self, cancel_pending: bool = False):
+        """Finish processing any pending requests and prepare to shut down.
+
+        :param cancel_pending: If we should cancel any pending requests
+        """
+        await self.framer.drain(cancel_pending)
+
+    @property
+    def drained(self):
+        """Check if the queue is drained."""
+        return self.framer.drained
+
     def idle_time(self) -> float:
         """Time before initiating next transaction (call **sync**).
 
@@ -148,15 +160,17 @@ class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]], ModbusProto
         :returns: The result of the request execution
         :raises ConnectionException: Check exception text.
         """
-        if not self.transport:
+        if not self.transport and self.reconnect_task is None:
             raise ConnectionException(f"Not connected[{self!s}]")
         return self.async_execute(request)
 
     # ----------------------------------------------------------------------- #
     # Merged client methods
     # ----------------------------------------------------------------------- #
-    async def async_execute(self, request) -> ModbusResponse:
+    async def _async_execute(self, request) -> ModbusResponse:
         """Execute requests asynchronously."""
+        if self.reconnect_task is not None:
+            await self.reconnect_task
         request.transaction_id = self.transaction.getNextTID()
         packet = self.framer.buildPacket(request)
 
@@ -182,6 +196,10 @@ class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]], ModbusProto
             )
 
         return resp  # type: ignore[return-value]
+
+    async def async_execute(self, request) -> ModbusResponse:
+        """Execute requests asynchronously with framer."""
+        return await self.framer.asyncProcess(self._async_execute(request))
 
     def callback_new_connection(self):
         """Call when listener receive new connection request."""
@@ -251,6 +269,10 @@ class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]], ModbusProto
 
     async def __aexit__(self, klass, value, traceback):
         """Implement the client with aexit block."""
+        cancel_pending = False
+        if klass is asyncio.CancelledError:
+            cancel_pending = True
+        await self.drain(cancel_pending)
         self.close()
 
     def __str__(self):
--- a/pymodbus/framer/base.py
+++ b/pymodbus/framer/base.py
@@ -60,6 +60,21 @@ class ModbusFramer:
             return True
         return self._header["uid"] in slaves
 
+    async def asyncProcess(self, exec_request):
+        """Process the request."""
+        return await exec_request
+
+    async def drain(self, cancel_pending: bool = False):
+        """Finish processing any pending requests and prepare to shut down.
+
+        :param cancel_pending: If we should cancel any pending requests
+        """
+
+    @property
+    def drained(self):
+        """Check if the queue is drained."""
+        return True
+
     def sendPacket(self, message):
         """Send packets on the bus.
 
--- a/pymodbus/framer/rtu_framer.py
+++ b/pymodbus/framer/rtu_framer.py
@@ -1,4 +1,9 @@
 """RTU framer."""
+from __future__ import annotations
+
+import asyncio
+import contextlib
+
 # pylint: disable=missing-type-doc
 import struct
 import time
@@ -64,6 +69,8 @@ class ModbusRtuFramer(ModbusFramer):
         self._min_frame_size = 4
         self.function_codes = decoder.lookup.keys() if decoder else {}
         self.message_handler = MessageRTU([0], True)
+        self._queue: asyncio.Queue | None = None
+        self._worker: asyncio.Task | None = None
 
     def decode_data(self, data):
         """Decode data."""
@@ -167,6 +174,67 @@ class ModbusRtuFramer(ModbusFramer):
             Log.debug("Frame advanced, resetting header!!")
             callback(result)  # defer or push to a thread?
 
+    async def asyncProcess(self, exec_request):
+        """Queue the request and start the worker task if needed."""
+        if self._queue is None:
+            self._queue = asyncio.Queue()
+        if self._worker is None:
+            self._worker = asyncio.create_task(self._worker_task())
+        future: asyncio.Future = asyncio.Future()
+        await self._queue.put((future, exec_request))
+        return await future
+
+    async def _worker_task(self):
+        """Worker task to process request queue."""
+        try:
+            while self._queue is not None:
+                future, exec_request = await self._queue.get()
+                try:
+                    future.set_result(await exec_request)
+                except asyncio.CancelledError as e:
+                    future.set_exception(e)
+                    raise e
+                except BaseException as e:
+                    future.set_exception(e)
+                finally:
+                    self._queue.task_done()
+        except asyncio.CancelledError as e:
+            while self._queue is not None:
+                try:
+                    future, _ = self._queue.get_nowait()
+                    future.set_exception(e)
+                    self._queue.task_done()
+                except asyncio.QueueEmpty:
+                    self._queue = None
+                    self._worker = None
+                    break
+            raise e
+
+    async def drain(self, cancel_pending: bool = False):
+        """Finish processing any pending requests and prepare to shut down.
+
+        Drains any remaining requests in the queue and shuts down the worker
+        task.
+
+        :param cancel_pending: If we should cancel any pending requests
+        """
+        worker = self._worker
+        queue = self._queue
+        if cancel_pending and worker is not None:
+            worker.cancel()
+        if queue is not None:
+            await queue.join()
+        if worker is not None and not worker.cancelled():
+            worker.cancel()
+        if worker is not None:
+            with contextlib.suppress(asyncio.CancelledError):
+                await worker
+
+    @property
+    def drained(self):
+        """Check if the queue is drained."""
+        return self._queue is None
+
     def buildPacket(self, message):
         """Create a ready to send modbus packet.
 
@@ -236,3 +304,8 @@ class ModbusRtuFramer(ModbusFramer):
         result = self.client.recv(size)
         self.client.last_frame_end = round(time.time(), 6)
         return result
+
+    def __del__(self):
+        """Cancel the worker task if needed."""
+        if self._worker is not None:
+            self._worker.cancel()


pymodbus/client/base.py
```diff
@@ -143,15 +143,17 @@ def execute(self, request: ModbusRequest | None = None):
...
             raise ConnectionException(f"Not connected[{self!s}]")
         return self.async_execute(request)
 
     # ----------------------------------------------------------------------- #
     # Merged client methods
     # ----------------------------------------------------------------------- #
-    async def async_execute(self, request) -> ModbusResponse:
+    async def _async_execute(self, request) -> ModbusResponse:
         """Execute requests asynchronously."""
+        if self.reconnect_task is not None:
```
janiversen, line 155: If there are not connection the APP should be notified, so that it can decide to stop reading data for a while.

Maybe a stupid example, but my solar inverter closes down every night between a couple of minutes uptown a couple of hours. So reconnect_max is 4 hours. The APP reads a single value every 50ms, think of the backlog even after a couple of minutes. A backlog bigger than just a few entries will cause problems with a lot of devices, because they typically reset the connection when receiving a burst of requests.

janiversen (changes requested):
The queue is nice for a running system, but when not connected it will lead to problems.

pymodbus/client/base.py
```diff
@@ -143,15 +143,17 @@ def execute(self, request: ModbusRequest | None = None):
         :returns: The result of the request execution
         :raises ConnectionException: Check exception text.
         """
-        if not self.transport:
+        if not self.transport and self.reconnect_task is None:
```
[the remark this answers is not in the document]
jameshilliard: > This looks wrong because there are only a queue for serial !

I think this should work for transports without queues as well although it seems to be more problematic with a queue, if we are reconnecting any requests need to wait for the re-connection to complete before they can execute.

> Apart from that, it is important that the APP is informed that we are in a non-operational state, remember reconnect times runs sometimes in hours.

Isn't it informed already via the `ModbusIOException` that gets thrown?

pymodbus/framer/rtu_framer.py
```diff
@@ -63,6 +65,8 @@ def __init__(self, decoder, client=None):
         self._end = b"\x0d\x0a"
         self._min_frame_size = 4
         self.function_codes = decoder.lookup.keys() if decoder else {}
+        self._queue = asyncio.Queue()
```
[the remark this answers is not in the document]
jameshilliard: I could initialize it as none and set it on demand like I am with the worker task.

pymodbus/client/base.py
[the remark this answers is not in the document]
janiversen: > I think this should work for transports without queues as well although it seems to be more problematic with a queue, if we are reconnecting any requests need to wait for the re-connection to complete before they can execute.

It doesn´t at least not consistently.


> Isn't it informed already via the `ModbusIOException` that gets thrown?

That is exactly the point, the Exception do not get thrown when reconnect is in progress !

pymodbus/client/base.py
janiversen: > I mean, the APP will get an exception, but this will give reasonable behavior as the queue processing with effectively halt any further reads until the reconnect_task completes.

Well you blocked the exception, so basically you will just be queuing endlessly.

> Is that like a full transport loss or does it just stop responding to reads?

It disconnects, because it reboots.

> I'm not sure how this APP works...but it would have to be doing something very wrong for the queue to grow unbounded, a normal read loop will await the read(and thus will be blocked until the reconnect task completes if disconnected), a read loop should not create unbounded tasks.

In case of the disconnect (without notifying the APP) a reconnect task is started, and the queue just keeps growing until the inverter accepts a connect, once the connect is accepted, the worker_task will overload the inverter with requests leading to a close of the connection....this continues.

And yes the APP issues several reads independently and then uses async.gather to get the results later.

The whole point of the queue is to allow the APP to make multiple requests in parallel so of course the queue will grow when reconnecting.

pymodbus/client/base.py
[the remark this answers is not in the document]
jameshilliard: > It doesn´t at least not consistently.

Hmm, what issues does it cause exactly?

> That is exactly the point, the Exception do not get thrown when reconnect is in progress !

Yeah, I mean it would be the read that triggered the reconnect essentially that you would use.

pymodbus/client/base.py
jameshilliard: > Well you blocked the exception, so basically you will just be queuing endlessly.

But we do a `await self.reconnect_task`, which will block the read loop and prevent it from queuing endlessly.

> It disconnects, because it reboots.

Is the transport serial or tcp?

> And yes the APP issues several reads independently and then uses async.gather to get the results later.

But you await the gather right which would block until reconnected? So you are still bounded to the amount of independent reads in the gather.

> The whole point of the queue is to allow the APP to make multiple requests in parallel so of course the queue will grow when reconnecting.

Well its purpose is more to safely allow multiple tasks to use the same client, sure it's possible to indefinitely queue requests but I don't see a normal usage pattern that should cause something like that to happen, the queue should generally be bounded however to say the amount of tasks or amount of requests in the awaited gather.

janiversen: > But we do a `await self.reconnect_task`, which will block the read loop and prevent it from queuing endlessly.
> 
As I see the code in the pull request, sure it enters an await....but that means the request is not terminated but pending, which again means an endless number of requests pending the await.

> > It disconnects, because it reboots.
> 
> Is the transport serial or tcp?

For the sake of this discussion it is serial....since this PR is about serial

actually it is tcp to a serial converter placed next to the inverter but tomorrow I might replace the converter with my RPI4 server.

> But you await the gather right which would block until reconnected? So you are still bounded to the amount of independent reads in the gather.

There is a task that creates a request every 50ms, and adds the future to a dynamic list of futures, this is the collection process. 

When the analysis starts it syncs the requests by calling async.gather with the list as it is at that point, but the collection process continues.

Anyhow it is a bigger problem if the APP is not informed about a no-connection state, this goes hand in hand with the APP reconnect_callback at transport layer. The combination allows the APP to pause and resume requests. Without this the queue needs to:
- gather for requests limitations (bursts) in the device.
- remove duplicate requests (issues due to lack of response)
- set an upper limit to ensure the queue is not limitless.

> Well its purpose is more to safely allow multiple tasks to use the same client, sure it's possible to indefinitely queue requests but I don't see a normal usage pattern that should cause something like that to happen, the queue should generally be bounded however to say the amount of tasks or amount of requests in the awaited gather.

We do not define how the APP works, so we need to protect against patterns we have identified.

I think this PR have a lot of merit, but the issue of not reporting no-connecting to the APP is a big concern for me (based on experience).

jameshilliard: > There is a task that creates a request every 50ms, and adds the future to a dynamic list of futures, this is the collection process.

This is giving off major anti-pattern vibes to me. Like if you are reading from a few different registers...you should never make a request to a particular register unless there are no pending identical reqeusts to that same register.

> When the analysis starts it syncs the requests by calling async.gather with the list as it is at that point, but the collection process continues.

So it endlessly creates tasks which than call asyncio.gather on futures? This doesn't sound normal to me, is this code open source somewhere btw? Its somewhat unclear what exactly is going on here since it doesn't sound right to me.

Like I think you could see a similar sort of issue happening with entirely different protocols with this design pattern, for example if you made unbounded requests to a normal http endpoint that say couldn't processes requests fast enough to prevent all your system queues/buffers from filling up...you would see overloading issues, memory leaks and responsiveness issue's due to deep queue depths that would somewhat mirror their modbus counterparts.

I think this is more of a don't footgun yourself by designing your app that will leak resources like this, keep your buffers and queues small/bounded otherwise you might get a [bufferbloat](https://en.wikipedia.org/wiki/Bufferbloat) like affect. I don't see an obvious way of fixing this other than fixing how the APP uses the API.

The purpose of this queue mechanism isn't really to encourage someone to have a large amount/unbounded number of in flight requests so much as make it as safe as possible if it happens and to ensure that the client can be shared among tasks properly without requests/responses getting out of sync. A common design pattern I would expect to be used here would have a tasks for different devices on a multidrop rs485 line using their own read loops and acting somewhat independently from each other.

> remove duplicate requests (issues due to lack of response)

Really I think it's best to just design the app to not make duplicate requests

> Anyhow it is a bigger problem if the APP is not informed about a no-connection state, this goes hand in hand with the APP reconnect_callback at transport layer.

I'm not opposed to adding some notification method, would this just be a client level callback?

> The combination allows the APP to pause and resume requests. 

The app shouldn't normally have to care much beyond using the async API's in a reasonable way, but I could see there being some use cases where a disconnect callback could be handy(such as for triggering some sort of fault notifications/alerts).

> We do not define how the APP works, so we need to protect against patterns we have identified.

Probably a good idea to try and encourage certain design patterns here as relying on a disconnect callback to prevent an unbounded queue seems a little sketchy/unreliable to me, but I could see a disconnect callback or something being a good escape hatch for cases where refactoring may be tricky.

pymodbus/framer/rtu_framer.py
```diff
@@ -104,6 +108,25 @@ def advanceFrame(self):
...
 
+    async def asyncProcess(self, exec_request):
+        """Queue the request and start the worker task if needed."""
+        if self._worker is None:
+            self._worker = asyncio.create_task(self._worker_task())
+        future = asyncio.Future()
+        await self._queue.put((future, exec_request))
+        return await future
+
+    async def _worker_task(self):
```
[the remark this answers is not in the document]
jameshilliard: I added cleanup handling/queue draining.

pymodbus/framer/rtu_framer.py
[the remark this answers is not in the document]
jameshilliard: now initialized to none

jameshilliard:
> You have 5 pull requests open, that seems to be around the same theme...could you please close the ones no longer relevant.

I closed the ones that are no longer relevant.

pymodbus/client/base.py
jameshilliard: > My comments about queuing when not connected stands, the rest of the PR is not blockers but only work to do.

A client should be able to avoid queuing when not connected like this I think:
```python
if client.transport:
    data = await client.read_holding_registers(addr, count, slave=1)
```

pymodbus/client/base.py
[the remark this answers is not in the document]
janiversen: This is still wrong because a reconnect can very well happen within the timeout period.

And with queues it is even more wrong, as one of the ideas of a queue is to buffer calls.

pymodbus/client/base.py
janiversen: This is still wrong, the await does not respect the timeout period, so it might wait forever.

Apart from that with a queue it should be buffered.

pymodbus/client/base.py
```diff
@@ -183,6 +197,10 @@ async def async_execute(self, request) -> ModbusResponse:
 
         return resp  # type: ignore[return-value]
 
+    async def async_execute(self, request) -> ModbusResponse:
+        """Execute requests asynchronously with framer."""
+        return await self.framer.asyncProcess(self._async_execute(request))
```
janiversen, line 202: Seems you are mixing architectural layers ! framer is responsible to build the frame not to execute.

The problem is that as we discussed earlier, the queue is in rtu_framer, but architectural it belongs in the transaction layer, which already have a queue concept in place (even though it depends on the device to handle the actual queue).

pymodbus/client/base.py
```diff
@@ -251,6 +269,10 @@ async def __aenter__(self):
 
     async def __aexit__(self, klass, value, traceback):
         """Implement the client with aexit block."""
+        cancel_pending = False
+        if klass is asyncio.CancelledError:
+            cancel_pending = True
+        await self.drain(cancel_pending)
```
janiversen, line 275: The needs to go into close ! Even though an app calls client.close() there are no limitation that the running loop is not continued, and the client object reused.

pymodbus/framer/base.py
```diff
@@ -60,6 +60,21 @@ def _validate_slave_id(self, slaves: list, single: bool) -> bool:
             return True
         return self._header["uid"] in slaves
 
+    async def asyncProcess(self, exec_request):
```
janiversen, line 63: As written earlier this breaks the architecture layer. However it is not a no-go.

pymodbus/framer/rtu_framer.py
```diff
@@ -167,6 +174,67 @@ def check_frame(self):
             Log.debug("Frame advanced, resetting header!!")
             callback(result)  # defer or push to a thread?
 
+    async def asyncProcess(self, exec_request):
+        """Queue the request and start the worker task if needed."""
+        if self._queue is None:
+            self._queue = asyncio.Queue()
```
janiversen, line 180: This belongs in init().

pymodbus/framer/rtu_framer.py
```diff
@@ -167,6 +174,67 @@ def check_frame(self):
...
+        if self._worker is None:
+            self._worker = asyncio.create_task(self._worker_task())
```
janiversen, line 182: If there is a demand for a constantly running task,, then it needs to be done in init.

pymodbus/framer/rtu_framer.py
```diff
@@ -167,6 +174,67 @@ def check_frame(self):
...
+        future: asyncio.Future = asyncio.Future()
+        await self._queue.put((future, exec_request))
+        return await future
```
janiversen, line 185: This will slow down normal requests, since they will always go into the queue and wait for a task switch.

A better approach would be:
a) execute directly, if nothing is executing
b) if something is executing then put future in queue and await the newly created future
c) execute something
d) pop from queue and release future
e) return with response

That way you do not need the worker task.

pymodbus/framer/rtu_framer.py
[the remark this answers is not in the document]
janiversen: Still a problem, seems the worker task is not needed at all.

pymodbus/framer/rtu_framer.py
```diff
@@ -236,3 +304,8 @@ def recvPacket(self, size):
         result = self.client.recv(size)
         self.client.last_frame_end = round(time.time(), 6)
         return result
+
+    def __del__(self):
```
janiversen, line 308: The belongs in drain(True)

We do not do cleanup in __del__ anywhere in the code, especially because the client object might be reused after a close().

pymodbus/framer/rtu_framer.py
```diff
@@ -236,3 +304,8 @@ def recvPacket(self, size):
...
+        """Cancel the worker task if needed."""
+        if self._worker is not None:
+            self._worker.cancel()
```
janiversen, line 311: This is not sufficient, you need to wait until it is actually cancelled. Especially windows leaves tasks hanging for quite a while.

janiversen (changes requested):
Still some larger changes outstanding.

PR #2032, opened 2024-02-20
Improve types for client

alexrudd2:
Solve another `9` errors with `mypy --check-untyped-defs`

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -216,7 +216,7 @@ class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]], ModbusProto
     # ----------------------------------------------------------------------- #
     # Internal methods
     # ----------------------------------------------------------------------- #
-    def send(self, request):
+    def send(self, request) -> int:  # type: ignore [empty-body]
         """Send request.
 
         :meta private:
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -20,9 +20,9 @@ try:
     PYSERIAL_MISSING = False
 except ImportError:
     PYSERIAL_MISSING = True
-
-if TYPE_CHECKING:
-    import serial  # type checkers do not understand the Raise RuntimeError in __init__()
+    if TYPE_CHECKING:  # always False at runtime
+        # type checkers do not understand the Raise RuntimeError in __init__()
+        import serial
 
 class AsyncModbusSerialClient(ModbusBaseClient, asyncio.Protocol):
     """**AsyncModbusSerialClient**.
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -272,9 +272,7 @@ class ModbusSerialClient(ModbusBaseSyncClient):
         """Read data from the underlying descriptor."""
         super().recv(size)
         if not self.socket:
-            raise ConnectionException(
-                self.__str__()  # pylint: disable=unnecessary-dunder-call
-            )
+            raise ConnectionException(str(self))
         if size is None:
             size = self._wait_for_data()
         if size > self._in_waiting():
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import asyncio
 import time
 from functools import partial
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from pymodbus.client.base import ModbusBaseClient, ModbusBaseSyncClient
 from pymodbus.exceptions import ConnectionException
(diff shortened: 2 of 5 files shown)


pymodbus/client/base.py
```diff
@@ -216,7 +216,7 @@ def build_response(self, tid):
     # ----------------------------------------------------------------------- #
     # Internal methods
     # ----------------------------------------------------------------------- #
-    def send(self, request):
+    def send(self, request) -> int:  # type: ignore [empty-body]
```
janiversen, line 219: Avoid the ignore by adding a dummy return 0, please.

pymodbus/client/serial.py
```diff
@@ -175,7 +177,7 @@ def __init__(
             stopbits=stopbits,
             **kwargs,
         )
-        self.socket = None
+        self.socket: serial.Serial | None = None
```
janiversen, line 180: I have to check this one, I am pretty sure it a leftover from when transport was integrated.

janiversen (changes requested):
Looks ok, with one small change for you, and a line I need to check before approving

pymodbus/client/serial.py
janiversen: Ahh it in the sync code, that is correct.

pymodbus/client/base.py
janiversen: I do wonder why it does not complain on line 225-229, it is exactly the same issue (recv returns bytes)...that just goes to show how (sorry for the expression) how incompetent and dangerous these checkers can be :-)

alexrudd2: `mypy` is not currently checking in CI since the entire definition is un-typed.  It complains if using [`--check-untyped-defs`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-check-untyped-defs).  (`pyright` is even stricter/smarter and complains about things `mypy` misses).

My intention with these PRs is to reduce the number of errors with `--check-untyped-defs` down to zero so we can turn on that feature in `pyproject.toml`.  (It started > 200 and is now ~80)

janiversen: You are doing a great job !!! and I try to faclitate it as much as I can, while using my experience to try and prevent things than can cause bigger problems.

You are running point on these changes, and I accept that our code becomes more C++ like everyday....no harm I actually prefer coding C++.

pymodbus/client/serial.py
alexrudd2: Just FYI, improving this type hint is what enabled the type checkers to find https://github.com/pymodbus-dev/pymodbus/pull/2031

pymodbus/client/base.py
alexrudd2: > Avoid the ignore by adding a dummy return 0, please.

I prefer to keep the type error for any calling code which tries to call `ModbusBaseClient.send()` instead of one of the subclasses.

alexrudd2: Actually, using `AbstractBaseClass` is even better since nobody should ever call this `ModbusBaseClient.send()`.  I tried this and got stuck since many of the tests use it. :(

alexrudd2:
> > I do wonder why it does not complain on line 225-229, it is exactly the same issue (recv returns bytes)...that just goes to show how (sorry for the expression) how incompetent and dangerous these checkers can be :-)

Actually, the situation is more complicated :)

https://github.com/pymodbus-dev/pymodbus/blob/a152762c247301152051bfad59dd2ab0ffab781c/pymodbus/client/base.py#L401-L406 does not seem correct.

If I keep adding type hints (mostly `bytes`), the type checkers eventually find this:
https://github.com/pymodbus-dev/pymodbus/blob/a152762c247301152051bfad59dd2ab0ffab781c/pymodbus/client/serial.py#L230-L232

Quite similar to https://github.com/pymodbus-dev/pymodbus/pull/2031, but I do not see the full picture yet.

pymodbus/client/base.py
alexrudd2: > You are running point on these changes, and I accept that our code becomes more C++ like everyday....no harm I actually prefer coding C++.
image

:)

janiversen: Make an issue regarding abstract and let's get this merged.

janiversen (dismissed):

janiversen:
The recv() is an old relict, which I thought was removed long time ago....the very old code v2x had that for testing reasons I think.

janiversen:
The _in_waiting() looks OK to me, but it the linters do not agree......

janiversen (approved):
You decide when you want to merge this and make more in a new PR

pymodbus/client/base.py
alexrudd2: https://github.com/pymodbus-dev/pymodbus/issues/2036

alexrudd2:
Down to 58!

Issue #1992, opened 2024-02-10
optimize slow tests.


janiversen:
the slowest tests are:

3.53s test_comm.py::TestCommModbusProtocol::test_connected_multiple
3.31s test_comm.py::TestCommModbusProtocol::test_connected
3.01s test_client_sync.py::TestSynchronousClient::test_basic_sync_serial_client
3.00s test_client_sync.py::TestSynchronousClient::test_serial_client_recv
2.52s test_client_server_sync.py::TestClientServerSyncExamples::test_server_client_twice
2.20s test_examples.py::TestSyncExamples::test_client_calls[serial-rtu-localhost]
1.86s test_client_server_sync.py::TestClientServerSyncExamples::test_combinations[serial-rtu-localhost]
1.85s test_examples.py::TestSyncExamples::test_sync_simple_client[serial-rtu-localhost]
1.53s test_examples.py::TestSyncExamples::test_client_calls[tcp-socket-localhost]
1.53s test_examples.py::TestSyncExamples::test_client_calls[udp-socket-localhost]
1.53s test_examples.py::TestSyncExamples::test_sync_simple_client[tcp-socket-localhost]
1.52s call     test/sub_examples/test_examples.py::TestSyncExamples::test_sync_simple_client[udp-socket-localhost]
1.52s call     test/sub_examples/test_client_server_sync.py::TestClientServerSyncExamples::test_combinations[tcp-socket-localhost]

Most likely the problem is missing use of NULL_MODEM (apart from test_connected*)

Another problem is asyncio.sleep(), which can in many cases be changed to a "wait with timeout".

PR #2033, opened 2024-02-20
Simplify asyncio and server list

alexrudd2:
Note: this code is very confusing to me, so @janiversen please review carefully.

Solves another 5 `mypy` errors, and speeds up the Window tests.

Instancing the class `_serverList` from its own method
(1) is... strange?
(2) is unnecessary to use the methods, which are all  `@classmethod`
(3) is unnecessary to keep track of the event loop, since every server class already has `self.loop`
==> Use the class methods directly

The methods `shutdown()` and `server_close()` are essentially redundant.
==> combine

There's no use for `asyncio.wait()` directly after an `await`.
==> remove

`time.sleep()` is inefficient, as described in https://github.com/pymodbus-dev/pymodbus/issues/1992#issue-2128311471)
==> replace with waiting+timeout for the `Future`

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import asyncio
 import os
-import time
 import traceback
 from contextlib import suppress
 
@@ -569,10 +568,6 @@ class _serverList:
         if not cls.active_server:
             raise RuntimeError("ServerAsyncStop called without server task active.")
         await cls.active_server.shutdown()
-        if os.name == "nt":
-            await asyncio.sleep(1)
-        else:
-            await asyncio.sleep(0)
         cls.active_server = None
 
     @classmethod
@@ -584,11 +579,8 @@ class _serverList:
         if not cls.active_server.loop.is_running():
             Log.info("ServerStop called with loop stopped.")
             return
-        asyncio.run_coroutine_threadsafe(cls.async_stop(), cls.active_server.loop)
-        if os.name == "nt":
-            time.sleep(10)
-        else:
-            time.sleep(0.1)
+        future = asyncio.run_coroutine_threadsafe(cls.async_stop(), cls.active_server.loop)
+        future.result(timeout=10 if os.name == 'nt' else 0.1)
 
 
 async def StartAsyncTcpServer(  # pylint: disable=invalid-name,dangerous-default-value
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -554,17 +554,12 @@ class _serverList:
 
     active_server: ModbusTcpServer | ModbusUdpServer | ModbusSerialServer
 
-    def __init__(self, server):
-        """Register new server."""
-        self.server = server
-        self.loop = asyncio.get_event_loop()
-
     @classmethod
     async def run(cls, server, custom_functions):
         """Help starting/stopping server."""
         for func in custom_functions:
             server.decoder.register(func)
-        cls.active_server = _serverList(server)
+        cls.active_server = server
         with suppress(asyncio.exceptions.CancelledError):
             await server.serve_forever()
 
@@ -573,7 +568,7 @@ class _serverList:
         """Wait for server stop."""
         if not cls.active_server:
             raise RuntimeError("ServerAsyncStop called without server task active.")
-        await cls.active_server.server.shutdown()
+        await cls.active_server.shutdown()
         if os.name == "nt":
             await asyncio.sleep(1)
         else:
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -284,10 +284,6 @@ class ModbusBaseServer(ModbusProtocol):
         return ModbusServerRequestHandler(self)
 
     async def shutdown(self):
-        """Shutdown server."""
-        await self.server_close()
-
-    async def server_close(self):
         """Close server."""
         if not self.serving.done():
             self.serving.set_result(True)
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -552,7 +552,7 @@ class _serverList:
     :meta private:
     """
 
-    active_server: ModbusTcpServer | ModbusUdpServer | ModbusSerialServer
+    active_server: ModbusTcpServer | ModbusUdpServer | ModbusSerialServer | None
 
     @classmethod
     async def run(cls, server, custom_functions):
--- a/test/sub_server/test_server_asyncio.py
+++ b/test/sub_server/test_server_asyncio.py
@@ -125,7 +125,7 @@ class TestAsyncioServer:
 
         # teardown
         if self.server is not None:
-            await self.server.server_close()
+            await self.server.shutdown()
             self.server = None
         if self.task is not None:
             await asyncio.sleep(0.1)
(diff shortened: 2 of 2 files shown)


pymodbus/server/async_io.py
```diff
@@ -556,19 +551,14 @@ class _serverList:
     :meta private:
     """
 
-    active_server: ModbusTcpServer | ModbusUdpServer | ModbusSerialServer
-
-    def __init__(self, server):
-        """Register new server."""
-        self.server = server
-        self.loop = asyncio.get_event_loop()
```
janiversen, line 564: We do need to register the server, that is sort of the whole purpose of the class (keep track of multiple servers).

it used in line 596 and can not be undefined (its interesting that the linters do not complain in line 596, since the class no longer have a loop variable.

pymodbus/server/async_io.py
```diff
@@ -556,19 +551,14 @@ class _serverList:
...
+    active_server: ModbusTcpServer | ModbusUdpServer | ModbusSerialServer | None
 
     @classmethod
     async def run(cls, server, custom_functions):
         """Help starting/stopping server."""
         for func in custom_functions:
             server.decoder.register(func)
-        cls.active_server = _serverList(server)
+        cls.active_server = server
```
janiversen, line 561: This is how we build a list of servers, shortcutting that means no list of servers.

pymodbus/server/async_io.py
```diff
@@ -577,11 +567,7 @@ async def async_stop(cls):
         """Wait for server stop."""
         if not cls.active_server:
             raise RuntimeError("ServerAsyncStop called without server task active.")
-        await cls.active_server.server.shutdown()
-        if os.name == "nt":
-            await asyncio.sleep(1)
-        else:
-            await asyncio.sleep(0)
+        await cls.active_server.shutdown()
```
janiversen, line 570: active_server is a _serverlist object, not the real server (name is quite confusing though).

pymodbus/server/async_io.py
```diff
@@ -577,11 +567,7 @@ async def async_stop(cls):
         """Wait for server stop."""
         if not cls.active_server:
             raise RuntimeError("ServerAsyncStop called without server task active.")
-        await cls.active_server.server.shutdown()
-        if os.name == "nt":
```
janiversen, line 581: These are needed to allow the async task to run.

pymodbus/server/async_io.py
```diff
@@ -593,11 +579,8 @@ def stop(cls):
         if not cls.active_server.loop.is_running():
             Log.info("ServerStop called with loop stopped.")
             return
-        asyncio.run_coroutine_threadsafe(cls.async_stop(), cls.active_server.loop)
-        if os.name == "nt":
-            time.sleep(10)
-        else:
-            time.sleep(0.1)
+        future = asyncio.run_coroutine_threadsafe(cls.async_stop(), cls.active_server.loop)
+        future.result(timeout=10 if os.name == 'nt' else 0.1)
```
janiversen, line 583: this one I like, there are a number of places in the code, where this can be used.

test/sub_server/test_server_asyncio.py
```diff
@@ -125,7 +125,7 @@ async def _setup_teardown(self):
 
         # teardown
         if self.server is not None:
-            await self.server.server_close()
+            await self.server.shutdown()
```
janiversen, line 128: We need to test both shutdown and server_close as they are both part of the API.

janiversen: Please leave one for server_close and the rest as shutdown.

janiversen (changes requested):
It looks to me as you have misunderstood what _serverlist does.

_serverlist maintains a list of running servers, allowing us to shutdown all tasks.

pymodbus/server/async_io.py
alexrudd2: They do not complain because each server has already has `self.loop`

https://github.com/pymodbus-dev/pymodbus/blob/fe809d9685a15f2b476b396d1dbee0cfca93b56f/pymodbus/server/async_io.py#L267

pymodbus/server/async_io.py
alexrudd2: This comment and below "These are needed to allow the async task to run" got confused because the diff view does not show the changes well.

Really the changes are this:
```python
-- await cls.active_server.server.shutdown()
++ await cls.active_server.shutdown()
```
The new code shuts down a server correctly since `active_server` is now a `server` object.

```
await cls.active_server.shutdown()
-- if os.name == "nt":
--  await asyncio.sleep(1)
-- else:
--  await asyncio.sleep(0)
```
I do not see why the sleeps are "needed to allow the asyncio tasks to run".  The `await cls.active_server.shutdown()` already yields control flow back to the event loop and the `await asyncio.sleep()` is redundant, no?

test/sub_server/test_server_asyncio.py
alexrudd2: see comment above regarding 0071ac11d0e050b2e168b86e6cde42898855e1a8

alexrudd2:
> We do need to register the server, that is sort of the whole purpose of the class (keep track of multiple servers).

> active_server is a _serverlist object, not the real server (name is quite confusing though).

I was indeed confused here.  The name suggests it's a list, but the code looks more like a mixin than a  list.  (I'm still not sure how the class method `run()` can _instance its own class_!)  Each server already knows the event loop (via `self.loop` in `ModbusBaseServer`) and handles its own start/stop (via `Start/StopServer`).

How do multiple servers work?

If it was something like below I would understand.
```
class _serverlist:
def __init__(self):
    self.servers = []
    self.active_server = None

async def run(cls, server, custom_functions):
   self.servers.append(server)
   self.active_server = server
```

alexrudd2:
I considered making this a discussion/issue, but think you prefer to talk about the code.  You can consider this PR a draft, I guess.  However, it does function ... somehow :)

Each commit is separate, so I'll cherry-pick https://github.com/pymodbus-dev/pymodbus/pull/2033/commits/dbe3cd5626ef841f07d106b3e3073665f54be23c as a separate PR.

pymodbus/server/async_io.py
```diff
@@ -284,10 +283,6 @@ def callback_new_connection(self):
         return ModbusServerRequestHandler(self)
 
     async def shutdown(self):
-        """Shutdown server."""
-        await self.server_close()
-
-    async def server_close(self):
```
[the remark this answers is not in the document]
janiversen: Ups, sorry I am corrected !! you are correct.

pymodbus/server/async_io.py
janiversen: aah...but we still need to register the server (or somehow have a list somewhere).

pymodbus/server/async_io.py
janiversen: We have had severe problems in the past, but not forcing a task switch which is essentially what asyncio.sleep(0) does. Not having a task switch mean e.g. that the task is cancelled but not yet terminated, not always but sometimes (especially on windows).

janiversen: So you are saying there are no reason to keep a list of active servers, but then we could remove _server_list totally.

_server_list was build to maintain a list of active server, which comes handy when debugging...I do not like to keep _server_list without the list.

pymodbus/server/async_io.py
alexrudd2: https://github.com/pymodbus-dev/pymodbus/pull/2034

Most other places in the code have either been removed already, or are only `0.1` seconds.

janiversen:
I understand your confusion, this is code that took quite a while to get right.

StartServer can be called multiple times (we start a async loop each time and call StartAsyncServer)
StartAsyncServer can be called multiple times, but do not return unless terminated ! 

ServerClose and ServerAsyncClose closes the newest started server, but remark it is not a class method, so it have no way of knowing which servers are running... That is where _server_list is used.

_server_list "remembers" the servers started in a class variable.

Run instantiates an object of itself, that is a pattern you see in a lot of code.

janiversen:
To be quite blunt the convenience functions Start* Close* are not what I like, I find it much more correct to instantiate the server object, and work directly.

However removing these function caused quite a fuzz years ago, so they remain, until I find a bigger excuse to make 4.0

alexrudd2:
Do you want me to split out https://github.com/pymodbus-dev/pymodbus/pull/2033/commits/0071ac11d0e050b2e168b86e6cde42898855e1a8?

> So you are saying there are no reason to keep a list of active servers, but then we could remove _server_list totally.
> _server_list was build to maintain a list of active server, which comes handy when debugging...I do not like to keep _server_list without the list.

I do not know enough to have a preference here.  Hopefully I have sparked enough discussion for you to improve this some other way.

janiversen:
Yes please. As we go down the road _server_list needs to die.

PR #2035, opened 2024-02-20
Improve HTTP server type hints

alexrudd2:
Supersedes https://github.com/pymodbus-dev/pymodbus/pull/2029

This solves another 15 errors with `mypy --checkuntyped defs pymodbus` (and a few for `pyright` that `mypy` does not find).

It also looks like it fixes a real bug with  `time.sleep()` not being called correctly.  (The existing code attempts to call `time.time()` instead.)


janiversen (approved):
LGTM, thanks

PR #2042, opened 2024-02-21
DictTransactionManager -> ModbusTransactionManager

alexrudd2:
After https://github.com/pymodbus-dev/pymodbus/pull/1966, there is only `DictTransaction﻿Manager`.

It can be combined with the base class.


janiversen (approved):
LGTM, thanks.

PR #2055, opened 2024-02-24
Reorganize datastore inheritance

alexrudd2:
This solves 8 more `mypy` errors, including a real problem in `ModbusSparseDataBlock`.

The method `default()` (inherited from `ModbusBaseDataBlock`) does not make any sense for a sparse datastore, where `self.values` is a dictionary.  If called, it will break the implementation.
==> Move it to `ModbusSequentialDataBlock`.

Also, `reset()` is defined in `ModbusBaseDataBlock` but overridden by `ModbusSparseDataBlock`.
==> Move it also to `ModbusSequentialDataBlock`.

### DIFF
--- a/pymodbus/datastore/store.py
+++ b/pymodbus/datastore/store.py
@@ -231,7 +231,6 @@ class ModbusSparseDataBlock(BaseModbusDataBlock):
         self._process_values(values)
         self.mutable = mutable
         self.default_value = self.values.copy()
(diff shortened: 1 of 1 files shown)


janiversen (approved):
LGTM, thanks.

However it scares me a bit, that no tests were affected .... we really need a lot more testing.

PR #2057, opened 2024-02-24
Improve payload types

alexrudd2:
Solve another 5 `mypy` errors in `pymodbus.payload`.

Should also be a tiny perfomance boost due to reducing allocations.


janiversen (approved):
LGTM, thanks.

Payload should later be changed to work with mixin, so you e.g can do read_holding and get int32 directly

PR #2060, opened 2024-02-25
Improve factory types

alexrudd2:
Unfortunately `mypy` is not smart enough to understand the Request/Response lists, so use `type: ignore`

Solves another 11 `mypy` errors.

﻿


janiversen (approved):
heh...finally you also saw a limit of mypy :-)

LGTM, thanks.

Ps. Please do a rebase, seems you are a bit behind.

PR #2075, opened 2024-03-04
Fix field encoding of Read File Record Response

justinsg:
The Modbus spec V1.1b3 states the following field order:

1. Function Code (0x14)
2. Resp. data Length (in bytes)
3. Sub-Req. x, File Resp. length (in registers)
4. Sub-Req. x, Reference Type: 0x6 (fixed)
5. Sub-Req. x, Record Data
6. Sub-Req. x+1, ...

Previously, field 4 was before field 3.

### DIFF
--- a/pymodbus/file_message.py
+++ b/pymodbus/file_message.py
@@ -181,7 +181,7 @@ class ReadFileRecordResponse(ModbusResponse):
         total = sum(record.response_length + 1 for record in self.records)
         packet = struct.pack("B", total)
         for record in self.records:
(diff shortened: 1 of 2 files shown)


janiversen (commented):
The changes are ok, but I have 2 concerns:
- it is very uncommon for the modbus org to make incompatible changes, so it confuses me.
- what do we do with all the devices that are on the 1.1 standard?

janiversen (approved):
Just checked that change is from 2012, and thus should be in all devices.

janiversen:
Thanks.

Issue #2078, opened 2024-03-05
Message_Parser example not working on latest pymodbus release


CapraTheBest:
### Versions

- Python: Python 3.8.1
- OS: Windows 11
- Pymodbus: 3.6.5
- Modbus Hardware (if used):

### Pymodbus Specific

- Server: tcp/rtu/ascii - sync/async
- Client: tcp/rtu/ascii - sync/async

### Description

When trying to decode any modbus message using the example called "message_parser.py" I get nothing, even if the message is valid and/or generated manually. I've tried running it with an old version (2.5.3) and there it worked flawlessly.

### Code and Logs
Intended behaviour:
```python
PS > python .\message_parser.py --framer socket -m 0812000000170104148000800080008000800080008000800080002004
================================================================================
Decoding Message b'0812000000170104148000800080008000800080008000800080002004'
================================================================================
ServerDecoder
--------------------------------------------------------------------------------
Unable to parse message - b'\x08\x12\x00\x00\x00\x17\x01\x04\x14\x80\x00\x80\x00\x80\x00\x80\x00\x80\x00\x80\x00\x80\x00\x80\x00\x80\x00 \x04' with <pymodbus.framer.socket_framer.ModbusSocketFramer object at 0x038F6928>
ClientDecoder
--------------------------------------------------------------------------------
name            = ReadInputRegistersResponse
transaction_id  = 0x812
protocol_id     = 0x0
slave_id        = 0x1
skip_encode     = 0x0
check           = 0x0
bits            =
                . []
registers       =
                . [32768, 32768, 32768, 32768, 32768, 32768, 32768, 32768,
                . 32768, 8196]
documentation   = Read/write input registers.

    This function code is used to read from 1 to approx. 125 contiguous
    input registers in a remote device. The Request PDU specifies the
    starting register address and the number of registers. In the PDU
    Registers are addressed starting at zero. Therefore input registers
    numbered 1-16 are addressed as 0-15.

    The requested registers can be found in the .registers list.

```

Actual behaviour:
```python
PS > python .\message_parser.py --framer socket -m 0812000000170104148000800080008000800080008000800080002004
================================================================================
Decoding Message b'0812000000170104148000800080008000800080008000800080002004'
================================================================================
ServerDecoder
--------------------------------------------------------------------------------
[...]

PR #2079, opened 2024-03-05
Updated Message_Parser example

CapraTheBest:
After updating the library to the latest version I started having issues with the Message_Parser example (https://github.com/pymodbus-dev/pymodbus/issues/2078). After looking into it I've found that it was still using old calls. I've removed them and also made the "-log" function work again properly. 
If there are any issues please let me know.


CapraTheBest:
Should be good now.

janiversen (approved):
LGTM, thanks.

PR #2082, opened 2024-03-05
Cleanup dead resetFrame code

alexrudd2:
`mypy` flags the `else` branch as unreachable because the `isinstance` check always passes.  I think this happened in https://github.com/pymodbus-dev/pymodbus/commit/ce67df9bf71c2a7ff8a7263fed1a602dfe49e935 but went unnoticed because the function isn't annotated.

https://github.com/pymodbus-dev/pymodbus/blob/214c7ed4ef27c686ad5d161b78477cc5aa6cb1cc/pymodbus/server/async_io.py#L162-L172

So this PR deletes the entire conditional, as well as the now-useless `reset_frame` flag.

Also, it cleans up the docstring mangled by https://github.com/pymodbus-dev/pymodbus/commit/18aa46349a0ceae65479b4b915f9657b3cb4854

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -146,7 +146,6 @@ class ModbusServerRequestHandler(ModbusProtocol):
         As a result, multiple clients can be interleaved without any
         interference between them.
         """
-        reset_frame = False
         while self.running:
             try:
                 await self.inner_handle()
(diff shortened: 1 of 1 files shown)


janiversen (approved):
LGTM, thanks

PR #2084, opened 2024-03-05
Improve simulator type hints

alexrudd2:
Solves another 3 `mypy` errors.

`config_types` below seemed like a good candidate for using a `TypedDict` (~equivalent to a `struct`)
﻿https://github.com/pymodbus-dev/pymodbus/blob/214c7ed4ef27c686ad5d161b78477cc5aa6cb1cc/pymodbus/datastore/simulator.py#L110-L117

However, I ran into issues:
(1) `mypy` doesn't support using variables as `TypedDict` keys: https://github.com/python/mypy/issues/7178.  This unfortunate deficiency means `TypedDict` is not suitable here until `mypy` improves.
(2) Using `type` as a key is messy, since it's a ["soft keyword"](https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords).  Maybe it would be better to use `kind`?  This should have a benefit beyond `mypy`.

### DIFF
--- a/pymodbus/datastore/simulator.py
+++ b/pymodbus/datastore/simulator.py
@@ -753,6 +753,7 @@ class ModbusSimulatorContext:
 
         :meta private:
         """
+        check: tuple
         if func_code in self._bits_func_code:
             # Bit access
             check = (CellType.BITS, -1)
@@ -762,7 +763,7 @@ class ModbusSimulatorContext:
             check = (CellType.UINT16, CellType.STRING)
             reg_step = 1
         else:
-            check = (CellType.UINT32, CellType.FLOAT32, CellType.STRING)  # type: ignore[assignment]
+            check = (CellType.UINT32, CellType.FLOAT32, CellType.STRING)
             reg_step = 2
 
         for i in range(real_address, real_address + count, reg_step):
--- a/pymodbus/datastore/simulator.py
+++ b/pymodbus/datastore/simulator.py
@@ -107,7 +107,7 @@ class Setup:
         """Initialize."""
         self.runtime = runtime
         self.config = {}
-        self.config_types = {
+        self.config_types: dict[str, dict[str, Any]] = {
             Label.type_bits: {
                 Label.type: CellType.BITS,
                 Label.next: None,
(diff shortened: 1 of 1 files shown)


pymodbus/datastore/simulator.py
```diff
@@ -762,7 +762,7 @@ def validate_type(self, func_code, real_address, count):
             check = (CellType.UINT16, CellType.STRING)
             reg_step = 1
         else:
-            check = (CellType.UINT32, CellType.FLOAT32, CellType.STRING)
+            check = (CellType.UINT32, CellType.FLOAT32, CellType.STRING)  # type: ignore[assignment]
```
janiversen, line 765: That sounds wrong...that you need to ignore the assignment. Line 762 is ok....check contains the legal CellTypes and can be any number (of the valid types).

janiversen (changes requested):

pymodbus/datastore/simulator.py
alexrudd2: `mypy` does not like the fact that one path has a `tuple` with 2 items, and another with 3.

I don't see any meaningful consequence to the warning, so I ignored it.

alexrudd2: Do you prefer the other alternative of annotating `check: tuple` before any of the paths define it?

alexrudd2: > sadly enough it is not possible to annotate without assigning

Why do you say that?  It's perfectly possible with no assignment or `Non`

janiversen: ok, lets see what you can come up with....I thought annotation was only possible with assigning.

pymodbus/datastore/simulator.py
```diff
@@ -748,11 +748,12 @@ def action_uptime(cls, registers, inx, cell, **_kwargs):
     # Internal helper methods
     # --------------------------------------------
 
-    def validate_type(self, func_code, real_address, count):
+    def validate_type(self, func_code, real_address, count) -> bool:
         """Check if request is done against correct type.
 
         :meta private:
         """
+        check: tuple
```
janiversen, line 756: That is surely good news, that did not work 1-2 years ago.

Thanks.

janiversen (approved):

PR #2085, opened 2024-03-05
Use a TypedDict for simulator type hints

alexrudd2:
See https://github.com/pymodbus-dev/pymodbus/pull/2084.
This is mostly a demonstration of how `TypedDict` could be used.

However, it made me question whether we should remove the `Label` abstraction.  
What is the purpose for cases where the label is identical to the key?  e.g. `Label.value == 'value'`.  

Would it not be easier to only use it when they differ e.g. `Label.co_size == 'co size'`?

Could the labels be generated programmatically?  (It looks like replacing a space with an underscore and remove `type_` prefix would work in all cases).

### DIFF
--- a/pymodbus/datastore/simulator.py
+++ b/pymodbus/datastore/simulator.py
@@ -5,7 +5,7 @@ import dataclasses
 import random
 import struct
 from datetime import datetime
-from typing import Any, Callable
+from typing import Any, Callable, TypedDict
 
 
 WORD_SIZE = 16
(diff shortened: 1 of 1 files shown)


janiversen:
The reason for 
```
Label.value = 'value'
```
is so the program can use Label.value, while the configuration json uses 'value', that is needed unless you come up with an alternative.

janiversen:
The label can possible be done programmatically....however Label is an enum, and should stay that way.

Apart from that, the whole simulator configuration is way to complex, and it is on my list to make a new version.

janiversen:
But as usual I am open to suggestions....

janiversen:
Just request review when you think it is ready.

alexrudd2:
I just tried the simulator for the first time today, so hopefully I have a better understanding soon.

PR #2090, opened 2024-03-07
Improve transport types

alexrudd2:
The first commit is straightforward.

For the second one, `mypy` complains that we are passing `asyncio.CancelledError` to `close` which only takes `Exception`.  As of Python 3.8 `asyncio.CancelledError` is [dervied from `BaseException` instead](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError).  It will therefore not be caught or suppressed by anything acting on `Exception`.

https://github.com/pymodbus-dev/pymodbus/blob/7c4a310e10abfc17b42330c3474758b67e50dda8/pymodbus/transport/serialtransport.py#L53-L55

I don't see any use in passing it, so I just used `None`.  The alternative is to just `type: ignore`

### DIFF
--- a/pymodbus/transport/serialtransport.py
+++ b/pymodbus/transport/serialtransport.py
@@ -154,8 +154,8 @@ class SerialTransport(asyncio.Transport):
                     self.intern_write_ready()
                 if self.sync_serial.in_waiting:
                     self.intern_read_ready()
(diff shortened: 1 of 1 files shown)


janiversen (approved):
LGTM, thanks,

PR #2091, opened 2024-03-07
Remove pointless try/except in polling_task

alexrudd2:
See https://github.com/pymodbus-dev/pymodbus/pull/2090.  This PR goes a step further.

This code is a stack of clever hacks.
https://github.com/pymodbus-dev/pymodbus/blob/d6704a2ff9abddd0de29d6d50a1650a9d71c084c/pymodbus/transport/serialtransport.py#L45-L48

The purpose of this code is to cancel `self.poll_task()` and then set it to `None` afterwards.

Because the code is sync, `await` cannot be used to ensure the cancellation.  
Hack #1 - use `ensure_future`

Then `ruff` complains with [`RUF006`](https://docs.astral.sh/ruff/rules/asyncio-dangling-task/).  This could have been safely ignored - nobody cares if the GC cleans up, since it's immediately set to `None`.  Instead...
Hack #2 - use `self.future` to hold a reference.

Next `mypy` complains that `self.future` is untyped.
Hack #3 - add an annotation to the constructor.

Looking at `polling_task`, I see no purpose to catching `asyncio.CancelledTask`.  All it does is call `self.close()`.... which is the very code that cancelled it!
I think the whole `try/Except` and stack of hacks can be removed. :)

### DIFF
--- a/pymodbus/transport/serialtransport.py
+++ b/pymodbus/transport/serialtransport.py
@@ -26,9 +26,8 @@ class SerialTransport(asyncio.Transport):
         self._poll_wait_time = 0.0005
         self.sync_serial.timeout = 0
         self.sync_serial.write_timeout = 0
-        self.future: asyncio.Task | None = None
 
-    def setup(self):
+    def setup(self) -> None:
         """Prepare to read/write."""
         if os.name == "nt" or self.force_poll:
             self.poll_task = asyncio.create_task(self.polling_task())
@@ -44,7 +43,6 @@ class SerialTransport(asyncio.Transport):
         self.flush()
         if self.poll_task:
             self.poll_task.cancel()
-            self.future = asyncio.ensure_future(self.poll_task)
             self.poll_task = None
         else:
             self.async_loop.remove_reader(self.sync_serial.fileno())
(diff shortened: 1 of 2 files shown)


alexrudd2:
If the `try/except` is still wanted, the solution is to just ignore ruff and delete `self.future`
```python
asyncio.ensure_future(self.poll_task)  # noqa: RUF006
```

janiversen (commented):
Your idea is not bad, but it seems you broke something so that the reconnect does not run.

The cancelled most likely comes from the same code, as you write, but it might also come from transport (I think),

janiversen:
If we can avoid the try/cancel I am all in favour, I just cannot see how the pyserial object is being closed if the task is cancelled.

janiversen:
hmmm the failing test in 3.10 is a tcp test, I am confused

alexrudd2:
Is there any way that `poll_task` gets cancelled except explicitly in `self.close()`?

janiversen:
I am not sure if asyncio.protocol cancels running tasks in some cases, apart from that I think you are correct that it is only self.close() that does it.

alexrudd2:
> hmmm the failing test in 3.10 is a tcp test, I am confused

I've noticed we have a rare flaky test again, but have not investigated further.

janiversen:
I will keep an eye open, I have had seen it lately.

PR #2094, opened 2024-03-07
Remove dead code from framer

alexrudd2:
`get_expected_response_length` was totally removed in https://github.com/pymodbus-dev/pymodbus/commit/9cf07243838c12563b2ad710efbb70fb1b0c0959#diff-126482fdc00ef924006c76c322c03c809c8b598b66f0af83ab9f5aade84a6a11L314-L323 and therefore this entire branch is useless.

﻿

### DIFF
--- a/pymodbus/transaction.py
+++ b/pymodbus/transaction.py
@@ -13,7 +13,6 @@ __all__ = [
 # pylint: disable=missing-type-doc
 import struct
 import time
-from contextlib import suppress
 from functools import partial
 from threading import RLock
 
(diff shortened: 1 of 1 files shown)


pymodbus/transaction.py
```diff
@@ -379,17 +378,6 @@ def _recv(self, expected_response_length, full):  # noqa: C901
...
                         length = struct.unpack(">H", read_min[4:6])[0] - 1
                         expected_response_length = h_size + length
-                    elif expected_response_length is None and isinstance(
-                        self.client.framer, ModbusRtuFramer
-                    ):
-                        with suppress(
-                            IndexError  # response length indeterminate with available bytes
-                        ):
-                            expected_response_length = (
-                                self.client.framer.get_expected_response_length(
```
janiversen, line 389: Actually I think this needs to be changed to calling these 2 methods:
```
pdu_class = self.decoder.lookupPduClass(func_code)
         return pdu_class.calculateRtuFrameSize(data)
```
I am quite sure I should not have removed it totally.

janiversen (changes requested):

janiversen:
Done (sadly enough without a PR, due to a lapse in mind).

Issue #1034, opened 2022-08-11
Modbus diagnostic calls only works with ModbusSocketFramer


janiversen:
using the new test_examples.py, it became clear that tcp/udp have problems with diagnostic calls when not using ModbusSocketFramer.

Problems are clearly with ModbusRtuFramer as well as ModbusAsciiFramer.

### Versions

* Python:
* OS:
* Pymodbus:
* Modbus Hardware (if used): 

### Pymodbus Specific
* Server: tcp/rtu/ascii - sync/async
* Client: tcp/rtu/ascii - sync/async

### Description

What were you trying, what has happened, what went wrong, and what did you expect?

### Code and Logs

```python
 # code and logs here.

```

janiversen:
Solved.

PR #2096, opened 2024-03-07
enable `mypy --check-untyped-defs`

alexrudd2:
image

When I started, there were over >150 `mypy` errors in untyped definitions.  They have all been slowly solved.

Remaining ignores:
* for `diag_message`, `pyright` is smart enough to infer `int`, but `mypy` is not.  Actually, this code has a lot of type problems and needs a closer look
* `ascii_framer` is going to be deleted soon.  (See https://github.com/pymodbus-dev/pymodbus/pull/2086)
* the `transaction` code also has messy types but is being reworked

### DIFF
--- a/pymodbus/diag_message.py
+++ b/pymodbus/diag_message.py
@@ -382,7 +382,7 @@ class ChangeAsciiInputDelimiterRequest(DiagnosticStatusSimpleRequest):
 
         :returns: The initialized response message
         """
-        char = (self.message & 0xFF00) >> 8
+        char = (self.message & 0xFF00) >> 8  # type: ignore[operator]
         _MCB._setDelimiter(char)  # pylint: disable=protected-access
         return ChangeAsciiInputDelimiterResponse(self.message)
 
--- a/pymodbus/transaction.py
+++ b/pymodbus/transaction.py
@@ -241,7 +241,7 @@ class ModbusTransactionManager:
                                 "/Unable to decode response"
                             )
                             response = ModbusIOException(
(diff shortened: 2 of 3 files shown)


janiversen:
The ascii_framer is not being removed.

Be very careful when you rework transaction, it has a lot of strange (but needed) code....and it seems I cut a little bit to deep, when removing a function in rtu_framer.

The request/response classes surely needs a refactoring, but that is not on my short list.

alexrudd2:
> The ascii_framer is not being removed.

I meant the type error (which is now already gone).

> Be very careful when you rework transaction, it has a lot of strange (but needed) code....and it seems I cut a little bit to deep, when removing a function in rtu_framer.

Yeah, your fixing the code satisfied `mypy`.   I do not intend to do further changes right now.  My goal of `--no-untyped-defs` is achieved with this PR.

alexrudd2:
> The request/response classes surely needs a refactoring, but that is not on my short list.

I believe this is already mentioned in https://github.com/pymodbus-dev/pymodbus/issues/1034

janiversen:
No #1034 is specifically about DIAG, my comment was about all the request/response classes.

I am f.x. convinced it would make the code simpler not to have separate a request and response class for each function code, but as wrote that is for later...right now my focus is the message layer, which is to followed by a new transaction layer with proper locking.

alexrudd2:
@laundmo has the typing situation improved for you with these recent changes?

For me all the `mypy` errors are cleared, and about 50% of the `pyright` ones.

PR #2097, opened 2024-03-07
Remove pointless sleep

### LINKED ISSUE #1998: Remove asyncio.sleep from production code (not test/examples)
Most places where there a asyncio.sleep() it in reality to wait for something to finish, this can mostly be replaced by "wait with timeout".

alexrudd2:
Closes #1998﻿

### DIFF
--- a/pymodbus/server/simulator/http_server.py
+++ b/pymodbus/server/simulator/http_server.py
@@ -260,7 +260,6 @@ class ModbusSimulatorServer:
         self.site = None
         if not self.serving.done():
             self.serving.set_result(True)
-        await asyncio.sleep(0)
 
     async def handle_html_static(self, request):
         """Handle static html."""


pymodbus/server/simulator/http_server.py
```diff
@@ -260,7 +260,6 @@ async def stop(self):
         self.site = None
         if not self.serving.done():
             self.serving.set_result(True)
-        await asyncio.sleep(0)
```
janiversen, line 263: As I just wrote on the issue, this used to be needed to allow other asyncio task to run (in this case the effect of the future set above).

A larger number of the old sporadic test errors was due to the fact that other tasks did not get a chance to run, and I am pretty sure this is the same here.

janiversen (changes requested):

janiversen:
Just checked on the 3.8 pytest error....this is a real bug, it means the reconnect task did not run, seems to be caused by some of the later changes (within a day or two).

pymodbus/server/simulator/http_server.py
alexrudd2: Sure, but look in context - it's the very last line of the function, so it would yield to other tasks anyways.

The only possible us would be hoping some other task runs before `await close()` returns...
(1) I don't see any calling code that would need this
(2) It's a very strange pattern anyways - if `close()` needs some other task to run it should just `await` that task.

janiversen: At least 3.8 do not guarantee a task switch at return, and I think it is a pattern that are kept in newer versions as well.

It is easy for you to say that close() should "just" await....but often it does not know the task, like in this case, who is waiting on the future.

Remember, when a function calls "await stop()" it expect everything to have been cleaned up, without the sleep(0) the future might still be dangling (this is not theoretical I hunted these bugs, when cleaning up sporadic test errors).

alexrudd2: OK, in that case I will close this PR (and the issue).

PR #2127, opened 2024-03-28
async datastore in modbus server

ilkka-ollakka:
Changed datastore to be used by async-flow by modbus server, allowing datastores to be async related to getValues/setValues use.

### DIFF
--- a/pymodbus/bit_read_message.py
+++ b/pymodbus/bit_read_message.py
@@ -155,7 +155,7 @@ class ReadCoilsRequest(ReadBitsRequestBase):
         """
         ReadBitsRequestBase.__init__(self, address, count, slave, **kwargs)
 
-    def execute(self, context):
+    async def execute(self, context):
         """Run a read coils request against a datastore.
 
         Before running the request, we make sure that the request is in
@@ -223,7 +223,7 @@ class ReadDiscreteInputsRequest(ReadBitsRequestBase):
         """
         ReadBitsRequestBase.__init__(self, address, count, slave, **kwargs)
 
-    def execute(self, context):
+    async def execute(self, context):
         """Run a read discrete input request against a datastore.
 
         Before running the request, we make sure that the request is in
--- a/pymodbus/bit_read_message.py
+++ b/pymodbus/bit_read_message.py
@@ -169,7 +169,9 @@ class ReadCoilsRequest(ReadBitsRequestBase):
             return self.doException(merror.IllegalValue)
         if not context.validate(self.function_code, self.address, self.count):
             return self.doException(merror.IllegalAddress)
-        values = context.getValues(self.function_code, self.address, self.count)
+        values = await context.async_getValues(
+            self.function_code, self.address, self.count
+        )
         if isinstance(values, ExceptionResponse):
             return values
         return ReadCoilsResponse(values)
@@ -237,7 +239,9 @@ class ReadDiscreteInputsRequest(ReadBitsRequestBase):
             return self.doException(merror.IllegalValue)
         if not context.validate(self.function_code, self.address, self.count):
             return self.doException(merror.IllegalAddress)
-        values = context.getValues(self.function_code, self.address, self.count)
+        values = await context.async_getValues(
+            self.function_code, self.address, self.count
+        )
         if isinstance(values, ExceptionResponse):
             return values
         return ReadDiscreteInputsResponse(values)
--- a/pymodbus/datastore/context.py
+++ b/pymodbus/datastore/context.py
@@ -1,18 +1,19 @@
 """Context for datastore."""
+
 # pylint: disable=missing-type-doc
 from pymodbus.datastore.store import ModbusSequentialDataBlock
 from pymodbus.exceptions import NoSuchSlaveException
 from pymodbus.logging import Log
 
 
-class ModbusBaseSlaveContext:  # pylint: disable=too-few-public-methods
+class ModbusBaseSlaveContext:
     """Interface for a modbus slave data context.
 
     Derived classes must implemented the following methods:
             reset(self)
             validate(self, fx, address, count=1)
-            getValues(self, fx, address, count=1)
-            setValues(self, fx, address, values)
+            getValues(self, fc_as_hex, address, count=1) or async_getValues(self, fc_as_hex, address, count=1)
+            setValues(self, fc_as_hex, address, values) or async_setValues(self, fc_as_hex, address, values)
     """
 
     _fx_mapper = {2: "d", 4: "i"}
@@ -26,6 +27,45 @@ class ModbusBaseSlaveContext:  # pylint: disable=too-few-public-methods
         :returns: one of [d(iscretes),i(nputs),h(olding),c(oils)
         """
         return self._fx_mapper[fx]
+
+    async def async_getValues(self, fc_as_hex, address, count=1):
+        """Get `count` values from datastore.
+
+        :param fc_as_hex: The function we are working with
+        :param address: The starting address
+        :param count: The number of values to retrieve
+        :returns: The requested values from a:a+c
+        """
+        return self.getValues(fc_as_hex, address, count)
+
+    async def async_setValues(self, fc_as_hex, address, values):
+        """Set the datastore with the supplied values.
+
+        :param fc_as_hex: The function we are working with
+        :param address: The starting address
+        :param values: The new values to be set
+        """
+        return self.setValues(fc_as_hex, address, values)
+
+    def getValues(self, fc_as_hex, address, count=1):
+        """Get `count` values from datastore.
+
+        :param fc_as_hex: The function we are working with
+        :param address: The starting address
+        :param count: The number of values to retrieve
+        :returns: The requested values from a:a+c
+        """
+        del fc_as_hex, address, count
+        return []
+
+    def setValues(self, fc_as_hex, address, values):
+        """Set the datastore with the supplied values.
+
+        :param fc_as_hex: The function we are working with
+        :param address: The starting address
+        :param values: The new values to be set
+        """
+        del fc_as_hex, address, values
 
 
 # ---------------------------------------------------------------------------#
--- a/pymodbus/datastore/remote.py
+++ b/pymodbus/datastore/remote.py
@@ -1,4 +1,5 @@
 """Remote datastore."""
+
 from pymodbus.datastore import ModbusBaseSlaveContext
 from pymodbus.exceptions import NotImplementedException
 from pymodbus.logging import Log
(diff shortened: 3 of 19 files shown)


pymodbus/bit_read_message.py
```diff
@@ -155,7 +155,7 @@ def __init__(self, address=None, count=None, slave=0, **kwargs):
         """
         ReadBitsRequestBase.__init__(self, address, count, slave, **kwargs)
 
-    def execute(self, context):
+    async def execute(self, context):
```
janiversen, line 158: This code is as example used in the client as well, meaning it MUST work for both sync and async.

pymodbus/client/base.py
```diff
@@ -203,7 +205,12 @@ def callback_data(self, data: bytes, addr: tuple | None = None) -> int:
 
         returns number of bytes consumed
         """
-        self.framer.processIncomingPacket(data, self._handle_response, slave=0)
+        task = asyncio.create_task(
```
janiversen, line 208: This is client code and have nothing to do with the datastore which is purely server.

janiversen (changes requested):
The idea of making datastore async is something on our wish list, but I see multiple problems with your PR:
- Client code is changed (datastore is solely for server)
- Async are added in several places, breaking the sync client
- There are far too many files changed for a simple datastore change.

I think a much simpler approach would be to add 2 new function in datastore:
- async_getValues() and async_setValues
That way most of your changes outside datastore is not needed.
The 2 new functions can simply call the old getValues() / setValues()

Apart from that the CI is pretty red, something that needs to be solved.

pymodbus/client/base.py
janiversen: Ahh I see that CI is now Green, one important step done.

You still need to limit your changes, especially so that NO client code is included.

But let me repeat, the idea of the PR is good and very welcome.

janiversen:
Also once the implementation is OK, you need to add test cases, that show the new code works.

pymodbus/client/base.py
ilkka-ollakka: I think I checked that sync client still works as expected, but I agree that change is quite wide.

I'll take another round to check more limited changes and try to reduce touching the client-used code changes.

Thanks for the quick comments.

janiversen: I am pretty sure, that adding 2 new async methods(), will meet your requirements, and limit all changes to the datastore directory + test.

janiversen: request me as reviewer when you are ready.

ilkka-ollakka: I think the new round is closer to your proposal on approach and now it should be contained only on serverside and datastore.

pymodbus/datastore/remote.py
```diff
@@ -38,25 +39,37 @@ def validate(self, _fc_as_hex, _address, _count):
         """
         return True
 
-    def getValues(self, fc_as_hex, _address, _count=1):
+    async def async_getValues(self, fc_as_hex, address, count=1):
```
ilkka-ollakka, line 42: I left these changes as separate commit, incase there is some desire to keep this still intact.

janiversen: it is the other way around, we do not want the cascade changes your original PR contained and still contains.

Having the async_getValues, should solve your problem, so please revert all the cascaded changes in the system (your PR still touches 31 files).

janiversen (changes requested):
review note:

Please remove all the cascaded changes, they are not needed, if you do not change the original methods.

ilkka-ollakka:
I reduced the execute change to be only related the ones that call get/setvalues and server to check if it needs to await the request.

janiversen:
You are still touching 19 files, far too many !

You did not do, what I asked you to do, to make 2 new methods, you renamed the old ones and that causes the cascade effect.

The server code is not ready for an async datastore, until we have removed the sync parts around it. There are a lot more problems than just making the methods async.

However having async methods allowing the datastore to be accessed outside is not a problem.

pymodbus/bit_read_message.py
janiversen, line 158: This is also called from the SYNC client.

pymodbus/bit_read_message.py
```diff
@@ -169,7 +169,9 @@ def execute(self, context):
             return self.doException(merror.IllegalValue)
         if not context.validate(self.function_code, self.address, self.count):
             return self.doException(merror.IllegalAddress)
-        values = context.getValues(self.function_code, self.address, self.count)
+        values = await context.async_getValues(
```
janiversen, line 172: This too is not exclusive for the server.

pymodbus/datastore/context.py
```diff
@@ -27,6 +28,45 @@ def decode(self, fx):
...
+
+    def getValues(self, fc_as_hex, address, count=1):
+        """Get `count` values from datastore.
+
+        :param fc_as_hex: The function we are working with
+        :param address: The starting address
+        :param count: The number of values to retrieve
+        :returns: The requested values from a:a+c
+        """
+        del fc_as_hex, address, count
```
janiversen, line 58: Why delete the variables, that does not make a lot of sense.

pymodbus/datastore/context.py
```diff
@@ -27,6 +28,45 @@ def decode(self, fx):
...
+        return []
+
+    def setValues(self, fc_as_hex, address, values):
+        """Set the datastore with the supplied values.
+
+        :param fc_as_hex: The function we are working with
+        :param address: The starting address
+        :param values: The new values to be set
+        """
+        del fc_as_hex, address, values
```
janiversen, line 68: Same here.

pymodbus/datastore/remote.py
```diff
@@ -38,25 +39,37 @@ def validate(self, _fc_as_hex, _address, _count):
...
         """Get values from real call in validate."""
         if fc_as_hex in self._write_fc:
             return [0]
         group_fx = self.decode(fc_as_hex)
         func_fc = self.__get_callbacks[group_fx]
-        self.result = func_fc(_address, _count)
+        kwargs = {}
+        if self.slave:
+            kwargs["slave"] = self.slave
+        self.result = await getattr(self._client, func_fc)(address, count, **kwargs)
```
janiversen, line 51: This looks wrong to me, the code is designed so we do not need to call getattr, which is very expensive function.

pymodbus/datastore/remote.py
```diff
@@ -74,44 +87,20 @@ def __build_mapping(self):
...
-            "h16": lambda a, v: self._client.write_registers(
-                a, v, **kwargs
-            ),
-            "i6": lambda a, v: self._client.write_register(
-                a, v, **kwargs
-            ),
-            "i16": lambda a, v: self._client.write_registers(
-                a, v, **kwargs
-            ),
+            "d5": "write_coil",
```
janiversen, line 96: This is not the same...but might be the reason I saw a getattr earlier.

We do not want that, apart from that what does it have to do with getValues being async ?

janiversen (changes requested):
Seems you are trying to make unrelated changed in the same PR, and it seems you have not thought about how the sync client works.

janiversen:
your description says:
"Changes codepaths to be async to allow datastore getValues/setValues to be async methods.!"

My suggestion have constantly been, add 2 new methods instead, that will allow you to have async access from outside the server.

You do not write it, but if your purpose is to write a new datastore, you are in for a far bigger job, because the server shares mixin, execute, transaction and framer code with the client both sync and async. So in order to make the datastore async internally, you need to change the server to not share code with any sync part (which includes a couple of examples). This is something that should NOT be done in a single PR, but probably at least 10 PRs

pymodbus/datastore/context.py
ilkka-ollakka: they are not used, so makes it more clear in that point. Mypy and ruff etc easily complain on those. But I agree it is more of taste and not really critical to be in my point of view.

pymodbus/datastore/remote.py
ilkka-ollakka: this is mainly due lamdas and await don't work together. It can be redone without getattr use most likely easily.

ilkka-ollakka:
> your description says: "Changes codepaths to be async to allow datastore getValues/setValues to be async methods.!"
> 

That was my original intent, to get async datastore to server. Not really to have async access from outside to datastore.

Sorry for causing frustration with my miscommunication on original description and intent related to this PR. I understand that it must have been frustrating.

> You do not write it, but if your purpose is to write a new datastore, you are in for a far bigger job, because the server shares mixin, execute, transaction and framer code with the client both sync and async.
> So in order to make the datastore async internally, you need to change the server to not share code with any sync part (which includes a couple of examples). This is something that should NOT be done in a single PR, but probably at least 10 PRs


From those I think only execute is actually one that blocks calling datastore as async way from server?

To be more splitted, would it be suitable approach for example to change ServerDecoder so that it returns AsyncReadCoilsRequest instead of ReadCoilsRequest etc, and those Async ones would be inherited from the original but have async execute? This way sync flow would be not touched, but it allows to take step closer to async datastore in server and work on that side without needing to touch sync execute-path?

I understand that I might not be enough, so more of mapping on things that are lacking and increasing my own understanding of the scope.

janiversen:
> From those I think only execute is actually one that blocks calling datastore as async way from server?

That is basically correct, but execute is also used in the client code.

> To be more splitted, would it be suitable approach for example to change ServerDecoder so that it returns AsyncReadCoilsRequest instead of ReadCoilsRequest etc, and those Async ones would be inherited from the original but have async execute? This way sync flow would be not touched, but it allows to take step closer to async datastore in server and work on that side without needing to touch sync execute-path?

No we do not want to double that amount of Request/Response classes. It is clear that execute should be async when used in asyncio, so maybe a async_execute in a message base could do the trix.....this is just a thought nothing more.

> I understand that I might not be enough, so more of mapping on things that are lacking and increasing my own understanding of the scope.

The easiest path, is to look at asyncio.py, because that is the whole server. You basically need to "pursuade" that file to call async_getValues instead of getValues. That is the best way (if possible), second best is to have an async_execute in 1 message base class.

But please no duplication of code and no cascading changes....allowing that is calling for trouble in future maintenance.

I think it is doable, I looked at it in depth about a year ago, when we wanted to make a better forwarder, but got stopped by more urgent matters.

ilkka-ollakka:
> No we do not want to double that amount of Request/Response classes. It is clear that execute should be async when used in asyncio, so maybe a async_execute in a message base could do the trix.....this is just a thought nothing more.

Ok, that sounds good starting point to recheck things.

> The easiest path, is to look at asyncio.py, because that is the whole server. You basically need to "pursuade" that file to call async_getValues instead of getValues. That is the best way (if possible), second best is to have an async_execute in 1 message base class.

Currently it does call execute and doesn't know anything about getValues/setValues, but it could be easily changed to call that async_execute. But that would cause change in some Request-classes where there is calls to getValues/setValues.
 
> But please no duplication of code and no cascading changes....allowing that is calling for trouble in future maintenance.

I'll check if I can do something to keep async_execute and execute with minimal code duplication.

janiversen:
You might need to look from where the execute is called, if it turns out to be only asyncio.py then things are simpler, but I suspect at least some axamples uses the execute.

janiversen:
A short look at the code, seems to suggest that the execute in the message classes is only used in asyncio.py (it is really a wrong name if that is the case).

If this is the case, the better approach would be to have:
- first PR, that changes the execute in asyncio to be async (hopefully without any side effects)
- second PR, that changes the execute in all message classes to be async (This is a massive change, so it should ONLY contain edits, no functional changes)
- third PR, that changes the datastore.

If you do it all in one PR, it is practically impossible to review, and if/when problems occur down the line, it is very difficult to pin point the cause.

ilkka-ollakka:
I'll split the things to 4 PR's and link those here.

pymodbus/datastore/context.py
janiversen: use "_" that is the standard way to tell a variable is unused.

It might not be critical from your POW, but from me as maintainer it is negative.

ilkka-ollakka: sure, I can change it to be with "_".

ilkka-ollakka:
Intent of this PR is splitted to separated PR's and linked here, so closing this one as it has ran it course

PR #2139, opened 2024-03-31
modbus_server:  execute to handle message execute to be coroutine or method

ilkka-ollakka:
Split from https://github.com/pymodbus-dev/pymodbus/pull/2127

execute() is used as callback, so we don't change it to async, but we use it to create async_task that can handle async calls

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -49,6 +49,7 @@ class ModbusServerRequestHandler(ModbusProtocol):
         self.receive_queue: asyncio.Queue = asyncio.Queue()
         self.handler_task = None  # coroutine to be run on asyncio loop
         self.framer: ModbusFramer
+        self.loop = asyncio.get_running_loop()
 
     def _log_exception(self):
         """Show log exception."""
@@ -173,6 +174,9 @@ class ModbusServerRequestHandler(ModbusProtocol):
         if self.server.request_tracer:
             self.server.request_tracer(request, *addr)
 
+        asyncio.run_coroutine_threadsafe(self._async_execute(request, *addr), self.loop)
+
+    async def _async_execute(self, request, *addr):
         broadcast = False
         try:
             if self.server.broadcast_enable and not request.slave_id:
@@ -181,9 +185,17 @@ class ModbusServerRequestHandler(ModbusProtocol):
                 # note response will be ignored
                 for slave_id in self.server.context.slaves():
                     response = request.execute(self.server.context[slave_id])
+                    # Temporary check while we move execute to async method
+                    if asyncio.iscoroutine(response):
+                        response = await response
             else:
                 context = self.server.context[request.slave_id]
                 response = request.execute(context)
+
+                # Temporary check while we move execute to async method
+                if asyncio.iscoroutine(response):
+                    response = await response
+
         except NoSuchSlaveException:
             Log.error("requested slave does not exist: {}", request.slave_id)
             if self.server.ignore_missing_slaves:


pymodbus/server/async_io.py
```diff
@@ -173,17 +175,28 @@ def execute(self, request, *addr):
         if self.server.request_tracer:
             self.server.request_tracer(request, *addr)
 
+        task = self.loop.create_task(self._async_execute(request, *addr))
```
janiversen, line 178: This is a lot of overhead, creating a task for every request. It is possible and efficient to call async methods from sync methods.

pymodbus/server/async_io.py
```diff
@@ -49,6 +49,8 @@ def __init__(self, owner):
         self.receive_queue: asyncio.Queue = asyncio.Queue()
         self.handler_task = None  # coroutine to be run on asyncio loop
         self.framer: ModbusFramer
+        self.request_tasks: set[asyncio.Task] = set()
```
janiversen, line 52: This is opening for parallel processing in the server, that is one big can of worms, please do not open that.

pymodbus/server/async_io.py
```diff
@@ -173,17 +175,28 @@ def execute(self, request, *addr):
...
         broadcast = False
+        response = None
         try:
             if self.server.broadcast_enable and not request.slave_id:
                 broadcast = True
                 # if broadcasting then execute on all slave contexts,
                 # note response will be ignored
                 for slave_id in self.server.context.slaves():
                     response = request.execute(self.server.context[slave_id])
+                    if asyncio.iscoroutine(response):
```
janiversen, line 192: A short comment telling this is temporary would be nice.

pymodbus/server/async_io.py
```diff
@@ -173,17 +175,28 @@ def execute(self, request, *addr):
...
+                        response = await response
             else:
                 context = self.server.context[request.slave_id]
                 response = request.execute(context)
+                if asyncio.iscoroutine(response):
```
janiversen, line 197: If you move the "if" outside the "else:" you do not need to duplicate the test.

Remark: the "for slave_id" is not the best programming, because it can only be there once.

pymodbus/server/async_io.py
```diff
@@ -196,8 +209,9 @@ def execute(self, request, *addr):
                 traceback.format_exc(),
             )
             response = request.doException(merror.SlaveFailure)
+
         # no response when broadcasting
-        if not broadcast:
+        if not broadcast and response is not None:
```
janiversen, line 214: Why can response be None now, when it wasn't before ?

janiversen (changes requested):
Good PR, the only real problem is the create_task, which we really do not want for performance reasons.

pymodbus/server/async_io.py
ilkka-ollakka: Seems that asyncio documentation seems to prefer create_task, but if you have proposal to alternative approach, I can change it.

pymodbus/server/async_io.py
ilkka-ollakka: I'll add mention about temporary

pymodbus/server/async_io.py
ilkka-ollakka: Ah, good point, I'll move it outside of else block.

ilkka-ollakka: Only problem with it being outside, is that with broadcast it doesn't raise exception in other than last request. But as this being temporary code, I think it is acceptable transition phase?

pymodbus/server/async_io.py
ilkka-ollakka: I set it to None at the beginning, I'll clear it out, so it is closer to original code.

pymodbus/server/async_io.py
ilkka-ollakka: I'll remove use of set

pymodbus/server/async_io.py
ilkka-ollakka: I changed this to run_coroutine_threadsafe, ensure_future seemed to end up using create_task.

pymodbus/server/async_io.py
ilkka-ollakka: I left these in place, mainly so that it still raises exceptions from execute in broadcast when it happens and it would end up leaving other execute calls to be not awaited.

pymodbus/server/async_io.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, owner):
         self.receive_queue: asyncio.Queue = asyncio.Queue()
         self.handler_task = None  # coroutine to be run on asyncio loop
         self.framer: ModbusFramer
+        self.event_loop = asyncio.get_running_loop()
```
janiversen, line 52: Please call the variable self.loop to be consistent with the rest of the code.

janiversen (approved):
One very small change and this seems ready to go.

pymodbus/server/async_io.py
ilkka-ollakka: changed

PR #2141, opened 2024-04-02
test_simulator: use unused_tcp_port fixture

ilkka-ollakka:
default port 8080 can be in use when running test in developer machine, so better to use port that is available


janiversen (approved):
Change looks good to me, please solve your CI error.

janiversen:
Thanks.

PR #2142, opened 2024-04-02
Request/Response: change execute to be async method

ilkka-ollakka:
contains only async/await edits, no functional changes otherwise. This combined with https://github.com/pymodbus-dev/pymodbus/pull/2144 allows to utilize async datastore calls in modbus server

Split from https://github.com/pymodbus-dev/pymodbus/pull/2127 and depends on https://github.com/pymodbus-dev/pymodbus/pull/2139

### DIFF
--- a/test/sub_client/test_client.py
+++ b/test/sub_client/test_client.py
@@ -392,7 +392,7 @@ class MockTransport:
     async def delayed_resp(self):
         """Send a response to a received packet."""
         await asyncio.sleep(0.05)
-        resp = self.req.execute(self.ctx)
+        resp = await self.req.execute(self.ctx)
         pkt = self.base.ctx.framer.buildPacket(resp)
         self.base.ctx.data_received(pkt)
 
--- a/pymodbus/pdu/bit_read_message.py
+++ b/pymodbus/pdu/bit_read_message.py
@@ -155,7 +155,7 @@ class ReadCoilsRequest(ReadBitsRequestBase):
         """
         ReadBitsRequestBase.__init__(self, address, count, slave, **kwargs)
 
-    def execute(self, context):
+    async def execute(self, context):
         """Run a read coils request against a datastore.
 
         Before running the request, we make sure that the request is in
@@ -223,7 +223,7 @@ class ReadDiscreteInputsRequest(ReadBitsRequestBase):
         """
         ReadBitsRequestBase.__init__(self, address, count, slave, **kwargs)
 
-    def execute(self, context):
+    async def execute(self, context):
         """Run a read discrete input request against a datastore.
 
         Before running the request, we make sure that the request is in
--- a/pymodbus/pdu/bit_write_message.py
+++ b/pymodbus/pdu/bit_write_message.py
@@ -79,7 +79,7 @@ class WriteSingleCoilRequest(ModbusRequest):
         self.address, value = struct.unpack(">HH", data)
         self.value = value == ModbusStatus.ON
 
-    def execute(self, context):
+    async def execute(self, context):
         """Run a write coil request against a datastore.
 
         :param context: The datastore to request from
@@ -213,7 +213,7 @@ class WriteMultipleCoilsRequest(ModbusRequest):
         values = unpack_bitstring(data[5:])
         self.values = values[:count]
 
-    def execute(self, context):
+    async def execute(self, context):
         """Run a write coils request against a datastore.
 
         :param context: The datastore to request from
--- a/pymodbus/pdu/diag_message.py
+++ b/pymodbus/pdu/diag_message.py
@@ -199,7 +199,7 @@ class DiagnosticStatusSimpleRequest(DiagnosticStatusRequest):
         DiagnosticStatusRequest.__init__(self, **kwargs)
         self.message = data
 
-    def execute(self, *args):
+    async def execute(self, *args):
         """Raise if not implemented."""
         raise NotImplementedException("Diagnostic Message Has No Execute Method")
 
@@ -245,7 +245,7 @@ class ReturnQueryDataRequest(DiagnosticStatusRequest):
             raise ModbusException(f"message({type(message)}) must be bytes")
         self.message = message
 
-    def execute(self, *_args):
+    async def execute(self, *_args):
         """Execute the loopback request (builds the response).
 
         :returns: The populated loopback response message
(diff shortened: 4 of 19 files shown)


janiversen (changes requested):
This is wrong and will not work for the sync client.

I thought you were working on the changes in asyncio in another PR, where you demonstrated that execute do not need to be async.

ilkka-ollakka:
In theory execute don't need to be async, but some of them are checking results of getValues/setValues, so those should be reworked to be out from execute in that case. I had the impression that only server-code called execute and this was ok approach. But I'll take second look of the comments and check what I missed.

janiversen (changes requested):
And btw it seems you have not analyzed the code. The execute is both sync and async (as in it returns a future).  Execute splits at a lower level to async_execute when called in an async environment

janiversen:
Did you look at how we define the T which is the return from the mixin methods, that secures you can use the modhid as sync/async and satisfy mypy.

janiversen:
you might be right that the execute in the different messages is only called from asyncio (and a couple of special examples), 
however I am afraid of the side effects, especially since you gain absolutely nothing by just declaring the execute async, and the very least you would need to call an async method in the datastore.

janiversen:
I will take a closer look, once you have a green CI.

ilkka-ollakka:
> you might be right that the execute in the different messages is only called from asyncio (and a couple of special examples), however I am afraid of the side effects, especially since you gain absolutely nothing by just declaring the execute async, and the very least you would need to call an async method in the datastore.

Yes, My initial idea was to do that change in separate PR, that I haven't yet opened. Just to make this one only async/await changes and in that point of view, simpler to review. Referring to https://github.com/pymodbus-dev/pymodbus/pull/2127#issuecomment-2028737549

So plan is to open separate PR that will add async_get/setValues to datastore and changes execute to use those with await, then fourth to change remote-datastore to use client as async flow.

ilkka-ollakka:
> you might be right that the execute in the different messages is only called from asyncio (and a couple of special examples), however I am afraid of the side effects, especially since you gain absolutely nothing by just declaring the execute async, and the very least you would need to call an async method in the datastore.

I'm also fine to leave this PR open until I have opened the remaining PR's, so you can check those aswell to see end goal better?

ilkka-ollakka:
> And btw it seems you have not analyzed the code. The execute is both sync and async (as in it returns a future). Execute splits at a lower level to async_execute when called in an async environment

Isn't that only in ModbusClient.execute codepath and not in server-codepath used ModbusRequest.execute ?

janiversen:
Sorry you are right.

janiversen:
I might not had said it correctly, but you need to attach this bottom-up, meaning e.g. this PR does not make sense until you have async datastore in place.

I review each PR as a closed part (which of course can form part of bigger picture), and each PR must be understandable and have a reason (it is not a reason, to state this is needed for a future change somewhere else). This PR adds "async" to a method that have no awaits, so it does not make sense.

I hope I have explained myself better....I am all in favor of having an async datastore, I just need to ensure that the code and the architecture do not break.

ilkka-ollakka:
> I might not had said it correctly, but you need to attach this bottom-up, meaning e.g. this PR does not make sense until you have async datastore in place.

That makes sense

> I review each PR as a closed part (which of course can form part of bigger picture), and each PR must be understandable and have a reason (it is not a reason, to state this is needed for a future change somewhere else). This PR adds "async" to a method that have no awaits, so it does not make sense.

Understandable

> I hope I have explained myself better....I am all in favor of having an async datastore, I just need to ensure that the code and the architecture do not break.

Yes, I'll try to get the remaining PR's open so things can be justified and revied easier.

janiversen:
btw just for information, you might have noticed that asyncio.py calls self.framer.processIncomingPacket which is not async (since it is shared with the sync client) and thus giving you problems. We are currently working a lot to refactor framer, and once complete it will be totally async.

In the meantime, you can replace the "lambda" with functols.partial calling the async executor, with async_execute as argument. This is not a needed change, but just to show you how to get around having 2 execute methods.

test/sub_client/test_client.py
```diff
@@ -392,7 +392,7 @@ def __init__(self, base, req, retries=0):
     async def delayed_resp(self):
         """Send a response to a received packet."""
         await asyncio.sleep(0.05)
-        resp = self.req.execute(self.ctx)
+        resp = await self.req.execute(self.ctx)
```
janiversen, line 395: This seems wrong ! the client should be affected by a server change.

ilkka-ollakka: that is RequestExecute, so for example ReadCoilsRequest, and as that mocks server-responses, it needs to do that async execute call.

ilkka-ollakka:
> You change all server execute to be async, but execute in asyncio is still sync ?

Yes, as execute is passed to callback for processIncomingPacket, and I wanted to scope down the changes as much as possible, I could have used partial there and remove the lambda, but I saw it as not required change at the moment in this scope.

test/sub_client/test_client.py
janiversen: You are correct, thanks for pointing it out.

janiversen:
Closing this since it changed to several small PR

ilkka-ollakka:
> Closing this since it changed to several small PR

Hi, I think you are mixing https://github.com/pymodbus-dev/pymodbus/pull/2127 and this PR, as this PR was split out from the bigger PR, which is 2127 and if you look the change of this PR, it is one of the small changes

janiversen:
Sounds correct, sorry for the mistake.

janiversen:
It's still draft though.

ilkka-ollakka:
> It's still draft though.

I rebased and re-opened the PR

ilkka-ollakka:
windows CI seems to fail on not having python

janiversen (changes requested):
The changes as such are OK, however I miss an await in asyncio.py where execute is called.... it seems to me that this PR guarantees that execute delivers a future.

ilkka-ollakka:
> The changes as such are OK, however I miss an await in asyncio.py where execute is called.... it seems to me that this PR guarantees that execute delivers a future.

That is already in, and was merged in https://github.com/pymodbus-dev/pymodbus/pull/2139

janiversen:
As far as I can see it is not changed.

_async_execute() still contains the check iscoroutine(), which is no longer needed.

ilkka-ollakka:
> As far as I can see it is not changed.
> 
> _async_execute() still contains the check iscoroutine(), which is no longer needed.

I'll add commit to remove it

janiversen (approved):
LGTM, thanks.

PR #2143, opened 2024-04-02
datastore: add async_setValues/getValues methods

ilkka-ollakka:
By default those just call self.set/getValues but can be used to implement more async datastore access in modbus server.

Split from https://github.com/pymodbus-dev/pymodbus/pull/2127

### DIFF
--- a/pymodbus/datastore/context.py
+++ b/pymodbus/datastore/context.py
@@ -1,9 +1,12 @@
 """Context for datastore."""
 
+from __future__ import annotations
+
 # pylint: disable=missing-type-doc
 from pymodbus.datastore.store import ModbusSequentialDataBlock
 from pymodbus.exceptions import NoSuchSlaveException
 from pymodbus.logging import Log
+from pymodbus.pdu import ExceptionResponse
 
 
 class ModbusBaseSlaveContext:
@@ -28,7 +31,7 @@ class ModbusBaseSlaveContext:
         """
         return self._fx_mapper[fx]
 
-    async def async_getValues(self, fc_as_hex, address, count=1):
+    async def async_getValues(self, fc_as_hex: int, address: int, count: int = 1) -> list[int | bool | None] | ExceptionResponse:
         """Get `count` values from datastore.
 
         :param fc_as_hex: The function we are working with
@@ -38,7 +41,7 @@ class ModbusBaseSlaveContext:
         """
         return self.getValues(fc_as_hex, address, count)
 
-    async def async_setValues(self, fc_as_hex, address, values):
+    async def async_setValues(self, fc_as_hex: int, address: int, values: list[int | bool]) -> None:
         """Set the datastore with the supplied values.
 
         :param fc_as_hex: The function we are working with
@@ -47,7 +50,7 @@ class ModbusBaseSlaveContext:
         """
         self.setValues(fc_as_hex, address, values)
 
-    def getValues( self, fc_as_hex, address, count=1):  # pylint: disable=unused-argument
+    def getValues(self, fc_as_hex: int, address: int, count: int = 1) -> list[int | bool | None] | ExceptionResponse:  # pylint: disable=unused-argument
         """Get `count` values from datastore.
 
         :param fc_as_hex: The function we are working with
@@ -57,7 +60,7 @@ class ModbusBaseSlaveContext:
         """
         return []
 
-    def setValues(self, fc_as_hex, address, values):
+    def setValues(self, fc_as_hex: int, address: int, values: list[int | bool]) -> None | ExceptionResponse:
         """Set the datastore with the supplied values.
 
         :param fc_as_hex: The function we are working with
--- a/pymodbus/datastore/context.py
+++ b/pymodbus/datastore/context.py
@@ -1,18 +1,19 @@
 """Context for datastore."""
+
 # pylint: disable=missing-type-doc
 from pymodbus.datastore.store import ModbusSequentialDataBlock
 from pymodbus.exceptions import NoSuchSlaveException
 from pymodbus.logging import Log
 
 
-class ModbusBaseSlaveContext:  # pylint: disable=too-few-public-methods
+class ModbusBaseSlaveContext:
     """Interface for a modbus slave data context.
 
     Derived classes must implemented the following methods:
             reset(self)
             validate(self, fx, address, count=1)
-            getValues(self, fx, address, count=1)
-            setValues(self, fx, address, values)
+            getValues(self, fc_as_hex, address, count=1) or async_getValues(self, fc_as_hex, address, count=1)
+            setValues(self, fc_as_hex, address, values) or async_setValues(self, fc_as_hex, address, values)
     """
 
     _fx_mapper = {2: "d", 4: "i"}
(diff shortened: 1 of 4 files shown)


pymodbus/datastore/context.py
```diff
@@ -27,6 +28,43 @@ def decode(self, fx):
         """
         return self._fx_mapper[fx]
 
+    async def async_getValues(self, fc_as_hex, address, count=1):
```
janiversen, line 31: new methods as well as changed methods must be typed, as part of our ongoing effort to having the codebase strongly typed.

janiversen (changes requested):
LGTM, but please add typing on new/changed methods.

pymodbus/datastore/context.py
ilkka-ollakka: updated typehints to be in place

ilkka-ollakka:
macos CI seems to fail from not able to create working venv from cache.

janiversen:
I had the same problem earlier today, but it went away....have you rebased to newest d3v.

ilkka-ollakka:
> I had the same problem earlier today, but it went away....have you rebased to newest d3v.

Yes, it should be up to date, and the PR itself doesn't show it being behind of dev branch

pymodbus/datastore/context.py
```diff
@@ -27,6 +31,43 @@ def decode(self, fx):
         """
         return self._fx_mapper[fx]
 
+    async def async_getValues(self, fc_as_hex: int, address: int, count: int = 1) -> list[int | bool] | ExceptionResponse:
```
janiversen, line 34: ExceptionResponse is not a valid return from getValues, however it can return None in the list.

pymodbus/datastore/context.py
```diff
@@ -27,6 +31,43 @@ def decode(self, fx):
...
+        """Get `count` values from datastore.
+
+        :param fc_as_hex: The function we are working with
+        :param address: The starting address
+        :param count: The number of values to retrieve
+        :returns: The requested values from a:a+c
+        """
+        return self.getValues(fc_as_hex, address, count)
+
+    async def async_setValues(self, fc_as_hex: int, address: int, values: list[int]) -> None | ExceptionResponse:
```
janiversen, line 44: setValues return None

Remember "raise" is not a return.

janiversen (changes requested):

pymodbus/datastore/context.py
janiversen: An ExceptionResponse might be generated at PDU level, the datastore is only concerned about values.

ilkka-ollakka: It might be the idea, but setValues will return response https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/datastore/remote.py#L60 if it is error one, I don't have opinion if it should raise it instead, but that is current implementation.

Ofcourse the baseclass for those responses might not be ExceptionResponse, that can be fixed.

pymodbus/datastore/context.py
ilkka-ollakka: Same for here, it will return Response https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/datastore/remote.py#L129 if it is error one, even message execute assumes so, https://github.com/pymodbus-dev/pymodbus/blob/dev/pymodbus/bit_write_message.py#L97 . I'll add the None to the list

pymodbus/datastore/context.py
janiversen: Again, remote is off and quite buggy. setValues should not return anything, but might raise and exception.

If you start testing remote, you will find a lot of problems....most of them based in the fact that the current datastore is not async.

janiversen:
Actually if you look at https://github.com/pymodbus-dev/pymodbus/blob/c4c14cab294e166b1efb1b7733b14fb3dcb019a3/pymodbus/bit_write_message.py#L90

you see the right way to use the datastore:
- validate() first
- then call get/setValues

However that collided with remote, who pr definition cannot implement a validate(), so a number of work arounds was invented....but they are going to be removed when refactoring and most likely validate() will also be removed, allowing get/setValues to do a raise instead.

ilkka-ollakka:
> Actually if you look at
> 
> https://github.com/pymodbus-dev/pymodbus/blob/c4c14cab294e166b1efb1b7733b14fb3dcb019a3/pymodbus/bit_write_message.py#L90
> 
> you see the right way to use the datastore:
> 
>     * validate() first
> 
>     * then call get/setValues
> 
> 
> However that collided with remote, which pr definition cannot implement a validate(), so a number of work arounds was invented....but they are going to be removed when refactoring and most likely validate() will also be removed, allowing get/setValues to do a raise instead.

That codeblock same assumes that it can get ExceptionResponse back (https://github.com/pymodbus-dev/pymodbus/blob/c4c14cab294e166b1efb1b7733b14fb3dcb019a3/pymodbus/bit_write_message.py#L94)

If things are gonna be removed when refactoring, then typehints should be changed accordingly, but I think currently typehints should indicate current flow?

janiversen:
Type hint should not reflect wrong code.

janiversen:
Also you have not changed bit_write_message, and I do not think that mypy complains about that.

Since remote inherits from ModsbuBaseSlaveContent, you do not need to make changes in that file. However if you choose to make changes, you should remove the problem.

janiversen:
Instead of having this, in reality stupid, conversation about adding faulty typing, I went ahead and changed the code.

The effect for this PR is that there are no need to change simulator.py

pymodbus/datastore/context.py
ilkka-ollakka: great, I updated the typehint for setValues

pymodbus/datastore/context.py
```diff
@@ -27,6 +31,43 @@ def decode(self, fx):
         """
         return self._fx_mapper[fx]
 
+    async def async_getValues(self, fc_as_hex: int, address: int, count: int = 1) -> list[int | bool | None] | ExceptionResponse:
```
janiversen, line 34: Exceptionresponse is wrong.

pymodbus/datastore/context.py
```diff
@@ -27,6 +31,43 @@ def decode(self, fx):
...
+    async def async_setValues(self, fc_as_hex: int, address: int, values: list[int|bool]) -> None:
+        """Set the datastore with the supplied values.
+
+        :param fc_as_hex: The function we are working with
+        :param address: The starting address
+        :param values: The new values to be set
+        """
+        return self.setValues(fc_as_hex, address, values)
+
+    def getValues(self, fc_as_hex: int, address: int, count: int = 1) -> list[int | bool | None] | ExceptionResponse:  # pylint: disable=unused-argument
```
janiversen, line 53: Exception response is wrong.

Why do you need to add getValues/setValues here, it was not needed earlier?

test/test_remote_datastore.py
```diff
@@ -21,45 +22,62 @@ def test_remote_slave_context(self):
         with pytest.raises(NotImplementedException):
             context.reset()
 
-    def test_remote_slave_set_values(self):
+    async def test_remote_slave_set_values(self):
```
janiversen, line 25: This should not be needed, since remote is not changed?

test/test_remote_datastore.py
```diff
@@ -21,45 +22,62 @@ def test_remote_slave_context(self):
...
+        client.write_registers = mock.MagicMock(
+            return_value=ExceptionResponse(0x10, 0x02)
+        )
 
         context = RemoteSlaveContext(client)
+        await context.async_setValues(0x0F, 0, [1])
         context.setValues(0x0F, 0, [1])
         # result = context.setValues(0x10, 1, [1])
-        context.setValues(0x10, 1, [1])
+        await context.async_setValues(0x10, 1, [1])
```
janiversen, line 37: Please o not change to async, the old methods must still work.

test/test_remote_datastore.py
```diff
@@ -21,45 +22,62 @@ def test_remote_slave_context(self):
...
         # assert result.exception_code == 0x02
         # assert result.function_code == 0x90
 
-    def test_remote_slave_get_values(self):
+    async def test_remote_slave_get_values(self):
```
janiversen, line 41: Same here.

test/test_sparse_datastore.py
```diff
@@ -1,10 +1,12 @@
 """Test framers."""
 
+import pytest
 
 from pymodbus.datastore import ModbusSparseDataBlock
 
 
-def test_check_sparsedatastore():
+@pytest.mark.asyncio()
```
janiversen, line 8: Please leve the old methods.

Instead please add an extra test with async (e.g. duplicate the existing). Since this PR adds async but do not remove sync, it is important to test that both variants work.

janiversen (changes requested):
It looks as if you are very close, seems you simply forgot some edits.

pymodbus/datastore/context.py
ilkka-ollakka: bit_read_messages.py and register_read_messages.py execute-calls seems to expect to receive ExceptionResponse from getValues.

And it was added here, as async_get/setValues refer to it in here.

test/test_sparse_datastore.py
ilkka-ollakka: sure, I can duplicate the test function and make it separate

pymodbus/datastore/context.py
janiversen: Well another problem from remote, ExceptionResponse is NOT valid in get/set values.

janiversen:
How are you doing with this one, seems you still have confusion about ExceptionResponse.

janiversen:
It set at "draft" so I assume you are still working.

janiversen:
hmmm seems I lost a message edit.

I went ahead and polished the PR, so it follows the rest of the code.

Superseded by #2165 (merged, with @ilkka-ollakka as committer)

PR #2161, opened 2024-04-13
Fixed kwargs not being expanded for actions on bit registers

sumguytho:


janiversen (approved):
LGTM, thanks.

Good catch !!!

janiversen:
A nice catch, seems you are using the simulator, interested in participating in defining how to configure the datastore simulator for the future ?

Background, we are going to replace all datastores with the simulator datastore. The other datastores will remain but will generate a simulator datastore. The challenge is to make a datastore simulator configuration easy to understand and use for application users.

sumguytho:
Hey, thanks for the offer but I'd decline. Glad I could help the project, though.

Issue #2173, opened 2024-04-22
TCP client frame mismatch under heavy load


martyy665:
### Versions

- Python: 3.11
- OS: Linux
- Pymodbus:  3.6.8

### Pymodbus Specific

- Client: tcp - async

### Description

After some periodic readouts from multiple slaves, wrong data are returned (data coming from another slave). It appears when the `tid` counter overflows 65535 and goes back to zero. Changing the line 507 in `pymodbus/transaction.py` below fixes the issue.

### Code and Logs

```python

     def getTransaction(self, tid):
499         """Return a transaction matching the referenced tid.
500
501         If the transaction does not exist, None is returned
502
503         :param tid: The transaction to retrieve
504
505         """
506         Log.debug("Getting transaction {}", tid)
507         if tid is None:
508             if self.transactions:
509                 return self.transactions.popitem()[1]
510             return None
511         return self.transactions.pop(tid, None)

```

janiversen:
Please submit a pull request.

It is an interesting observation, we do not have a test case for this.

martyy665:
Closed by approved PR.

janiversen:
It seems I slept when approving this, because it breaks a lot of tests. Problem is that tid == 0 is an allowed transaction,

Will submit a fix shortly,

PR #2174, opened 2024-04-23
Fix condition to eliminate TID overflow issue

martyy665:
Fix tid overflow issue when tid is zero.

These two snippets are NOT the same - the difference appears when the numeric variable reaches zero (i.e. overflows):

```
if not tid :
    ...
```

```
if tid is None :
    ....
```
Reference:
#2173


janiversen (approved):
You are correct they are not the same. yours test for none and zero.

janiversen:
Thanks.

PR #2177, opened 2024-04-23
Don't initialize the event loop until needed.

jameshilliard:
Initializing the event loop during class init can cause issues in some cases where the event loop is not yet running.

### DIFF
--- a/pymodbus/server/async_io.py
+++ b/pymodbus/server/async_io.py
@@ -49,7 +49,6 @@ class ModbusServerRequestHandler(ModbusProtocol):
         self.receive_queue: asyncio.Queue = asyncio.Queue()
         self.handler_task = None  # coroutine to be run on asyncio loop
         self.framer: ModbusFramer
-        self.loop = asyncio.get_running_loop()
 
     def _log_exception(self):
         """Show log exception."""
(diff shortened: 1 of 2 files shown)


janiversen:
How can the event loop not be running in an async application?

I know of only one situation where this happens, and that is if you create the client/servers in a sync program and then start the event loop, but I have yet to understand the benefits of that.

The reason the event loop check was moved into init, was because setup is done in init, which includes calling asyncio functions, not having a loop caused some of the code to not work properly (as in hsvinhg multiple loops).

I will review the code, but look forward to understand why you find this needed.

jameshilliard:
> How can the event loop not be running in an async application?

Seemed to be due to initializing at module import or something along those lines. When updating an application of ours from `3.6.4` to `3.6.8` we were hitting `no running event loop` errors.

> The reason the event loop check was moved into init, was because setup is done in init, which includes calling asyncio functions, not having a loop caused some of the code to not work properly (as in hsvinhg multiple loops).

Ah, ok, so maybe this isn't correct then. We refactored things and got it working without this change so might not be needed.

janiversen:
It is not one of our imports, it happens when you create a client/server object.

PR #2191, opened 2024-05-08
Fix writing to serial (rs485) on windows os.

andrew-harness:
The issue with the write method on Windows is that it's trying to use self.sync_serial.fileno(), which doesn't work on Windows because the pySerial library doesn't provide a valid file descriptor.

To fix this, you can modify the write method to directly write to the serial port without using asyncio's add_writer method.

### DIFF
--- a/pymodbus/transport/serialtransport.py
+++ b/pymodbus/transport/serialtransport.py
@@ -55,8 +55,17 @@ class SerialTransport(asyncio.Transport):
     def write(self, data) -> None:
         """Write some data to the transport."""
         self.intern_write_buffer.append(data)
-        if not self.poll_task:
+        if not self.poll_task and os.name != "nt":
             self.async_loop.add_writer(self.sync_serial.fileno(), self.intern_write_ready)
+        elif not self.poll_task and os.name == "nt":
+            self.poll_task = self.async_loop.create_task(self.write_task())
+
+    async def write_task(self):
+        """Write data to the serial port."""
+        while self.intern_write_buffer:
+            data = self.intern_write_buffer.pop(0)
+            self.sync_serial.write(data)
+            await asyncio.sleep(0)  # Yield control to the event loop
 
     def flush(self) -> None:
         """Clear output buffer and stops any more data being written."""
--- a/pymodbus/transport/serialtransport.py
+++ b/pymodbus/transport/serialtransport.py
@@ -13,7 +13,7 @@ with contextlib.suppress(ImportError):
 class SerialTransport(asyncio.Transport):
     """An asyncio serial transport."""
 
-    force_poll: bool = False
+    force_poll: bool = os.name == "nt"
 
     def __init__(self, loop, protocol, *args, **kwargs) -> None:
         """Initialize."""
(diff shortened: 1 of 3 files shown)


pymodbus/transport/serialtransport.py
```diff
@@ -55,8 +55,17 @@ def close(self, exc: Exception | None = None) -> None:
     def write(self, data) -> None:
         """Write some data to the transport."""
         self.intern_write_buffer.append(data)
-        if not self.poll_task:
+        if not self.poll_task and os.name != "nt":
```
janiversen, line 58: The if/elif is not very logical it seems you could simplify it a lot.

Secondly I am still not convinced this is the right way, at least please explain why CI that also runs on windows works.

janiversen (changes requested):
Tests are missing ! the transport part have a 100% coverage and we would like to keep that.

pymodbus/transport/serialtransport.py
janiversen: The high level tests uses null modem, which basically bypasses pyserial, but there are low level tests that uses pyserial (transport itself), however they do NOT use the serial interface but the socket interface, and that might be the difference.

Apologies if I come across as a bit negative, but the problem is that I have to maintain the code long term so while I look positive at (nearly) any PR, I give a feedback that ensures it is not making maintenance harder.

Precise comments:
- if / elif, as you admitted it can be simplified, this is important because in a year from now I (or someone else) will start thinking what is the reason that it is complex, what have I overlooked...

- Cross platform: this PR should not be closed, it solves a problem, we just need to get it right. A simplification would be to only use the create_task you suggest, and remove the add_writer totally....I am all in a favor of that.

So to sum up:
- please simplify the if / elif
- I suggest to remove add_writer in total and only use your create task.
- Please add tests to ensure the coverage is 100% (pytest --cov or just check the CI run)

I look forward to review a new version from you...I hope you can see that I am positive and your PR have value !

andrew-harness: I looked into this a bit more as I saw other instances of the fileno() being called and no errors being thrown. It looks like what was happening is the write function is called before the setup function is ran.

In setup it is specifically setting the poll_task on windows.

```
    def setup(self) -> None:
        """Prepare to read/write."""
        if os.name == "nt" or self.force_poll:
            self.poll_task = asyncio.create_task(self.polling_task())
            self.poll_task.set_name("SerialTransport poll")
        else:
            self.async_loop.add_reader(self.sync_serial.fileno(), self.intern_read_ready)
        self.async_loop.call_soon(self.intern_protocol.connection_made, self)
```

I put a break point in the write function at the add_writer line and noticed that it was being hit. Which according to the setup function it shouldn't be since it is Windows. I then added a break point to the setup function and noticed that the write function is called once before the setup function is ran.

Adding a timeout after the connect and before a read_input_registers (or other modbus operation) allowed the setup function to finish before the write operation. I then did not receive any exceptions.

Looking at the create_serial_connection you can see that the transport.setup is scheduled to run once the event loop becomes idle. This means that read and write operations can happen before the setup function is run.

Calling the setup function to run immediately mitigates this issue.

```
async def create_serial_connection(
    loop, protocol_factory, *args, **kwargs
) -> tuple[asyncio.Transport, asyncio.BaseProtocol]:
    """Create a connection to a new serial port instance."""
    protocol = protocol_factory()
    transport = SerialTransport(loop, protocol, *args, **kwargs)
    transport.setup()
    # loop.call_soon(transport.setup)
    return transport, protocol
```

I think my original changes should be rejected and this change be implemented instead or the connected function 
 (that is typically awaited after client creation) wait for the connection_made handler to finish, which is called from the setup function.

Thoughts?

janiversen: Sounds correct, please update the PR

Issue #2204, opened 2024-05-27
Use pyserial RS485 - not serial_from_url


samskiter:
### Versions

- Python: 3.11.2
- OS: Linux  6.6.20+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.20-1+rpt1 (2024-03-07) aarch64 GNU/Linux
- Pymodbus: 3.6.8
- Modbus Hardware (if used): MAX481 + RPI4B UART0 + RTS0

### Pymodbus Specific

- Server: rtu - sync AND async

### Description

I'm trying to use pymodbus with the UART on the RPi 4B with a half duplex RS485 chip that needs nRE/DE input to correctly drive/receive over half duplex comms. I have been trying to set RS485 mode into the serial port opened and [found this existing issue](https://github.com/raspberrypi/linux/issues/6189) where someone else had the same problem

This is confounded by [a bug ](https://github.com/raspberrypi/linux/issues/6189) on the RPi/with pyserial that causes the UART to be pernickity about how it is configured.

I'd like to change the RS485 mode on the serial port created by pymodbus, but instead it seems I need to swap the whole serial port out.

It seems that it would be sensible for pymodbus to open an RS485 serial port, rather than a plain RS232 UART.

BONUS would be to allow users to configure the RS485 port directly through PyModbus.

### Code and Logs

I have tried to configure RS485 settings on the serial port after connection like this:

```python
    def blah() -> None:
        self.async_client = ModbusClient.AsyncModbusSerialClient(
            port=self.protocol.port,
            baudrate=self.protocol.baud_rate,
            byte_size=self.protocol.byte_size,
            parity=self.protocol.parity,
            stopbits=self.protocol.stop_bits,
        )
        await self.connect()
        return self.protocol

    async def connect(self) -> None:
        """
        Connects to the Modbus device.
        """
        await self.async_client.connect()

        transport: SerialTransport = self.async_client.transport
        serial_object: Serial = transport.sync_serial
        serial_object.rtscts = True
        serial_object.rs485_mode = serial.rs485.RS485Settings(
            rts_level_for_tx=False,
            rts_level_for_rx=True,
            loopback=False,
            delay_before_tx=0.0000035,
            delay_before_rx=0.0000001,
        )
        self.connected = True
```
But I hit the bug mentioned above (https://github.com/raspberrypi/linux/issues/6189)

[...]

PR #2205, opened 2024-05-27
Update use pyserial RS485

samskiter:
Fixes: https://github.com/pymodbus-dev/pymodbus/issues/2204


samskiter:
Can't get this passing - it seems that serial_for_url is used by this library, but there's no way to use that to produce instances of pyserial's RS485 class. Perhaps the solution is to implement some kind of custom URL scheme for this purpose?

janiversen (changes requested):
Seems your fix breaks quite a number of tests, please fix.

samskiter:
Open to suggestions - I can't seem to fix it!

janiversen:
I have no suggestions, the serial library is only used in serial transport and there are no other libraries being used.

serial_for_url is essential for pymodbus since it allows simulating serial communication on a socket, but it's strange if the rs485 do not supply it.

samskiter:
OK, I think this is actually a bit of an awkward solution, but ultimately the correct thing to do.

The purpose of using the RS485 class is to be able to configure RS485Settings, and without injecting those, I'll have to reach inside pymodbus and grab things I shouldn't to set rs485_mode on the sync_serial client. So I opted to allow these settings to be injected along with us in the commparams - there is precedent for params in there to be unused for some transports (e.g. the `port` parameter is not used if you have a `serial` commtype)

I hope this is an acceptable fix

test/transport/test_serial.py
```diff
@@ -55,10 +55,10 @@ async def test_external_methods(self, inx):
         comm.sync_serial.read = mock.MagicMock(return_value="abcd")
         comm.sync_serial.write = mock.MagicMock(return_value=4)
         comm.sync_serial.fileno = mock.MagicMock(return_value=2)
-        comm.sync_serial.async_loop.add_writer = mock.MagicMock()
-        comm.sync_serial.async_loop.add_reader = mock.MagicMock()
-        comm.sync_serial.async_loop.remove_writer = mock.MagicMock()
-        comm.sync_serial.async_loop.remove_reader = mock.MagicMock()
+        comm.async_loop.add_writer = mock.MagicMock()
```
samskiter, line 58: Note that I think this code was incorrect anyway, but my changes made this issue rear its head

pymodbus/client/base.py
```diff
@@ -80,6 +80,7 @@ def __init__(
                 parity=kwargs.get("parity", None),
                 stopbits=kwargs.get("stopbits", None),
                 handle_local_echo=kwargs.get("handle_local_echo", False),
+                rs485_settings=kwargs.get("rs485_settings", None),
```
janiversen, line 83: Parameter doc missing, please see serial.py line 35++

Since this is a serial parameter only, it should be moved to serial.py

right now you are adding rs485_settings to all types.

In general it should be a line in comm_params.

pymodbus/transport/serialtransport.py
```diff
@@ -8,25 +8,36 @@
...
+        self.sync_serial = self._serial_for_args(rs485_settings, *args, **kwargs)
         self.intern_write_buffer: list[bytes] = []
         self.poll_task: asyncio.Task | None = None
         self._poll_wait_time = 0.0005
         self.sync_serial.timeout = 0
         self.sync_serial.write_timeout = 0
 
+    def _serial_for_args(self, rs485_settings, *args, **kwargs) -> serial.Serial:
+        sync_serial: serial.Serial
+        if rs485_settings is not None:
```
janiversen, line 33: remove "is not None", that is the same but shorter.

pymodbus/transport/transport.py
```diff
@@ -98,6 +100,9 @@ class CommParams:
     parity: str = ''
     stopbits: int = -1
 
+    # RS485
+    rs485_settings: RS485Settings | None = None
```
janiversen, line 104: Please combine with line 196.

pymodbus/transport/transport.py
```diff
@@ -188,14 +192,17 @@ def __init__(
             self.comm_params.comm_type = CommType.TCP
             parts = host.split(":")
             host, port = parts[1][2:], int(parts[2])
-        self.init_setup_connect_listen(host, port)
 
-    def init_setup_connect_listen(self, host: str, port: int) -> None:
+        rs485_settings = self.comm_params.rs485_settings
```
janiversen, line 196: something if not correct here, in client/base.py you create a local variable, and do not use comm_params.

pymodbus/transport/transport.py
```diff
@@ -188,14 +192,17 @@ def __init__(
...
+        self.init_setup_connect_listen(host, port, rs485_settings)
```
janiversen, line 197: No need for a local variable, use comm_params directly.

test/transport/test_serial.py
```diff
@@ -15,7 +15,7 @@
 
 
 @mock.patch(
-    "pymodbus.transport.serialtransport.serial.serial_for_url", mock.MagicMock()
+    "pymodbus.transport.serialtransport.SerialTransport._serial_for_args", mock.MagicMock()
```
janiversen, line 18: you are mocking our own, meaning you miss to test the mocked function.

test/transport/test_serial.py
janiversen, line 58: you need to update your dev, you are several commits behind.

test/transport/test_serial.py
```diff
@@ -96,7 +96,7 @@ async def test_write_force_poll(self):
         """Test write with poll."""
         SerialTransport.force_poll = True
         transport, protocol = await create_serial_connection(
-            asyncio.get_running_loop(), mock.Mock, url="dummy"
+            asyncio.get_running_loop(), mock.Mock, None, url="dummy"
```
janiversen, line 99: So there are NO test at all of the rs485 settings.

Transport have 100% coverage, please ensure you still keep that.

janiversen (changes requested):
Minor details, the only bigger thing is your use (or not use) of comm_params.

We are in the process of preparing v3.7.0 so if you hurry this PR can be part of 3.7.0

pymodbus/transport/serialtransport.py
samskiter: Can do - but while in this case it is the same - it's a pattern I like to avoid as sometimes things can evaluate to falsey when you don't expect them so.

Simple example:


```

maybe_bool: Optional[bool] = get_result()

if maybe_bool:
    print("no bool returned!")
else:
    ...
```

test/transport/test_serial.py
samskiter: Sure thing - when you say coverage are you talking about LOC coverage?

janiversen: I am talking about coverage. Please see the CI run.

janiversen:
It is easiest to run locally "pytest --cov"

test/transport/test_serial.py
samskiter: what do you mean - the diff wouldn't show if there wasn't a diff against dev

pymodbus/transport/transport.py
samskiter: No local variable created in client/base.py - I don't understand

pymodbus/client/base.py
samskiter: done

pymodbus/client/serial.py
```diff
@@ -15,6 +15,7 @@
 
 try:
     import serial
+    from serial.rs485 import RS485Settings
```
janiversen, line 18: It is already imported in the line above, do not import twice.

pymodbus/transport/serialtransport.py
```diff
@@ -8,25 +8,36 @@
 
 with contextlib.suppress(ImportError):
     import serial
+    import serial.rs485
```
janiversen, line 11: It is already imported in the line above.

pymodbus/transport/transport.py
```diff
@@ -58,6 +58,8 @@
 from functools import partial
 from typing import Any
 
+from serial.rs485 import RS485Settings
```
janiversen, line 61: What happens if serial is not installed ?? Please import as serial is imported in other places.

test/transport/test_serial.py
```diff
@@ -7,6 +7,7 @@
 
 import pytest
 import serial
+from serial.rs485 import RS485Settings
```
janiversen, line 10: Already imported in the line above, please do not import twice.

janiversen (changes requested):

janiversen:
We have for other reasons decided to make a v3.6.9 that will be done this weekend, so I need an updated PR from you if you want it included.

janiversen:
Still working on this ?

yawor:
I write a new software for already existing hardware with RS485 transceiver, but it is based on an AllWinner H3 SBC, where hardware flow control doesn't work very well, but I need to control RTS pin to control the transceiver's direction. So I really would like to have this feature merged, but (in all due respect to @samskiter) I don't think this is implemented properly.

SerialTransport adapts a non-asyncio `serial.Serial` class by adding async readers and writers to monitor serial port's file descriptor and they invoke callbacks to either read or write the data in a non-blocking fashion using normally blocking read and write methods. But `RS485.write` method adds additional, blocking delays (for example, it always flushed the port after each write), which should not be executed in the async writer's callbacks, because it's going to block the loop.

Also there's no guarantee that the async writer is going to call the callback only once per `SerialTransport.write` call as it's possible that the writer would split larger data into smaller chunks. In that case there would be multiple calls to `RS485.write` method for a single `SerialTransport.write` call, causing RTS to be toggled (potentially multiple times) in the middle of a single modbus operation, causing it to fail.

That's why the optional software RTS control should be implemented directly in the `SerialTransport` class, which is actually aware when the whole frame has been sent and RTS can be switched from TX to RX. It would be even better if it could just call program provided callbacks (either sync or async) before transmission starts and after it ends. That way a program could even use any available generic GPIO pin to control the transceiver direction, not just RTS pin.

janiversen:
In general I agree with your arguments, we have over time had many problems with serial, a lot because it is no longer actively maintained.

We have been thinking of changing serial to another library, but never got time to actually do it.

I would however like to see a good RS485 integration! and e.g. enhac3 the transport layer to control RTS would also be welcomed.

So in short, and since this PR looks abandoned, I hope you will submit your ideas as a PR.

janiversen:
I think RTS control can be implemented in the current serial transport, with an extra parameter to turn it on.

The best way would be to make a PR that implement it in serial transport with test, and a second that add the parameter to comm_params and the client/server api.

samskiter:
Sorry for my delays, I will try and get this pr over the line!

janiversen:
@samskiter good luck, please resolve the conflicts first and then have a close look at the comments from @yawor which I think have a lot of merit.

Ohh and do not forget the review comments I made.

jameshilliard:
> But `RS485.write` method adds additional, blocking delays (for example, it always flushes the port after each write), which should not be executed in the async writer's callbacks, because it's going to block the loop.

So it appears pyserial also has some native rs485 support that AFAIU would avoid these issues, reworked this a bit to use that instead in #2460.

PR #2223, opened 2024-06-24
Add should_respond attribute to ModbusRequest

jameshilliard:
Allows for custom ModbusRequest classes where responses may not be expected.

### DIFF
--- a/examples/client_custom_msg.py
+++ b/examples/client_custom_msg.py
@@ -17,6 +17,7 @@ from pymodbus import FramerType
 from pymodbus.client import AsyncModbusTcpClient as ModbusClient
 from pymodbus.pdu import ModbusExceptions, ModbusRequest, ModbusResponse
 from pymodbus.pdu.bit_read_message import ReadCoilsRequest
+from pymodbus.pdu.register_write_message import WriteSingleRegisterRequest
 
 
 # --------------------------------------------------------------------------- #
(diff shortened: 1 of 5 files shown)


janiversen (changes requested):
This is strictly against the modbus protocol, but it could work and it does not interfere with the normal operation, 

However such a change really needs a number of specific test cases (the major problem seems to be if a custom class is added and then used with a sync client), also the possibility needs to be documented, with at least one example.

Please be aware that we are in the process of releasing v3.6.9 I have not tested if that will cause conflicts in this PR.

jameshilliard:
> However such a change really needs a number of specific test cases (the major problem seems to be if a custom class is added and then used with a sync client), also the possibility needs to be documented, with at least one example.

I added some test cases/examples, not sure if I'm doing that correctly but they seem to pass. By the way when I was testing this didn't seem to work well with say the socket framer(not sure if there's a way to make it compatible with that or if it should just be disabled for non-RTU framers.

janiversen:
Either it should work with all framers, or the library should warn the user so we can avoid issues.

I will take a look at the code tomorrow.

jameshilliard:
> Either it should work with all framers, or the library should warn the user so we can avoid issues.

Yeah, so it's a bit unclear to me how this interacts with real socket framer hardware, we only use the RTU framer ourselves so I don't really have anything to test with.

janiversen:
> > Either it should work with all framers, or the library should warn the user so we can avoid issues.
> 
> Yeah, so it's a bit unclear to me how this interacts with real socket framer hardware, we only use the RTU framer ourselves so I don't really have anything to test with.

That is a very normal case, and the reason we have a test harness. Actually I am not too sure why this PR depend on the type of framer, it tells me something might be wrong.

janiversen:
Still working on this ?

jameshilliard:
Going to close this for now, still may make sense eventually to add this but the sync transaction code is too messy at the moment for me to implement this, might circle back once it's refactored. I managed to get a firmware update from the hardware vendor that fixed the bug that was causing write registers messages to not get responses anyways.

PR #2224, opened 2024-06-26
Simplify some BinaryPayload pack operations

jameshilliard:
Slightly cleaner and easier to read.

### DIFF
--- a/pymodbus/payload.py
+++ b/pymodbus/payload.py
@@ -11,6 +11,8 @@ __all__ = [
     "BinaryPayloadDecoder",
 ]
 
+from array import array
+
 # pylint: disable=missing-type-doc
 from struct import pack, unpack
 
@@ -21,9 +23,6 @@ from pymodbus.utilities import (
     pack_bitstring,
     unpack_bitstring,
 )
-
-
-WC = {"b": 1, "h": 2, "e": 2, "i": 4, "l": 4, "q": 8, "f": 4, "d": 8}
 
 
 class BinaryPayloadBuilder:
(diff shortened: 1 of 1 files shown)


pymodbus/utilities.py
```diff
@@ -182,3 +186,11 @@ def hexlify_packets(packet):
     if not packet:
         return ""
     return " ".join([hex(int(x)) for x in packet])
+
+
+def system_endian():
```
janiversen, line 191: This looks wrong, because byte order can be different from word order.

It seems wrong that we add endian utilities, we should avoid that.

janiversen (changes requested):
To be honest it is quite diffecult to see if the changes have any side effects.

This needed to be tested extra in the test harness, please do "protest --cov" and check that all code is covered.

pymodbus/utilities.py
jameshilliard: So the issue here is that the array class which I'm using (because it provides a nice way to operate on 16 bit integer lists as bytes) encodes integers with the system native byte format. In pymodbus AFAIU our registers list is encoded as a list of big endian integers so when converting from a register list to a python array(which has a convenient bytes method) we have to conditionally byteswap when converting if our native system byteformat is little endian so that the individual python array elements get ordered as big endian bytes. So this system endianness check is only actually being used when encoding/decoding to the list of integers to/from bytes. We don't need to care about word order at this point either since we don't change word order when doing the register list to array to bytes conversions.

Once we are actually encoding/decoding integers i.e. in `_pack_words`/`_unpack_words` we can simply assume we are already big endian format and perform the byteswaping and wordswapping operations without having to care about system native byte format.

jameshilliard:
> This needed to be tested extra in the test harness, please do "protest --cov" and check that all code is covered.

I went ahead and removed the system endian dependent conversion operations, all these changes now should have full coverage.

pymodbus/payload.py
```diff
@@ -69,15 +68,14 @@ def _pack_words(self, fstring: str, value) -> bytes:
...
-            payload = payload[::-1]
-
-        fstring = self._byteorder + "H"
-        return b"".join(pack(fstring, word) for word in payload)
+        if Endian.LITTLE in {self._byteorder, self._wordorder}:
+            value = array("H", value)
+            if self._byteorder == Endian.LITTLE:
+                value.byteswap()
+            if self._wordorder == Endian.LITTLE:
+                value.reverse()
```
janiversen, line 76: this is wrong because it reverses the byte order not the word order

e.g. value = 0123 result should be 2301, NOT 3210

janiversen (approved):
LGTM, thanks.

PR #2230, opened 2024-07-13
Add optional request params to reject truncated responses and error responses.

jameshilliard:
Adds tracking for truncated responses and the ability for the retry handler to reject truncated and error responses based on request flags.

### DIFF
--- a/pymodbus/pdu/pdu.py
+++ b/pymodbus/pdu/pdu.py
@@ -102,13 +102,17 @@ class ModbusRequest(ModbusPDU):
 
     function_code = -1
 
-    def __init__(self, slave=0, **kwargs):
+    def __init__(self, slave=0, error_responses=True, truncated_responses=True, **kwargs):
         """Proxy to the lower level initializer.
 
         :param slave: Modbus slave slave ID
+        :param error_responses: If we should accept an error response.
+        :param truncated_responses: If we should accept a truncated response.
         """
         super().__init__(slave, **kwargs)
         self.fut = None
+        self.error_responses = error_responses
+        self.truncated_responses = truncated_responses
 
     def doException(self, exception):
         """Build an error response based on the function.
@@ -148,6 +152,11 @@ class ModbusResponse(ModbusPDU):
         self.bits = []
         self.registers = []
         self.request = None
+
+    @property
+    def truncated(self) -> bool:
+        """Check if the response has been truncated."""
+        return False
 
     def isError(self) -> bool:
         """Check if the error is a success or failure."""
--- a/pymodbus/pdu/register_read_message.py
+++ b/pymodbus/pdu/register_read_message.py
@@ -109,6 +109,14 @@ class ReadRegistersResponseBase(ModbusResponse):
         """
         return self.registers[index]
 
+    @property
+    def truncated(self) -> bool:
+        """Check if the response has been truncated."""
+        if isinstance(self.request, ReadRegistersRequestBase):
+            if len(self.registers) < self.request.count:
+                return True
+        return False
+
     def __str__(self):
         """Return a string representation of the instance.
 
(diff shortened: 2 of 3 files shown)


janiversen (changes requested):
First of all it seems you forgot to cover the sync client.

The purpose of the clients in the library is to "send requests and return healthy responses with automatic retry" ! This PR starts to add an extra layer that changes the definition to "send requests and return expected responses with automatic retry", it might be better to that in a higher level client (which correspond to the discussion about managing a whole device).

pymodbus/pdu/register_read_message.py
```diff
@@ -109,6 +109,14 @@ def getRegister(self, index):
         """
         return self.registers[index]
 
+    @property
+    def truncated(self) -> bool:
+        """Check if the response has been truncated."""
```
janiversen, line 114: You are missing a lot of requests that also have length

if implemented it must be implemented for all requests !

pymodbus/pdu/pdu.py
```diff
@@ -102,12 +102,15 @@ class ModbusRequest(ModbusPDU):
 
     function_code = -1
 
-    def __init__(self, slave=0, **kwargs):  # pylint: disable=useless-parent-delegation
+    def __init__(self, slave=0, error_responses=True, truncated_responses=True, **kwargs):  # pylint: disable=useless-parent-delegation
```
janiversen, line 105: Error response have nothing to do with truncate, and that is something the app need to handle.

janiversen (commented):
A few review comments

pymodbus/pdu/register_read_message.py
jameshilliard: Yeah, I just started with this one, still need to go through the others and add truncation properties.

pymodbus/pdu/pdu.py
jameshilliard: > Error response have nothing to do with truncate

So I added that variable to allow tweaking the retry handler behavior for error responses similar to truncated responses since depending on desired behavior on may want to retry on error until successful.

> that is something the app need to handle

I had added this since I wasn't seeing a good approach once could use to handle this app side due to the retry handler and counter being part of the client.

jameshilliard:
> First of all it seems you forgot to cover the sync client.

Yeah, the sync client code is rather...tricky IMO, trying to get it working there.

> The purpose of the clients in the library is to "send requests and return healthy responses with automatic retry" !

IMO this is just allowing the definition of a healthy response to be configured per request.

> This PR starts to add an extra layer that changes the definition to "send requests and return expected responses with automatic retry", it might be better to that in a higher level client (which correspond to the discussion about managing a whole device).

I'm not sure the higher level client layer would be where it makes sense to implement retry logic, that layer I'm thinking would be more for providing a register mapping layer AFAIU but the retry logic seems to me that it would need to be handled at a lower level.

janiversen:
having a short read, is normal (I am sitting with a device right now that does it), f.x. if the app read registers non existent. As a consequence it is something the app should handle !

Many of the error responses, are not dynamic in nature, so adding them to the retry does not work, apart from that the app needs to check the error flag (as pr other discussion, it is a major change which we can think of for 4.0).

So I do not see a lot of value in this PR (apart from the review comments).

janiversen:
Still working on this one ?

jameshilliard:
> having a short read, is normal (I am sitting with a device right now that does it)

I mean, I made it a flag tied to the specific request because while truncation may be normal for some devices/registers, it may be not normal for others and indicate some other issue.

> As a consequence it is something the app should handle !

At the moment I'm using a wrapper that implements an additional layer of retry logic for both truncation and error flags(retries do seem to reliably mitigate the issue at least with the devices I'm working with), but having two retry loops(one in the app one in the library) effectively doing the same thing seems a bit weird.

> Many of the error responses, are not dynamic in nature, so adding them to the retry does not work, apart from that the app needs to check the error flag (as pr other discussion, it is a major change which we can think of for 4.0).

Hmm, I'm guessing this is somewhat firmware dependent, I think having a retry on error capability is still seems somewhat useful in case apps want to delegate retries to the library.

> Still working on this one ?

Yes, but somewhat prioritizing the transaction and tests cleanup at the moment.

PR #2238, opened 2024-07-18
Add more typing to transaction and base client

jameshilliard:
Also fix some issues identified by additional checks.

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -4,9 +4,9 @@ from __future__ import annotations
 import asyncio
 import socket
 from abc import abstractmethod
-from collections.abc import Awaitable, Callable
+from collections.abc import Callable, Coroutine
 from dataclasses import dataclass
-from typing import Any, cast
+from typing import Any, Union, cast
 
 from pymodbus.client.mixin import ModbusClientMixin
 from pymodbus.client.modbusclientprotocol import ModbusClientProtocol
@@ -20,7 +20,7 @@ from pymodbus.transport import CommParams, CommType
 from pymodbus.utilities import ModbusTransactionState
 
 
-class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]]):
+class ModbusBaseClient(ModbusClientMixin[Coroutine[Any, Any, Union[ModbusResponse, None]]]):
     """**ModbusBaseClient**.
 
     Fixed parameters:
@@ -144,7 +144,7 @@ class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]]):
             return 0
         return self.last_frame_end + self.silent_interval
 
-    def execute(self, request: ModbusRequest):
+    def execute(self, request: ModbusRequest) -> Coroutine[Any, Any, ModbusResponse | None]:
         """Execute request and get response (call **sync/async**).
 
         :param request: The request to process
@@ -158,7 +158,7 @@ class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]]):
     # ----------------------------------------------------------------------- #
     # Merged client methods
     # ----------------------------------------------------------------------- #
-    async def async_execute(self, request) -> ModbusResponse:
+    async def async_execute(self, request) -> ModbusResponse | None:
         """Execute requests asynchronously."""
         request.transaction_id = self.ctx.transaction.getNextTID()
         packet = self.ctx.framer.buildPacket(request)
@@ -186,9 +186,9 @@ class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]]):
                 f"ERROR: No response received after {self.retries} retries"
             )
 
-        return resp  # type: ignore[return-value]
+        return resp
 
-    def build_response(self, request: ModbusRequest):
+    def build_response(self, request: ModbusRequest) -> asyncio.Future[ModbusResponse]:
         """Return a deferred response for the current request."""
         my_future: asyncio.Future = asyncio.Future()
         request.fut = my_future
@@ -225,7 +225,7 @@ class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusResponse]]):
         )
 
 
-class ModbusBaseSyncClient(ModbusClientMixin[ModbusResponse]):
+class ModbusBaseSyncClient(ModbusClientMixin[Union[ModbusResponse, bytes, ModbusIOException]]):
     """**ModbusBaseClient**.
 
     Fixed parameters:
@@ -279,7 +279,7 @@ class ModbusBaseSyncClient(ModbusClientMixin[ModbusResponse]):
         **kwargs: Any,
     ) -> None:
         """Initialize a client instance."""
-        ModbusClientMixin.__init__(self)  # type: ignore[arg-type]
+        ModbusClientMixin.__init__(self)  # type: ignore[arg-type] # pylint: disable=non-parent-init-called
         self.comm_params = CommParams(
             comm_type=comm_type,
             comm_name="comm",
@@ -346,7 +346,7 @@ class ModbusBaseSyncClient(ModbusClientMixin[ModbusResponse]):
             return 0
         return self.last_frame_end + self.silent_interval
 
-    def execute(self, request: ModbusRequest) -> ModbusResponse:
+    def execute(self, request: ModbusRequest) -> ModbusResponse | bytes | ModbusIOException:
         """Execute request and get response (call **sync/async**).
 
         :param request: The request to process
--- a/pymodbus/transaction.py
+++ b/pymodbus/transaction.py
@@ -29,7 +29,7 @@ from pymodbus.framer import (
     ModbusTlsFramer,
 )
 from pymodbus.logging import Log
-from pymodbus.pdu import ModbusRequest
+from pymodbus.pdu import ModbusRequest, ModbusResponse
 from pymodbus.transport import CommType
 from pymodbus.utilities import ModbusTransactionState, hexlify_packets
 
@@ -167,13 +167,13 @@ class SyncModbusTransactionManager(ModbusTransactionManager):
         else:
             self.base_adu_size = -1
 
-    def _calculate_response_length(self, expected_pdu_size):
+    def _calculate_response_length(self, expected_pdu_size: int) -> int | None:
         """Calculate response length."""
         if self.base_adu_size == -1:
             return None
         return self.base_adu_size + expected_pdu_size
 
-    def _calculate_exception_length(self):
+    def _calculate_exception_length(self) -> int | None:
         """Return the length of the Modbus Exception Response according to the type of Framer."""
         if isinstance(self.client.framer, (ModbusSocketFramer, ModbusTlsFramer)):
             return self.base_adu_size + 2  # Fcode(1), ExceptionCode(1)
@@ -183,7 +183,9 @@ class SyncModbusTransactionManager(ModbusTransactionManager):
             return self.base_adu_size + 2  # Fcode(1), ExceptionCode(1)
         return None
 
-    def _validate_response(self, request: ModbusRequest, response, exp_resp_len, is_udp=False):
+    def _validate_response(
+            self, request: ModbusRequest, response: bytes | int, exp_resp_len: int | None, is_udp=False
+    ) -> bool:
         """Validate Incoming response against request.
 
         :param request: Request sent
@@ -208,7 +210,7 @@ class SyncModbusTransactionManager(ModbusTransactionManager):
             return mbap.get("length") == exp_resp_len
         return True
 
-    def execute(self, request: ModbusRequest):  # noqa: C901
+    def execute(self, request: ModbusRequest) -> ModbusResponse | bytes | ModbusIOException:  # noqa: C901
         """Start the producer to send the next request to consumer.write(Frame(request))."""
         with self._transaction_lock:
             try:
@@ -333,7 +335,9 @@ class SyncModbusTransactionManager(ModbusTransactionManager):
                 self.client.close()
                 return exc
 
-    def _retry_transaction(self, retries, reason, packet, response_length, full=False):
+    def _retry_transaction(
+            self, retries: int, reason: str, request: ModbusRequest, response_length: int | None, full=False
+    ) -> tuple[bytes, str | Exception | None]:
         """Retry transaction."""
         Log.debug("Retry on {} response - {}", reason, retries)
         Log.debug('Changing transaction state from "WAITING_FOR_REPLY" to "RETRYING"')
@@ -350,9 +354,11 @@ class SyncModbusTransactionManager(ModbusTransactionManager):
                 if response_length == in_waiting:
                     result = self._recv(response_length, full)
                     return result, None
-        return self._transact(packet, response_length, full=full)
+        return self._transact(request, response_length, full=full)
 
-    def _transact(self, request: ModbusRequest, response_length, full=False, broadcast=False):
+    def _transact(
+            self, request: ModbusRequest, response_length: int | None, full=False, broadcast=False
+    ) -> tuple[bytes, str | Exception | None]:
         """Do a Write and Read transaction.
 
         :param packet: packet to be sent
@@ -368,16 +374,12 @@ class SyncModbusTransactionManager(ModbusTransactionManager):
             packet = self.client.framer.buildPacket(request)
             Log.debug("SEND: {}", packet, ":hex")
             size = self._send(packet)
-            if (
-                isinstance(size, bytes)
-                and self.client.state == ModbusTransactionState.RETRYING
-            ):
+            if size and self.client.state == ModbusTransactionState.RETRYING:
                 Log.debug(
                     "Changing transaction state from "
                     '"RETRYING" to "PROCESSING REPLY"'
                 )
                 self.client.state = ModbusTransactionState.PROCESSING_REPLY
-                return size, None
             if self.client.comm_params.handle_local_echo is True:
                 if self._recv(size, full) != packet:
                     return b"", "Wrong local echo"
@@ -405,23 +407,22 @@ class SyncModbusTransactionManager(ModbusTransactionManager):
             result = b""
         return result, last_exception
 
-    def _send(self, packet: bytes, _retrying=False):
+    def _send(self, packet: bytes, _retrying=False) -> int:
         """Send."""
         return self.client.framer.sendPacket(packet)
 
-    def _recv(self, expected_response_length, full) -> bytes:  # noqa: C901
+    def _recv(self, expected_response_length: int | None, full: bool) -> bytes:  # noqa: C901
         """Receive."""
         total = None
         if not full:
             exception_length = self._calculate_exception_length()
+            min_size = expected_response_length
             if isinstance(self.client.framer, ModbusSocketFramer):
                 min_size = 8
             elif isinstance(self.client.framer, ModbusRtuFramer):
                 min_size = 4
             elif isinstance(self.client.framer, ModbusAsciiFramer):
                 min_size = 5
-            else:
-                min_size = expected_response_length
 
             read_min = self.client.framer.recvPacket(min_size)
             if len(read_min) != min_size:
@@ -462,7 +463,7 @@ class SyncModbusTransactionManager(ModbusTransactionManager):
                     if expected_response_length is not None:
                         expected_response_length -= min_size
                         total = expected_response_length + min_size
-                else:
+                if func_code >= 0x80 and exception_length:
                     expected_response_length = exception_length - min_size
                     total = expected_response_length + min_size
             else:
--- a/pymodbus/framer/old_framer_base.py
+++ b/pymodbus/framer/old_framer_base.py
@@ -80,7 +80,7 @@ class ModbusFramer:
         """
         return self.client.send(message)
 
-    def recvPacket(self, size: int) -> bytes:
+    def recvPacket(self, size: int | None) -> bytes:
         """Receive packet from the bus.
 
         With specified len


janiversen (commented):
Changes look good, but there are a couple of CI problems.

jameshilliard:
> there are a couple of CI problems

Hmm, fixed everything except a test coverage percentage check, not sure what to do about that one.

jameshilliard:
Tests passing now.

.github/workflows/ci.yml
```diff
@@ -98,7 +98,7 @@ jobs:
       - name: ruff
         if: matrix.run_lint == true
         run: |
-          ruff .
+          ruff check .
```
janiversen, line 101: We do not mix CI changes with other changes, simply because changing CI requires a bit more than just a PR.

pyproject.toml
```diff
@@ -240,7 +240,7 @@ exclude_also = [
     "if __name__ == .__main__.:",
     ]
 skip_covered = true
-fail_under = 90.0
+fail_under = 89.0
```
janiversen, line 243: A not acceptable change, we want to have a higher % not lower !!

janiversen (changes requested):

pyproject.toml
```diff
@@ -62,17 +62,17 @@ documentation = [
 development = [
     "build>=1.2.1",
     "codespell>=2.3.0",
-    "coverage>=7.5.3",
-    "mypy>=1.10.0",
-    "pylint>=3.2.3",
+    "coverage>=7.6.0",
```
janiversen, line 65: Bump 3party versions needs to be in a seperate PR; only with the bump.

janiversen (changes requested):

janiversen:
And just for the fun of it, I run into lowering the % in #2239, but instead of lowering the bar, I will examine the cause and add tests.

pyproject.toml
jameshilliard: Moved to #2241.

pyproject.toml
jameshilliard: Hmm, since this code is going to be [overhauled anyways](https://github.com/pymodbus-dev/pymodbus/pull/2232#pullrequestreview-2176716262) does test coverage percentage matter all that much right now? Getting typing working better should hopefully make it easier to identify unused codepaths and do further cleanup so I'm not sure we should be actually be all that concerned about coverage at the moment for things that are being refactored like the transaction code. I'm suspecting that there's some unused code branches in the sync client transaction code.

.github/workflows/ci.yml
jameshilliard: Hmm, tests don't pass without this change, btw I also added this to the dev deps bump PR in #2241 since it is needed there as well I think for tests to pass.

pymodbus/transaction.py
```diff
@@ -368,16 +374,13 @@ def _transact(self, request: ModbusRequest, response_length, full=False, broadca
             packet = self.client.framer.buildPacket(request)
             Log.debug("SEND: {}", packet, ":hex")
             size = self._send(packet)
-            if (
-                isinstance(size, bytes)
-                and self.client.state == ModbusTransactionState.RETRYING
-            ):
+            if self.client.state == ModbusTransactionState.RETRYING:
```
janiversen, line 377: This change do not look correct, we only want to enter the "if" if we received something.

This have nothing to do with typing, and it is not covered by tests.

pymodbus/transaction.py
```diff
@@ -368,16 +374,13 @@ def _transact(self, request: ModbusRequest, response_length, full=False, broadca
...
                 Log.debug(
                     "Changing transaction state from "
                     '"RETRYING" to "PROCESSING REPLY"'
                 )
                 self.client.state = ModbusTransactionState.PROCESSING_REPLY
-                return size, None
+                return b"", None
```
janiversen, line 383: This is wrong, we only change to REPLY when we have a reply.

This have nothing to do with typing, and it is not covered by tests.

pymodbus/transaction.py
```diff
@@ -420,8 +423,10 @@ def _recv(self, expected_response_length, full) -> bytes:  # noqa: C901
                 min_size = 4
             elif isinstance(self.client.framer, ModbusAsciiFramer):
                 min_size = 5
-            else:
+            elif expected_response_length:
```
janiversen, line 426: This do not seem right, however it is not covered by tests.

pymodbus/transaction.py
```diff
@@ -463,13 +468,17 @@ def _recv(self, expected_response_length, full) -> bytes:  # noqa: C901
                         expected_response_length -= min_size
                         total = expected_response_length + min_size
                 else:
+                    if exception_length is None:
```
janiversen, line 471: Not covered by test.

pymodbus/transaction.py
```diff
@@ -463,13 +468,17 @@ def _recv(self, expected_response_length, full) -> bytes:  # noqa: C901
...
+                        exception_length = 0
                     expected_response_length = exception_length - min_size
                     total = expected_response_length + min_size
             else:
                 total = expected_response_length
         else:
             read_min = b""
             total = expected_response_length
+        if expected_response_length is None:
```
janiversen, line 480: Not covered by test.

janiversen (changes requested):
May I politely suggest, you do typing in this PR and not a lot of code changes !

The typing in this PR is OK, the code changes are not covered by tests, and a number of the changes seems wrong.

pymodbus/transaction.py
jameshilliard: I removed the `isinstance(size, bytes)` condition, which based on the return type of `self._send(packet)` is always an integer. This was flagged by the type checker.

pymodbus/transaction.py
jameshilliard: > This is wrong, we only change to REPLY when we have a reply.

The issue here is that we can't return an integer, the calling code expects bytes to be returned.

pymodbus/transaction.py
jameshilliard: This was flagged by the type-checker, if `expected_response_length` is none we can't assign `min_size` to `expected_response_length` as we pass `min_size` to `self.client.framer.recvPacket` which expects an integer.

pymodbus/transaction.py
jameshilliard: This was flagged by the type-checker, if `exception_length` is `None` we would try to subtract from `exception_length` in `expected_response_length = exception_length - min_size` which would not work.

pymodbus/transaction.py
jameshilliard: This was flagged by the type-checker, `expected_response_length` can't be `None` as we pass `expected_response_length` to `self.client.framer.recvPacket` which expects an integer.

pyproject.toml
jameshilliard: > You do not need to “suspect”, use “pytest —cov” and you know.

I mean, the code additions were all entirely to fix type check errors, I have no idea if some of those branches in the sync client transaction code can even be hit in practice.

> Exactly refactoring code is a good reason to have near 100% coverage before starting, because it helps detect side effects.

Ideally but seems main problem is the type checker is flagging a bunch of issues(lack of checks for `None` for example) and I wouldn't know how to write tests to hit those conditions, the sync transaction code is super messy and hard to follow, at least with type checks we can start restricting the allowable types in various args/variables more easily.

jameshilliard:
> May I politely suggest, you do typing in this PR and not a lot of code changes !

All code changes made are directly due to issues being flagged by the typing additions.

janiversen:
Well some of them are wrong.

janiversen:
Please see my review comments, and when done request a new review.

pymodbus/transaction.py
janiversen: The code as it is works or not ?

In any case a test case would show the current code to be faulty, and your new code to be correct.

pymodbus/transaction.py
janiversen: Seems you need more typing.

pymodbus/transaction.py
janiversen: What has that to do with that you have not added test cases, that show why the old code was broken and the new code works.

pymodbus/transaction.py
janiversen: Did you read my comment ?

.github/workflows/ci.yml
alexrudd2: https://github.com/pymodbus-dev/pymodbus/pull/2241#issuecomment-2239907728

pymodbus/client/base.py
```diff
@@ -4,9 +4,9 @@
 import asyncio
 import socket
 from abc import abstractmethod
-from collections.abc import Awaitable, Callable
+from collections.abc import Callable, Coroutine
 from dataclasses import dataclass
-from typing import Any, cast
+from typing import Any, Union, cast
```
alexrudd2, line 9: Use `from __future__ import annotations` to match the rest of the codebase.

This allows using `X | Y` for instead of `Union[X | Y]` [Union types ](https://docs.python.org/3/library/stdtypes.html#union-type)

alexrudd2 (changes requested):
I am commenting only on the use of `from typing import Union`.

Although I think the other typing changes are worthwhile (in particular the existing code does not perfectly handle `None` or `ModbusException`), when I last looked at them I was unable to find a clean solution.

pymodbus/client/base.py
jameshilliard: > Use `from __future__ import annotations` to match the rest of the codebase.

https://github.com/pymodbus-dev/pymodbus/blob/72e3399d41c20719ac3e63412fd1be2b8cca3840/pymodbus/client/base.py#L2

That import already exists.

> This allows using `X | Y` for instead of `Union[X | Y]` [Union types ](https://docs.python.org/3/library/stdtypes.html#union-type)

Not sure why exactly, but the normal style doesn't seem to work here.

alexrudd2: ??

Works for me with both `pyright` and `mypy`.  Did you mismatch a bracket, perhaps?

image

pyproject.toml
janiversen: Well maybe you have defined the wrong types, which at least seems to be a possibility.

If you change code, we need to verify why it (in your opinion) did not work earlier, and why it works now. I strongly believe that it is a sound and natural demand ! we do not do changes just because the typing you define causes problems.

That all being said, there might of  and most likely are bugs in the code, which proper typing will bring to the surface, but I am sure you understand, why proof is needed..

If it's that hard to write tests, then maybe it is also hard to write the correct typing, and not just guessing what is correct !! meaning you could always reduce the PR to the part where you know how to write tests.

Adding types to code you do not understand, it quite a mountain to climb !! it is hard enough to add typing to code you understand. Just to give you an example, we had discussions during month how to add typing to mixin.py until someone suggested templates.

pymodbus/transaction.py
janiversen: But it still needs to get argumented and tested. I see quite a different behavior, because you return a fixed b"", which basically never should be returned in a retry situation.

pymodbus/transaction.py
janiversen: None can be converted to a 0 without problems.

pymodbus/client/base.py
jameshilliard: > Works for me with both `pyright` and `mypy`. Did you mismatch a bracket, perhaps?

Oh, the error is coming from `pylint`, not `pyright` or `mypy`:
```python
$ pylint --recursive=y examples pymodbus test
************* Module pymodbus.client.base
pymodbus/client/base.py:23:61: E1131: unsupported operand type(s) for | (unsupported-binary-operation)
pymodbus/client/base.py:228:45: E1131: unsupported operand type(s) for | (unsupported-binary-operation)
```

pymodbus/transaction.py
jameshilliard: > The code as it is works or not ?

I guess? I mean the tests seem to pass either way so it's a bit hard to tell.

I went ahead and removed this return entirely which also seems to work.

pymodbus/transaction.py
jameshilliard: Apparently tweaking the style here a bit using original behavior got it passing the type checker.

pymodbus/transaction.py
jameshilliard: > What has that to do with that you have not added test cases, that show why the old code was broken and the new code works.

Tweaked this a bit more to get it passing the type checker a slightly different way, I'm not sure how we would actually hit this codepath however so a bit hard to write tests.

pymodbus/transaction.py
jameshilliard: > None can be converted to a 0 without problems.

Updated to no longer convert to 0.

pymodbus/client/base.py
alexrudd2: Bah, appears to be a `pylint` bug.  https://github.com/pylint-dev/pylint/issues/7381.

Use `# pylint: disable=unsupported-binary-operation`, I guess.

pymodbus/transaction.py
jameshilliard: > I see quite a different behavior, because you return a fixed b"", which basically never should be returned in a retry situation.

I removed the return entirely and changed this check to verify returned size is not 0.

pymodbus/client/base.py
jameshilliard: Hmm, the normal `|` style seems to be entirely broken on python 3.9:
```python
ERROR test/framers/test_old_framers.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/framers/test_old_framers.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/framers/test_old_framers.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_client/test_client.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_client/test_client.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_client/test_client.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_client/test_client_sync.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_client/test_client_sync.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_client/test_client_sync.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_examples/test_client_server_async.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_examples/test_client_server_async.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_examples/test_client_server_async.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_examples/test_client_server_sync.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_examples/test_client_server_sync.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_examples/test_client_server_sync.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_examples/test_examples.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_examples/test_examples.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/sub_examples/test_examples.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/test_network.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/test_network.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/test_network.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

alexrudd2: To date, the test have not had type hints (since it seemed like low value).  No problem with adding them, but in this case you will need to add `from __future__ import annotations` for the first time.

jameshilliard: > No problem with adding them, but in this case you will need to add `from __future__ import annotations` for the first time.

Tried this, doesn't seem to help, looks to just be a limitation of python 3.9.

janiversen: I agree with @alexrudd2 adding typing to tests, hav3 no real value and will most likely just complicate writing tests.

Please do not go down that route.

janiversen:
Please stop extending this PR, you still have not solved the basic problems in transaction.py

Adding test files to the mix, will only ensure that you will not ge5 the PR reviewed !

pymodbus/client/base.py
jameshilliard: > adding typing to tests, hav3 no real value and will most likely just complicate writing tests.

So I'm not really trying to add types to the tests so much as make it so that we don't get type errors when we typecheck the tests.

jameshilliard:
> Please stop extending this PR, you still have not solved the basic problems in transaction.py

Yeah, I should probably split off those test fixes, it ended up being more fixes than I expected. Was trying to fix the mypy typing errors in tests calling transaction.py to make it easier to validate typing there. I found a few issues already with the tests not always using the correct types at least.

janiversen:
the only reason to mix test files with production code, is if the production code changes needs changes in test !!!

That is an unbreakable rule, just like the CI rule.

However I can just as well tell you, your test changes are unwanted because the only complicate tests and does nothing to improve the production code.

jameshilliard:
> However I can just as well tell you, your test changes are unwanted because the only complicate tests and does nothing to improve the production code.

I'm confused, the test changes are fixing bugs in the tests flagged by mypy. If the mocking in the tests do not use the correct types then the tests are much less useful aren't they?

pyproject.toml
jameshilliard: > Well maybe you have defined the wrong types, which at least seems to be a possibility.

The tests also appear to use the wrong types in a number of places which is why I was trying to fix those in #2249.

> If it's that hard to write tests, then maybe it is also hard to write the correct typing, and not just guessing what is correct !! meaning you could always reduce the PR to the part where you know how to write tests.

It's also hard to use the tests to validate typing issues when there's a lot of [typing issues being flagged by mypy](https://github.com/pymodbus-dev/pymodbus/pull/2249#issuecomment-2241202971) even on `dev`.

> Adding types to code you do not understand, it quite a mountain to climb !! it is hard enough to add typing to code you understand. Just to give you an example, we had discussions during month how to add typing to mixin.py until someone suggested templates.

If we fix the typing errors in the tests it makes it easier to validate that the types here are correct in transaction.py doesn't it?

janiversen:
We do not run mypy and a number of the other tools an test, specifically because there are a larger number of lines flagged, which actually are ok or even needed.

janiversen:
And can we please agree "bugs" are when something do not want as we want, violating types is surely not in that category (and in many cases because we want it).

pymodbus/transaction.py
```diff
@@ -405,23 +407,22 @@ def _transact(self, request: ModbusRequest, response_length, full=False, broadca
...
         """Send."""
         return self.client.framer.sendPacket(packet)
 
-    def _recv(self, expected_response_length, full) -> bytes:  # noqa: C901
+    def _recv(self, expected_response_length: int | None, full: bool) -> bytes:  # noqa: C901
         """Receive."""
         total = None
         if not full:
             exception_length = self._calculate_exception_length()
+            min_size = expected_response_length
```
janiversen, line 419: This change makes no sense, you set the variable just to overwrite it, that is pure waste.

janiversen (changes requested):
There are code changes, which seems untested, and I believe the code changes might be incorrect.

Would be nice to have a test that shows why the old code was wrong, and the new code solves the problem.

pymodbus/transaction.py
jameshilliard: For some reason mypy wasn't liking the previous version here. Not sure what the best approach would be.

pymodbus/client/base.py
jameshilliard: > typecchecking tests is an unwanted feature.

I'm still confused about this, I mean if the tests have incorrect types then they are likely to cause issues.

> Even if we were to consider it, you should NOT mix production code changes with test changes.

Yeah, going to split things out more.

.github/workflows/ci.yml
```diff
@@ -93,7 +93,7 @@ jobs:
       - name: mypy
         if: matrix.run_lint == true
         run: |
-          mypy pymodbus
+          mypy pymodbus test/test_transaction.py test/transport/test_comm.py
```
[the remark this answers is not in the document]
jameshilliard: > Why are you adding ci.yaml when you know that is not the way to make CI changes ?

Was trying to use CI to validate some tests(there was a flaky test that was failing on windows). I removed this now.

test/framers/test_old_framers.py
```diff
@@ -1,4 +1,6 @@
 """Test framers."""
+from __future__ import annotations
```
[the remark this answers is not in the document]
jameshilliard: removed

jameshilliard:
> We do not run mypy and a number of the other tools an test, specifically because there are a larger number of lines flagged, which actually are ok or even needed.

I'm aware there are currently a lot of mypy errors with the tests...but I see no indication that these aren't fixable by cleaning up the test code in various ways, for example [this](https://github.com/pymodbus-dev/pymodbus/pull/2251) eliminates `method-assign` mypy errors in `test/test_transaction.py` while also eliminating some ugly hacks with how the mocking was being done.

> And can we please agree "bugs" are when something do not want as we want, violating types is surely not in that category (and in many cases because we want it).

I mean, tests using the wrong types are kinda a problem IMO as it makes it more difficult to tell if the production code is correct.

alexrudd2:
> And can we please agree "bugs" are when something do not want as we want, violating types is surely not in that category (and in many cases because we want it).

Here I must disagree.  :)  Although `mypy` has some false positives, making it happy fixes bugs and prevents new ones.  I'd say 20% of the time type errors are actual logic errors, which is high enough to make it worthwhile.

janiversen:
> Here I must disagree.  :)  Although `mypy` has some false positives, making it happy fixes bugs and prevents new ones.  I'd say 20% of the time type errors are actual logic errors, which is high enough to make it worthwhile.

Well if there are actual logic errors, then the code do not work as we want, and thus it is a bug.....problem is that mypy is very do not allow to use the full power of python, being a non typed language, e.g. assigning different types to a variable makes mypy mark it, but it's legal and often in test a nice way to test different parameter types,

jameshilliard:
> Well if there are actual logic errors, then the code do not work as we want, and thus it is a bug.....problem is that mypy is very do not allow to use the full power of python

I mean, I think there's usually ways to make tests play nice with mypy while making the test code cleaner like in #2251 and for cases where it's not possible can just add ignores.

> being a non typed language

Python is both a [strongly typed](https://en.wikipedia.org/wiki/Strong_and_weak_typing) and a [dynamically typed](https://en.wikipedia.org/wiki/Type_system#Dynamic_type_checking_and_runtime_type_information) language, it's certainly not a "non typed language" whatever that means.

> assigning different types to a variable makes mypy mark it, but it's legal and often in test a nice way to test different parameter types,

IMO not being sufficiently strict with types is a major issue that makes the transaction code hard to work with.

janiversen:
> I mean, I think there's usually ways to make tests play nice with mypy while making the test code cleaner like in #2251 and for cases where it's not possible can just add ignores.

I am sure there are.

> Python is both a [strongly typed](https://en.wikipedia.org/wiki/Strong_and_weak_typing) and a [dynamically typed](https://en.wikipedia.org/wiki/Type_system#Dynamic_type_checking_and_runtime_type_information) language, it's certainly not a "non typed language" whatever that means.

It is a non typed language, because you can start using variables without giving them a type, and a variable freely be assigned all kind of types. Python allows typing, but the interpreter do not use that information, which is the definition of a non typed language.



> IMO not being sufficiently strict with types is a major issue that makes the transaction code hard to work with.

Yes but it one of the strong features of python, if you want strict typing then start using e,g, C++

jameshilliard:
> It is a non typed language, because you can start using variables without giving them a type, and a variable freely be assigned all kind of types.

So...dynamically typed I think would be what you're referring to here. Where did the term "no typed language" come from? I'm not seeing much of anything in google about that term.

> Python allows typing, but the interpreter do not use that information, which is the definition of a non typed language.

This is...not accurate at all, what do you think a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) is if the interpreter doesn't use typing information?

Python is a strongly typed language, strong typing means that variables do have a type and that the type matters when performing operations on a variable.

> Yes but it one of the strong features of python

I think you're mixing up Python with JavaScript-like languages that are both weakly typed and dynamically typed. Python's type system is more lenient than C++ while still being strongly typed which I prefer to weakly typed languages like JavaScript.

> if you want strict typing then start using e,g, C++

C++ is statically typed with fairly strong typing. Python is dynamically typed with similarly strong typing(I think in some ways even stronger than C++), one reason I like Python is due to it being dynamically and strongly typed. I try to stay away from JavaScript-like languages when possible in general primarily due to them using weak typing which generally results in more bugs.

janiversen:
> So...dynamically typed I think would be what you're referring to here. Where did the term "no typed language" come from? I'm not seeing much of anything in google about that term.


No dynamically typed (e,g, java) is different, as the name indicates a type is added dynamically, and the interpreter uses that to e.g. assign storage.

> This is...not accurate at all, what do you think a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) is if the interpreter doesn't use typing information?

For the interpreter everything is stored in a object class, of course within that class there can be types.


> I think you're mixing up Python with JavaScript-like languages that are both weakly typed and dynamically typed. Python's type system is more lenient than C++ while still being strongly typed which I prefer to weakly typed languages like JavaScript.

No I am not ! I have been around from before all these fancy languages, and spent quite a lot of time helping make parsers and interpreters.

Fact is, you can write a perfectly valid python program, without using a single type definition, and if you care to take a look at the python parser, it skips all type information, which is actually currently one of the bigger problems for making an optimizer in coython.

Of course the object class used by the runtime, do checks on assignment etc, but that is due to the internals of the object, not the interpreter,



> C++ is statically typed with fairly strong typing. Python is dynamically typed with similarly strong typing(I think in some ways even stronger than C++), one reason I like Python is due to it being dynamically and strongly typed. I try to stay away from JavaScript-like languages when possible in general primarily due to them using weak typing which generally results in more bugs.

Your sentence is wrong. python allow typing, but do not use that information as it is purely optional.

Easy proof, make an array with 1.000.000 entries with a mixture of integers and floats, measured the actual memory footprint. Then type the array and use only integers, the active memory footprint is the same. A typed language would use the typing information to optimize the memory footprint.

But anyhow believe what you want to believe, I am too old to give free lessons in compiler techniques.

jameshilliard:
> No dynamically typed (e,g, java) is different, as the name indicates a type is added dynamically, and the interpreter uses that to e.g. assign storage.

Java is actually statically typed, JavaScript is dynamically typed at least based on the definitions that I'm familiar with:
["A statically-typed language is a language (such as Java, C, or C++) where variable types are known at compile time. In most of these languages, types must be expressly indicated by the programmer; in other cases (such as OCaml), type inference allows the programmer to not indicate their variable types."](https://developer.mozilla.org/en-US/docs/Glossary/Static_typing)

> For the interpreter everything is stored in a object class, of course within that class there can be types.

Everything may be an object but "there can be types" isn't very precise as ["Every object has an identity, a type and a value."](https://docs.python.org/3/reference/datamodel.html#objects-values-and-types), I think this is somewhat important for making python's strong typing system work.

> Fact is, you can write a perfectly valid python program, without using a single type definition, and if you care to take a look at the python parser, it skips all type information, which is actually currently one of the bigger problems for making an optimizer in coython.

That's true, ["The Python runtime does not enforce function and variable type annotations. They can be used by third party tools such as type checkers, IDEs, linters, etc."](https://docs.python.org/3/library/typing.html). It is of course also true that it's harder to optimize dynamically typed languages like Python.

> Your sentence is wrong. python allow typing, but do not use that information as it is purely optional.

The sort of [typing](https://docs.python.org/3/library/typing.html) I think you are referring to is just optional type annotations, so it is true that typing is optional in that sense. However ["In a dynamically typed language, a variable is simply a value bound to a name; the value has a type -- like "integer" or "string" or "list" -- but the variable itself doesn't."](https://wiki.python.org/moin/Why%20is%20Python%20a%20dynamic%20language%20and%20also%20a%20strongly%20typed%20language).

> Easy proof, make an array with 1.000.000 entries with a mixture of integers and floats, measured the actual memory footprint. Then type the array and use only integers, the active memory footprint is the same. A typed language would use the typing information to optimize the memory footprint.

That's true, ["The Python runtime does not enforce function and variable type annotations."](https://docs.python.org/3/library/typing.html). ["In a dynamically typed language, objects still have a type, but it is determined at runtime."](https://wiki.python.org/moin/Why%20is%20Python%20a%20dynamic%20language%20and%20also%20a%20strongly%20typed%20language)

> But anyhow believe what you want to believe, I am too old to give free lessons in compiler techniques.

I think for the most part you are just using different terminology than what I'm familiar with.

janiversen:
Most people do not care about how the compiler works as long as it does the job, which is a pity because it often leads to programs with a bad performance.

If you solely look at typing from the outside, you really miss a lot, in order to make big scale programs. 

As you correctly noted dynamically typed means a variable is assigned a type at runtime...but you missed the point everything in python is inherited from THE Object so the type of everything in python is Object, This is the reason why you can e,g, assign a string to a function, and thus totally change how it works.

janiversen:
But as I said, stay in your beliefs! 

Fact for this project, typing is accepted because a couple of contributors care about it. Every change that complicates the code, without adding other benefit than pleasing mypy is not accepted.

jameshilliard:
> Most people do not care about how the compiler works as long as it does the job, which is a pity because it often leads to programs with a bad performance.

I am [heavily involved](https://github.com/buildroot/buildroot/commits?author=jameshilliard) with [python toolchain/package integration](https://github.com/buildroot/buildroot/blob/2024.05.1/DEVELOPERS#L1473) as part of the open source [buildroot](https://buildroot.org/) embedded Linux cross-compilation project and have worked on complex cross language toolchain integrations such as buildroot's python rust extension integration(used by libraries like python-cryptography which use [pyo3 bindings](https://github.com/PyO3/pyo3)). So I'm not at all unfamiliar with a lot of lower level details regarding the Python toolchain/interpreter. Python is not a particular easy language to cross compile either but buildroot has excellent support for python cross compilation partially due to a lot of work I've done on it(pep517 build system integrations for example).

> you missed the point everything in python is inherited from THE Object so the type of everything in python is Object, This is the reason why you can e,g, assign a string to a function, and thus totally change how it works.

I don't think I'm disagreeing with that, that's just an implementation detail of python's dynamic typing system.

PR #2241, opened 2024-07-18
Update dev dependencies

jameshilliard:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,17 +62,17 @@ documentation = [
 development = [
     "build>=1.2.1",
     "codespell>=2.3.0",
-    "coverage>=7.5.3",
-    "mypy>=1.10.0",
-    "pylint>=3.2.3",
+    "coverage>=7.6.0",
+    "mypy>=1.10.1",
+    "pylint>=3.2.5",
     "pytest>=8.2.2",
-    "pytest-asyncio>=0.23.7",
+    "pytest-asyncio>=0.23.8",
     "pytest-cov>=5.0.0",
     "pytest-profiling>=1.7.0",
     "pytest-timeout>=2.3.1",
     "pytest-xdist>=3.6.1",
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -98,7 +98,7 @@ jobs:
       - name: ruff
         if: matrix.run_lint == true
         run: |
-          ruff .
+          ruff check .
```
janiversen, line 101: This a mixture of CI and version bumps, as written earlier:
- CI changes are in seperate PR
- version bumps are in a seperate PR.

This is done for several reasons:
- mixed changes are harder to bisect
- version bumps need to be clearly identifiable in the changelog for users
- CI changes and version bumps needs extra handling in github actions

janiversen (changes requested):

.github/workflows/ci.yml
jameshilliard: Split the CI change out to #2242.

janiversen: This one still contains the CI part.

jameshilliard: Removed it from this one(although tests fail again without that).

janiversen: The test should not be affected by the 3rd party version numbers, but I will ask CI to run on this PR (due to CI setup it is more than just starting CI).

janiversen (approved):
Lgtm

alexrudd2:
The change from `ruff .` to `ruff check .` is not a change in CI functionality, it's a required migration for Ruff `0.5.0`.

See documentation of the change here: https://astral.sh/blog/ruff-v0.5.0#removed-deprecated-features

janiversen:
@alexrudd2 I agree, I read the same, but nonetheless CI changes are in a separate CI, which are merged.

PR #2244, opened 2024-07-19
remove kwargs client

janiversen:
kwargs is gone from all clients !!!

Hopefully that will reduce the number of questions about wrong parameters.

### DIFF
--- a/pymodbus/client/tcp.py
+++ b/pymodbus/client/tcp.py
@@ -4,13 +4,14 @@ from __future__ import annotations
 import select
 import socket
 import time
+from collections.abc import Callable
 from typing import Any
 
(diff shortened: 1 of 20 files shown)


pymodbus/client/tcp.py
```diff
@@ -54,24 +57,45 @@ async def run():
 
     socket: socket.socket | None
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
         host: str,
-        port: int = 502,
         framer: FramerType = FramerType.SOCKET,
+        port: int = 502,
```
sjrodahl, line 64: For reference: this breaks the interface for those who used port as a positional argument

janiversen: Yes, as noted in API changes.

PR #2249, opened 2024-07-20
Fix mypy type checking errors in test_transaction.py

jameshilliard:
This fixes a number of issues with tests using the wrong types that was flagged by mypy.

PR #2250, opened 2024-07-20
Fix some mypy type checking errors in test_transaction.py

alexrudd2:
Let's start over with #2249.

I've cherry-picked the least contentious changes.  Rather than just making `mypy` happy, let's use it to make simpler and more robust tests.


janiversen (approved):
I still see a number of the changes, as just different styles...but at least the changes are understandable.

PR #2251, opened 2024-07-20
Use mock.patch.object to avoid protected access errors.

jameshilliard:
This seems to be cleaner and play better with mypy.


janiversen (approved):
Cute, it even makes the code more compact.

alexrudd2:
Agreed, it's a clean and easy to understand pattern, and removes a lot of pylint noise too.

PR #2252, opened 2024-07-21
Fix remaining mypy errors in test_transaction.py

jameshilliard:
Initialize class attributes to desired values instead of `None`.

Use `mock.patch` for `test_execute` `client.framer` methods.

Use `mock.create_autospec` for mocking the modbus clients.

### DIFF
--- a/test/test_transaction.py
+++ b/test/test_transaction.py
@@ -2,11 +2,14 @@
 from itertools import count
 from unittest import mock
 
+from pymodbus.client.base import ModbusBaseSyncClient
 from pymodbus.exceptions import (
     ModbusIOException,
 )
 from pymodbus.factory import ServerDecoder
+from pymodbus.framer import ModbusFramer
 from pymodbus.pdu import ModbusRequest
+from pymodbus.pdu.bit_read_message import ReadDiscreteInputsRequest
 from pymodbus.transaction import (
     ModbusAsciiFramer,
     ModbusRtuFramer,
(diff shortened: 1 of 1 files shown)


test/test_transaction.py
```diff
@@ -23,26 +27,34 @@
...
-    _ascii = None
-    _manager = None
-    _tm = None
+    client: ModbusBaseSyncClient
+    decoder: ServerDecoder
+    _tcp: ModbusSocketFramer
+    _tls: ModbusTlsFramer
+    _rtu: ModbusRtuFramer
+    _ascii: ModbusAsciiFramer
+    _manager: SyncModbusTransactionManager
```
jameshilliard, line 36: So...if we remove these mypy doesn't care but pylint then complains, when they are set to `None` then mypy does complain, best solution I've come up with so far is to just set the types here like this as this plays well with both mypy and pylint.

test/test_transaction.py
```diff
@@ -23,26 +27,34 @@
...
-        self.client = None
+        self.client = mock.create_autospec(ModbusBaseSyncClient)
+        self.client.framer = mock.create_autospec(ModbusFramer)
         self.decoder = ServerDecoder()
-        self._tcp = ModbusSocketFramer(decoder=self.decoder, client=None)
-        self._tls = ModbusTlsFramer(decoder=self.decoder, client=None)
-        self._rtu = ModbusRtuFramer(decoder=self.decoder, client=None)
-        self._ascii = ModbusAsciiFramer(decoder=self.decoder, client=None)
+        self._tcp = ModbusSocketFramer(
+            decoder=self.decoder, client=mock.create_autospec(ModbusTcpClient)
```
janiversen, line 47: We are not using the client, and if the production code start using the client, having it as None ensures a failure. Your change removes that detection.

test/test_transaction.py
```diff
@@ -165,7 +186,7 @@ def test_execute(self, mock_get_transaction, mock_recv, mock_time):
         mock_recv.reset_mock(
             side_effect=ModbusIOException()
         )
-        client.framer.processIncomingPacket.side_effect = mock.MagicMock(
+        mock_process_incoming_packet.reset_mock(
```
janiversen, line 189: Change of functionality.

janiversen (changes requested):
This PR basically introduces a lot of noise, without solving any problems.

If the global variables disturb, then rewrite the tests to use class/local variables.

Removing the global variables would probably lead to a simpler code, since there are really no reason to have them global.

It would be nice if you, instead of just making a pull request like this blindly, thought about how to solve the problem and actually solved it.

janiversen:
Ps. pleasing mypy is NOT a reason! nor is having strongly typing in tests.

test/test_transaction.py
jameshilliard: Ah, so looks like mypy wasn't flagging this as a type issue but pycharm was, weird. it's getting picked up as an issue since this type hint does not have `None` as an option, but adding `None` causes other type check issues. I'll look at that later reverted for now.

https://github.com/pymodbus-dev/pymodbus/blob/385bbc5ef6d2189f7426961e7461e6330ff0208c/pymodbus/framer/old_framer_base.py#L36

test/test_transaction.py
jameshilliard: Changed to:
```python
mock_process_incoming_packet.side_effect = ModbusIOException()
```

jameshilliard:
> Ps. pleasing mypy is NOT a reason! nor is having strongly typing in tests.

Came up with another approach that works on both mypy and pylint that gets rid of the type hints.

janiversen:
Easy, do NOT use mypy ! pylint works.

It is NOT an objective for this project to have mypy run cleanly on test (pylint already runs cleanly).

This constant flow of PRs, targeting the same file and with only cosmetic changes, are taking a lot of my time. I need to do a in depth review, because there are often changes that change the tests, which clearly is unwanted.

janiversen:
And please stop force pushing every commit, push the last one before requesting a review, you are using our CPU minutes, causing important runs to be delayed.

alexrudd2:
> And please stop force pushing every commit, push the last one before requesting a review, you are using our CPU minutes, causing important runs to be delayed.

Actually yeah, @jameshilliard it should be straightforward to run the CI on a fork, I don't think anything needs credentials.

test/test_transaction.py
alexrudd2: Yeah, agreed.  I forgot about `pylint` since I almost never run it :S

PR #2253, opened 2024-07-21
Use mock.patch.object to avoid protected access errors.

jameshilliard:

### DIFF
--- a/test/framers/test_tbc_transaction.py
+++ b/test/framers/test_tbc_transaction.py
@@ -12,6 +12,7 @@ from pymodbus.transaction import (
     ModbusRtuFramer,
     ModbusSocketFramer,
     ModbusTlsFramer,
+    ModbusTransactionManager,
     SyncModbusTransactionManager,
 )
 
@@ -90,7 +91,9 @@ class TestTransaction:  # pylint: disable=too-many-public-methods
             )
 
     @mock.patch("pymodbus.transaction.time")
-    def test_execute(self, mock_time):
+    @mock.patch.object(SyncModbusTransactionManager, "_recv")
+    @mock.patch.object(ModbusTransactionManager, "getTransaction")
+    def test_execute(self, mock_get_transaction, mock_recv, mock_time):
         """Test execute."""
         mock_time.time.side_effect = count()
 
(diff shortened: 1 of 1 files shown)


test/framers/test_tbc_transaction.py
```diff
@@ -90,7 +91,9 @@ def test_calculate_exception_length(self):
             )
 
     @mock.patch("pymodbus.transaction.time")
-    def test_execute(self, mock_time):
+    @mock.patch.object(SyncModbusTransactionManager, "_recv")
```
janiversen, line 94: This is wrong, it change the scope from class to local.

test/framers/test_tbc_transaction.py
```diff
@@ -353,19 +353,18 @@ def callback(data):
         # for name in ("transaction_id", "protocol_id", "slave_id"):
         #     assert getattr(expected, name) == getattr(actual, name)
 
-    def test_tcp_framer_packet(self):
+    @mock.patch.object(ModbusRequest, "encode")
```
janiversen, line 356: Seems a number of the patch objects could be done at class level.

janiversen (changes requested):
A few comments, but the most important one:

You already submitted #2251 to solve this problem, seems you did not complete your work. Having 2 PR solving the same problem in the same file is over the top, and wasting my time, since I have to review them.

Please do not submit more PR to test_transaction.py unless they contain more than cosmetic changes, meaning real code change.

jameshilliard:
> Having 2 PR solving the same problem in the same file is over the top, and wasting my time, since I have to review them.

Um, it's a different file right? Path for this PR looks to be `test/framers/test_tbc_transaction.py` while #2251 is for `test/test_transaction.py`. So `test/framers/test_tbc_transaction.py` is rather similar to `test/test_transaction.py`, are we supposed to have both of these tests or just one?

janiversen:
If you had a look at the files you would see it’s the same file, just copied to prepare for major changes.

jameshilliard:
> If you had a look at the files you would see it’s the same file,

Yeah...I missed that somehow, had assumed it wasn't an exact duplicate so had applied the same fixes from the other one.

> just copied to prepare for major changes.

Um, so I should make sure changes are duplicated between the files?

janiversen:
No you should not touch a file marked "to be changed" unless you want to change it as intended.

jameshilliard:
> change it as intended.

How is it intended to be changed?

janiversen:
Read the issues, it will be integrated in framers.

PR #2258, opened 2024-07-23
Feature/simulator addressing

Esco441-91:
# Summary

In order to be able to limit the simulation to only specific ModBus unit address, check for existence of `unit_address` in the configuration file. If present, construct the `ModbusServerContext` as specified for non-single unit address.

Note that field `address` is already reserved for TCP addressing. Hence the term `unit_address` should be close enough for the context.

# Testing

1. Only tested this manually on personal setup, so no extensive testing was done.
2. Made sure the setup behaved the same as before without the field being present.
3. Ran pytest, no regressions

### DIFF
--- a/doc/source/library/simulator/config.rst
+++ b/doc/source/library/simulator/config.rst
@@ -69,7 +69,7 @@ The entry “framer” allows the following values:
 - “socket” to use :class:`pymodbus.framer.ModbusSocketFramer`.
 
 Optional entry "device_id" will limit server to only accept if. If
-not set, the server will respond to all addresses.
+not set, the server will accept all device id.
 
 .. warning::
 
--- a/doc/source/library/simulator/config.rst
+++ b/doc/source/library/simulator/config.rst
@@ -68,7 +68,7 @@ The entry “framer” allows the following values:
 - “tls” to use :class:`pymodbus.framer.ModbusTlsFramer`,
 - “socket” to use :class:`pymodbus.framer.ModbusSocketFramer`.
 
-Optional entry "device_id" will limit server to only accept if. If
+Optional entry "device_id" will limit server to only accept a single id. If
 not set, the server will accept all device id.
 
 .. warning::
--- a/doc/source/library/simulator/config.rst
+++ b/doc/source/library/simulator/config.rst
@@ -68,6 +68,9 @@ The entry “framer” allows the following values:
 - “tls” to use :class:`pymodbus.framer.ModbusTlsFramer`,
 - “socket” to use :class:`pymodbus.framer.ModbusSocketFramer`.
 
+Optional entry "device_id" will enable address-specific server addressing. If
+not set, the server will respond to all addresses.
+
 .. warning::
 
     not all "framer" types can be used with all “comm” types.
--- a/doc/source/library/simulator/config.rst
+++ b/doc/source/library/simulator/config.rst
@@ -68,7 +68,7 @@ The entry “framer” allows the following values:
 - “tls” to use :class:`pymodbus.framer.ModbusTlsFramer`,
 - “socket” to use :class:`pymodbus.framer.ModbusSocketFramer`.
 
-Optional entry "device_id" will enable address-specific server addressing. If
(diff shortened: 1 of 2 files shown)


doc/source/library/simulator/config.rst
```diff
@@ -68,6 +68,9 @@ The entry “framer” allows the following values:
 - “tls” to use :class:`pymodbus.framer.ModbusTlsFramer`,
 - “socket” to use :class:`pymodbus.framer.ModbusSocketFramer`.
 
+Optional entry "unit_address" will enable address-specific server addressing. If
```
janiversen, line 71: The modbus term is device_id or slave (whic( we use throughout), please change.

pymodbus/server/simulator/http_server.py
```diff
@@ -154,7 +154,15 @@ def __init__(
         self.datastore_context = ModbusSimulatorContext(
             device, custom_actions_dict or {}
         )
-        datastore = ModbusServerContext(slaves=self.datastore_context, single=True)
+        datastore = None
+        if "unit_address" in server:
+            # Designated ModBus unit address. Will only serve data if the address matches
+            address = int(server["unit_address"])
```
janiversen, line 160: No need to make a local variable, with a confusing name, combine the lines.

janiversen (changes requested):

pymodbus/server/simulator/http_server.py
Esco441-91: Fixed

doc/source/library/simulator/config.rst
Esco441-91: Changed to device_id.

(I'm grateful for your comment. The preferred terminology is sometimes not easy to deduce in cases like this 😃 )

janiversen (changes requested):

doc/source/library/simulator/config.rst
```diff
@@ -68,6 +68,9 @@ The entry “framer” allows the following values:
 - “tls” to use :class:`pymodbus.framer.ModbusTlsFramer`,
 - “socket” to use :class:`pymodbus.framer.ModbusSocketFramer`.
 
+Optional entry "device_id" will enable address-specific server addressing. If
```
janiversen, line 71:
```suggestion
Optional entry "device_id" will limit server to only accept if. If
```

doc/source/library/simulator/config.rst
```diff
@@ -68,6 +68,9 @@ The entry “framer” allows the following values:
...
+not set, the server will respond to all addresses.
```
janiversen, line 72:
```suggestion
not set, the server will accept all device id.
```

pymodbus/server/simulator/http_server.py
```diff
@@ -154,7 +154,14 @@ def __init__(
         self.datastore_context = ModbusSimulatorContext(
             device, custom_actions_dict or {}
         )
-        datastore = ModbusServerContext(slaves=self.datastore_context, single=True)
+        datastore = None
+        if "device_id" in server:
+            # Designated ModBus unit address. Will only serve data if the address matches
+            datastore = ModbusServerContext(slaves={int(server["device_id"]): self.datastore_context}, single=False)
```
janiversen, line 160:
```suggestion
            datastore = ModbusServerContext(slaves={int(server["device_id"]): self.datastore_context}, single=True)
```

janiversen (changes requested):

doc/source/library/simulator/config.rst
```diff
@@ -68,6 +68,9 @@ The entry “framer” allows the following values:
 - “tls” to use :class:`pymodbus.framer.ModbusTlsFramer`,
 - “socket” to use :class:`pymodbus.framer.ModbusSocketFramer`.
 
+Optional entry "device_id" will limit server to only accept if. If
```
janiversen, line 71:
```suggestion
Optional entry "device_id" will limit server to only accept a single id. If
```

janiversen (changes requested):

janiversen (approved):

Issue #2259, opened 2024-07-23
Server and client decoders handle exceptions differently


alexrudd2:
`ServerDecoder`
https://github.com/pymodbus-dev/pymodbus/blob/1b2cb7eb6651c2cb9ec0eb8323e645b0758fc26b/pymodbus/factory.py#L92-L102

`ClientDecoder`
https://github.com/pymodbus-dev/pymodbus/blob/1b2cb7eb6651c2cb9ec0eb8323e645b0758fc26b/pymodbus/factory.py#L229-L241

The generic `except Exception` was added a long time ago in [`1.5.0`](https://github.com/pymodbus-dev/pymodbus/pull/294/files#diff-09a427404023a7916a4afacf2ff25eac10b00120218a033b4f866ab778076850)

It doesn't seem right to have them differ; which is preferred?

janiversen:
you are right, the first one is better (without the generic exception).

PR #2260, opened 2024-07-23
More descriptive decoder exceptions

alexrudd2:
Closes https://github.com/pymodbus-dev/pymodbus/issues/2259.

Three current tests are relevant for badly encoded `Response`.

(1) 
`test_client_factory_fails()` calls `ClientDecoder.decode(None)` and expects `None`.
This would cause a `TypeError`, but it's monkey-patched to throw a `ModbusException` instead.  (The monkey patching was never necessary with the catch-all `Except`)

(2) `test_server_factory_fails()` calls  `ServerDecoder.decode(None)` and expects `None`.
This would cause a `TypeError`, but it's monkey-patched to throw a `ModbusException` instead.  This is caught normally.

I'm assuming that in (1) and (2) we should actually throw the `TypeError` to the caller, so I changed as such.  I see no valid use case for decoding `None`...

(3)  `test_processincomingpacket_not_ok()` calls `ClientDecoder.decode(b'\x03\x01\x00\n')` and expects a `ModbusIOException`.  The real exception is a `struct.error` because the payload has only 3 bytes.  The register values must be an even number of bytes i.e. 6 bytes for 3 registers.
image

This bad message is caught by `decode()`, losing information on the error.  Later on `frameProcessIncomingPacket` notices that `decode` returned None and `raise ModbusIOException`.

I think it's better to explicitly raise a descriptive exception in this case.

### DIFF
--- a/pymodbus/factory.py
+++ b/pymodbus/factory.py
@@ -236,8 +236,6 @@ class ClientDecoder:
             return self._helper(message)
         except ModbusException as exc:
             Log.error("Unable to decode response {}", exc)
-        except Exception as exc:  # pylint: disable=broad-except
-            Log.error("General exception: {}", exc)
         return None
 
     def _helper(self, data: str):
--- a/pymodbus/pdu/register_read_message.py
+++ b/pymodbus/pdu/register_read_message.py
@@ -4,6 +4,7 @@
 # pylint: disable=missing-type-doc
 import struct
 
+from pymodbus.exceptions import ModbusIOException
 from pymodbus.pdu import ModbusExceptions as merror
 from pymodbus.pdu import ModbusRequest, ModbusResponse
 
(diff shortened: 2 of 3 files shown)


alexrudd2:
Regarding (3), I'm not sure if that's actually an invalid value or not.

https://github.com/pymodbus-dev/pymodbus/blob/d32947517087214d28d71cdcbf3c02355526ab68/test/framers/test_old_framers.py#L499-L512

RapidScada's [parser](https://rapidscada.net/modbus/) claims it's valid.... but I don't trust it.  It's parsing the last two bytes twice (once as data and once as CRC!)
image


Anyways, this is all in `test_old_framer` so maybe it's going to get deleted... anyways I got distracted hunting down https://github.com/pymodbus-dev/pymodbus/pull/2261

PR #2261, opened 2024-07-23
Cleanup ReadWriteMultipleRegistersResponse and testing

alexrudd2:
There is no reason `ReadWriteMultipleRegistersResponse` can't inherit from `ReadHoldingRegistersResponse`.  The sole differences are:
(1) the function code (OK)
(2) `decode()` not initializing `self.registers`

https://github.com/pymodbus-dev/pymodbus/blob/d32947517087214d28d71cdcbf3c02355526ab68/pymodbus/pdu/register_read_message.py#L385-L392

This is a bug - if for some dumb reason `decode()` is called twice, it will keep appending registers to `self.registers`.  As far as I can tell, this bug (and corresponding incorrect test code) has been there since https://github.com/pymodbus-dev/pymodbus/commit/b02f5fe0761010606075943eced6fbffd22ce3d4  (13 years!)

Changing this also makes it apparent that the test code can be readily simplified.


janiversen (approved):
Wow, nice cleanup!

PR #2262, opened 2024-07-24
buildPacket can be used for Request and Response

alexrudd2:
Fixes a small oversight from https://github.com/pymodbus-dev/pymodbus/pull/2232
﻿


janiversen (approved):
We are all humans, so making small mistakes is natural.

PR #2282, opened 2024-08-04
Add socket arg to AsyncModbusTcpClient

dsblendo:
@janiversen Let me know if this is a reasonable request. I would like to pass an already created socket directly into `AsyncModbusTcpClient`, to pass it directly into `loop.create_connection()`.

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -50,6 +50,7 @@ from __future__ import annotations
 
 import asyncio
 import dataclasses
+import socket
 import ssl
 from abc import abstractmethod
 from collections.abc import Callable, Coroutine
@@ -87,6 +88,7 @@ class CommParams:
     host: str = "localhost" # On some machines this will now be ::1
     port: int = 0
     source_address: tuple[str, int] | None = None
+    socket: socket.socket | None = None
     handle_local_echo: bool = False
 
     # tls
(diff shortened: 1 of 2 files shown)


pymodbus/transport/transport.py
```diff
@@ -87,6 +88,7 @@ class CommParams:
     host: str = "localhost" # On some machines this will now be ::1
     port: int = 0
     source_address: tuple[str, int] | None = None
+    socket: socket.socket | None = None
```
janiversen, line 91: This worries me quite a bit, because comm_params was easy printable parameters.

pymodbus/transport/transport.py
```diff
@@ -231,6 +233,7 @@ def init_setup_connect_listen(self, host: str, port: int) -> None:
                 self.handle_new_connection,
                 host,
                 port,
+                sock=self.comm_params.socket,
```
janiversen, line 236: This affects both the server (listen) and client (connect), but you have not changed the server part.

This really needs to be tested, what happens if you pass none.

janiversen (changes requested):
I have to think more about this change, for a couple of reasons:
- how to handle reconnect (could reconnect to another server)
- how to interpret host/port parameters, socket comes connected
- How to handle the server (transport is shared)
- this needs new tests cases in transport to ensure there are no side effects.

With the questions properly solved I do not objections.

dsblendo:
@janiversen thanks for the feedback. I will review the lift required to make these changes and meet your requirements to see if it is worthwhile.

PR #2286, opened 2024-08-08
Fix example docstrings for PDU changes

alexrudd2:
The docstrings were missed in https://github.com/pymodbus-dev/pymodbus/pull/2160


janiversen (approved):

janiversen:
I really wonder why the tests did not catch this, I will check that tomorrow.

PR #2288, opened 2024-08-08
Clean up ModbusControlBlock

alexrudd2:
Use `_` instead of `__` for private attributes.  Name mangling is not required and this makes `pylint` and `mypy` happier.

Also, use properties as intended instead of calling a setter.


janiversen:
Wow, you just touched the oldest untouched code.

I have no idea why it was done with __.

PR #2290, opened 2024-08-09
Simplify framer test setup

alexrudd2:
`DummyFramer` can be entirely eliminated with `@mock.patch`

Also `getattr()` can return a default value if the attribute doesn't exist.

Note: an alternative is `@mock.patch.object(Framer, "__abstractmethods__", set())` but I like `@mock.patch.multiple`'s keyword argument a bit better.


janiversen (approved):
Nice reduction of code, thanks.

PR #2291, opened 2024-08-11
Update repl requirement to >= 2.0.4

dhoomakethu:
* Update repl requirements to be in sync with pymodbus 3.7.x

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ serial = [
     "pyserial>=3.5"
 ]
 repl = [
-   "pymodbus-repl>=2.0.3"
+   "pymodbus-repl>=2.0.4"
 ]
 
 simulator = [


pyproject.toml
```diff
@@ -47,7 +47,7 @@ serial = [
     "pyserial>=3.5"
 ]
 repl = [
-   "pymodbus-repl>=2.0.3"
+   "pymodbus-repl>=2.0.4"
```
janiversen, line 50: Seems you did not release that version.

janiversen (changes requested):

pyproject.toml
alexrudd2: It has now been released on [PyPI](https://pypi.org/project/pymodbus-repl/2.0.4/)

janiversen (approved):

PR #2292, opened 2024-08-12
Properly process 'slaves' argument, add explicit 'zero_mode' argument

Cavaler:
- 'slaves' argument is not treated as a boolean anymore, but really sets the number of slaves, as its description suggested
- added argument to explicitly set zero_mode, can be used with single slave as well

### DIFF
--- a/examples/helper.py
+++ b/examples/helper.py
@@ -76,7 +76,6 @@ def get_commandline(server=False, description=None, extras=None, cmdline=None):
             help="set number of slaves, default is 0 (any)",
             default=0,
             type=int,
-            nargs="+",
         )
         parser.add_argument(
             "--context",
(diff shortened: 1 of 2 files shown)


examples/helper.py
```diff
@@ -76,7 +76,13 @@ def get_commandline(server=False, description=None, extras=None, cmdline=None):
             help="set number of slaves, default is 0 (any)",
             default=0,
             type=int,
-            nargs="+",
+        )
+        parser.add_argument(
+            "--zero-mode",
```
janiversen, line 81: Zero mode is no longer a supported parameter.

janiversen (changes requested):
please remove zero_mode as the lower layers no longer support this.

examples/helper.py
janiversen: That is to say it is in the process of being removed.

Cavaler: Removed

janiversen (commented):
LGTM, thanks.

PR #2309, opened 2024-09-07
Fix type hints for write_register and write_registers

### LINKED ISSUE #2308: Type Hint for write_registers value parameter wrong?
`BinaryPayloadBuilder.build(…)` returns `list[bytes]`
```python
def build(self) -> list[bytes]:
```
but `ModbusTcpClient.write_registers(…)` expects `list[int]` as values argument
```python
def write_registers(self, address: int, values: list[int], slave: int = 1, skip_encode: bool = False) -> T:
``` 

According to my tests, write_registers however actually expects `list[bytes]` so I think this type hint there is wrong? Can anybody confirm or deny that?

rauberdaniel:
Fixes for type hints on `write_register` and `write_registers` which actually expect `bytes` / `list[bytes]` instead of `int` / `list[int]`

Closes #2308

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -111,7 +111,7 @@ class ModbusClientMixin(Generic[T]):  # pylint: disable=too-many-public-methods
         """
         return self.execute(pdu_bit_write.WriteSingleCoilRequest(address, value, slave=slave))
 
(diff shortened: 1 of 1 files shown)


janiversen (commented):
Lgtm, thanks

janiversen (approved):

PR #2310, opened 2024-09-08
Update BinaryPayloadDecoder._unpack_words documentation

MrWaloo:
There is no parameter "fstring" for this function

### DIFF
--- a/pymodbus/payload.py
+++ b/pymodbus/payload.py
@@ -330,7 +330,7 @@ class BinaryPayloadDecoder:
         # Change Word order if little endian word order  #
         # Pack values back based on correct byte order   #
         # ---------------------------------------------- #
-        :param fstring:
+
         :param handle: Value to be unpacked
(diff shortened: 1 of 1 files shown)


pymodbus/payload.py
```diff
@@ -330,7 +330,7 @@ def _unpack_words(self, handle) -> bytes:
         # Change Word order if little endian word order  #
         # Pack values back based on correct byte order   #
         # ---------------------------------------------- #
-        :param fstring:
+
```
janiversen, line 333: Good catch, except this is an internal function, where we tend not to document the parameters.

janiversen (approved):

janiversen:
Solve in #2312, since you have not allowed maintainers to update your PR.

I have kept you as author.

Thanks for the PR.

MrWaloo:
Sorry, I did not check the options

PR #2318, opened 2024-09-12
Improvements for example/contib/solar

MrWaloo:
The timeout value has been set to 5 seconds and now, there are no errors anymore
Rounding improvement
Some code visualization improvements too
A faulty register address (37114 instead of 32114)

Sorry for all that (it was late, but it's not an excuse)
At least this code works now


janiversen (approved):
LGTM, thanks

PR #2319, opened 2024-09-14
Fix encoding & decoding of ReadFileRecordResponse

justinsg:
Encode fix:
```
The sub-req file response length is in bytes, not registers, and needs
to also count the reference type byte. We already calculate the correct
value in `FileRecord.response_length` so make use of it here.
```

Decode fix:
```
The `count` increment was incorrect, because `response_length` already
includes the 1 byte for reference type. This moved the pointer too far
ahead, leading to incorrect slicing of record_data.

Also corrected the unit test for this; it seems to be copied from the
Modbus spec example but had a typo (repeated \x05) and did not assert
correctness of the second sub-req in the PDU.
```


janiversen (commented):
LGTM, thanks

PR #2327, opened 2024-09-20
Fixes the unexpected implementation of the ModbusSerialClient.connected property

MAKOMO:
Fixes the implementation of the connected property in sync client to just check if socket exists without calling connect. See Discussion #2325 and Discussion #1854.

Note that this `connected` property [is not used by the pymodbus lib currently](https://github.com/search?q=repo%3Apymodbus-dev%2Fpymodbus%20connected&type=code).


janiversen (commented):
LGTM, thanks

janiversen (approved):
LGTM, thanks

PR #2345, opened 2024-09-29
allow timeout to slaves on tcp clients

ahcm-dev:
I am connecting to a TCP to RTU modbus gateway which transparently passes on modbus requests to serial slaves on the downstream side. All works fine most of the time.

The application (home assistant) uses the AsyncModbusTCPClient to send periodic requests from a number of different threads to fetch data from different sensors on the different configured slaves.

If one slave malfunctions, or is powered off, a modbus request to it times out (though the gateway ack's the request at the TCP layer). This then causes the connection to be closed, and all the queued pending requests to the other slaves are failed. (Even though they would have succeeded if allowed to proceed).

While the connection quickly re-establishes again, the same thing happens as the polling function repolls the same unresponsive slave, and hence blocks successful activity from other healthy slaves.

I get good behaviour if in the tcp client we do not actually close the connection if the command to close with reconnect==True is sent. This relies on the tcp socket being relied upon to perform an unexpected close if the tcp timers time out.


ahcm-dev:
Error in pull request

janiversen:
You need to base the pull request on the newest dev.

PR #2346, opened 2024-09-30
Don't close/reopen tcp connection on single modbus message timeout

ahcm-dev:
I am connecting to a TCP to RTU modbus gateway which transparently passes on modbus requests to serial slaves on the downstream side. All works fine most of the time.

The application (home assistant) uses the AsyncModbusTCPClient to send periodic requests from a number of different threads to fetch data from different sensors on the different configured slaves.

If one slave malfunctions, or is powered off, a modbus request to it times out (though the gateway ack's the request at the TCP layer). This then causes the connection to be closed, and all the queued pending requests to the other slaves are failed. (Even though they would have succeeded if allowed to proceed).

While the connection quickly re-establishes again, the same thing happens as the polling function repolls the same unresponsive slave, and hence blocks successful activity from other healthy slaves.

I get good behaviour if in the tcp client we do not actually close the connection if the command to close with reconnect==True is sent. This relies on the tcp socket being relied upon to perform an unexpected close if the tcp timers time out.


pymodbus/client/tcp.py
```diff
@@ -86,8 +86,16 @@ def __init__(  # pylint: disable=too-many-arguments
         )
 
     def close(self, reconnect: bool = False) -> None:
-        """Close connection."""
```
janiversen, line 89: This is wrong, this method is called by the app when it wants to close the connection.

Why would you prevent the app from closing the connection ? Your problem is down in the transport layer.

janiversen (changes requested):

pymodbus/client/tcp.py
ahcm-dev: Not sure how this can be resolved in the transport layer. The issue is the call to close() in async_execute() in client/base.py after too many retries of a particular request. The remote end will never respond to this request as it addresses an inactive slave. It will however respond to other requests, and closing the connection as a result causes these to fail too.

janiversen: if the issue if the call to close in async_execute, then that would be a better place to fix the problem rather than prohibiting the app to close a connection.

Be aware your PR highlighted some old code, that should have been updated long time ago, it will be so very shortly.

janiversen:
??? I thought you had a problem, that you tried to solve ?

Your solution, as I pointed out, had serious side-effects and thus was not viable...but a working solution would be reviewed positively.

ahcm-dev:
Apologies trying to redo some changes.

ahcm-dev:
Closed in error - trying to reopen

ahcm-dev:
Pull request now updated

PR #2354, opened 2024-10-04
ModbusSlaveContext, remove zero_mode parameter.

janiversen:

### DIFF
--- a/API_changes.rst
+++ b/API_changes.rst
@@ -4,8 +4,7 @@ Versions (X.Y.Z) where Z > 0 e.g. 3.0.1 do NOT have API changes!
 
 API changes 3.8.0
 -----------------
-
-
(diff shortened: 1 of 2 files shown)


smurfix:
*Sigh* what's the rationale for this change?

My codebase contains a data store which pretty much depends on accesses to it being zero-based, and now I need to refactor that rather intrusively …

janiversen:
I do not think there needs to be a rationale to be discussed publicly !

Why do you need to "refactor that rather intrusively", you make a new class that inherits from ModbusSlaveContext, and does whatever you need done.

Anyhow ModbusSlaveContext is really very old with a number of nasty problems, so it is in the process of being replaced by a datastore that is more flexible and powerful.

PR #2357, opened 2024-10-08
Readme file renamed

theyashjani:
renamed the readme.rst => README.rst


janiversen (commented):
you need to update the reference to readme.rst as well, without that documents cannot be generated

test locally:
```
cd doc
./build_html
```

janiversen (changes requested):
Please ensure document generation works.

janiversen (approved):
hmmm sphinx accept lower case filenames, so it actually works.

janiversen:
Thanks, it will be available automatically on readthedocs in a couple of minutes (only dev !!!!)

theyashjani:
> you need to update the reference to readme.rst as well, without that documents cannot be generated
> 
> test locally:
> 
> ```
> cd doc
> ./build_html
> ```

understood

theyashjani:
> Thanks, it will be available automatically on readthedocs in a couple of minutes (only dev !!!!)

sure sir

PR #2366, opened 2024-10-09
Test on Python 3.13

alexrudd2:


alexrudd2:
The problem is apparently pytest-profiling, which hasn't bothered to release since 2019... but they are apparently resurrecting.

janiversen (approved):
LGTM, thanks.

janiversen:
Have not seen you for quite a while, nice to see your PR´s again. 

If you feel in mypy/ruff mode, then both framer and serial, should be totally typed. Next up for me is pdu and then transaction.

You missed to change the README, to tell others that we now test on python 3.13 (Thanks for doing that!!).

janiversen:
Your PR is always welcome !

Issue #2378, opened 2024-10-14
Feedback about the removal of `broadcast_enable` in 3.7.0+


matesh:
### Versions

- Python: 3.11
- OS: Yocto Linux
- Pymodbus: 3.7.x
- Modbus Hardware (if used):

### Pymodbus Specific

- Client: tcp/rtu - sync

### Description

The changes in the API introduced in 3.7.0 regarding broadcasting are in my opinion problematic. The below is some feedback of my experience with hundreds of different devices and complex systems using modbus TCP/RTU in the energy sector.

The change assumes that messages sent to slave==0 are intended to be broadcast, however, this is not always the case. I deal with several devices that act as a gateway into a system of multiple devices, which have slave==0 by default for their internal operational data, and all other unit IDs are devices addressed behind this gateway device. Previously, using the broadcast_enable flag upon instantiation of the client, it could be clearly defined how the client has to behave when slave==0: 

- When disabled, it was handled as if it was a standard device on the bus, expecting a response to any requests. 
- When enabled, requests were handled as a broadcast message, not expecting a response. 

The behaviour is now solely determined by the slave ID which results in undesired behaviour in case when responses are indeed expected from slave==0. When sending a read request to slave==0, the client doesn't take timeout in consideration, not expecting a response and thus requests fail, even though these requests are legitimate.

The assumption made in the documentation, that "experience shows that modern devices do not allow broadcast, mostly because it is inheriently dangerous" is in my opinion incorrect. In industrial control, broadcasting is rather important. We need to write setpoints in the same registers of multiple identical devices on the same RS485 bus (or RS485 bus behind a TCP/RTU gateway), and writing registers individually into each device one by one takes too much time. Broadcasting is thus important in terms of timely delivery of register writes simultaneously into all devices on the same bus.

The assumption "With slave=0 the application can get upto 254 responses on a single request" is in my opinion incorrect: Devices (even modern ones) with modbus implemented properly, ignore read commands on slave==0, only receive write commands (if they support broadcasting at all, otherwise they ignore those messages too), and regardless whether they accepted broadcast write commands or not, they do not respond/confirm the failure or success of any write commands, not even by means of exceptions. I have not yet worked with a device that responded to write broadcast commands, and definitely not read commands on slave==0, unless of course as described above, when they had a slave==0 on purpose, in a well established case.

[...]

PR #2382, opened 2024-10-15
Add `no_response_expected` argument to write requests

matesh:
Adds the `no_response_expected` argument to write requests, see discussion in #2378 

All requests will expect and wait for a response, exept write requests when the `no_response_expected=True` argument is set.

Assumptions:
1. When a read request is made to the broadcast address (slave==0), it is treated as a request to any other slave and at least one response is expected and waited for. If one wasn't expecing a response, a read request would not be made intentionally to the broadcast address
2. When a write request is made to the broadcast address, opposed to the specification, at least one response is expected and waited for.
3. If a response to a write request is not expected, the `no_response_expected=True` argument can be set when `write_coil`, `write_register`, `write_registers`, `write_file_record` and `mask_write_register` are called. This will result in the modbus message sent, then no valid response returned without delay.

### DIFF
--- a/doc/source/client.rst
+++ b/doc/source/client.rst
@@ -191,9 +191,9 @@ The logical devices represented by the device is addressed with the :mod:`slave=
 With **Serial**, the comm port is defined when creating the object.
 The physical devices are addressed with the :mod:`slave=` parameter.
 
-:mod:`slave=0` is used as broadcast in order to address all devices. To accommodate non-standard behaviour of devices, all request calls to :mod:`slave=0` will expect and wait for at least one response. If multiple devices on the bus are expected to respond to broadcast requests (for example for the purpose of device detection), the application can get upto 254 responses on a single request, and this is not handled with the normal API calls. If an application is expecting multiple responses to a broadcast request, it must call :mod:`client.execute` and deal with the responses.
+:mod:`slave=0` is used as broadcast (in accordance with the modbus standard). To accommodate non-standard behaviour of devices, all request calls to :mod:`slave=0` will expect and wait for one response. If multiple devices on the bus are expected to respond to broadcast requests (for example for the purpose of device detection), the application can get upto 254 responses on a single request, and this is not handled with the normal API calls. If an application is expecting multiple responses to a broadcast request, it must call :mod:`client.execute` and deal with the responses.
 
-If no response is expected to a broadcast or unicast write request, the :mod:`no_response_expected=True` argument can be used in the normal write API calls. However, these write operations will not return indication of failure or success, and will not respect the turnaround delay after sending a broadcast message, therefore, it is the application's responsibility to allow time for the slaves to process the broadcast request before the next request is made on the bus.
+If no response is expected to a request, the :mod:`no_response_expected=True` argument can be used in the normal API calls. However, these requests will not return indication of failure or success, and will not respect the turnaround delay after sending a broadcast message, therefore, it is the application's responsibility to allow time for the slaves to process the broadcast request before the next request is made on the bus.
 
 
 Client response handling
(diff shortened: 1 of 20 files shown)


doc/source/client.rst
```diff
@@ -191,14 +191,9 @@ The logical devices represented by the device is addressed with the :mod:`slave=
...
 The physical devices are addressed with the :mod:`slave=` parameter.
 
-:mod:`slave=0` is used as broadcast in order to address all devices.
-However experience shows that modern devices do not allow broadcast, mostly because it is
-inheriently dangerous. With :mod:`slave=0` the application can get upto 254 responses on a single request,
-and this is not handled with the normal API calls!
-
-The simple request calls (mixin) do NOT support broadcast, if an application wants to use broadcast
-it must call :mod:`client.execute` and deal with the responses.
+:mod:`slave=0` is used as broadcast in order to address all devices. To accommodate non-standard behaviour of devices, all request calls to :mod:`slave=0` will expect and wait for at least one response. If multiple devices on the bus are expected to respond to broadcast requests (for example for the purpose of device detection), the application can get upto 254 responses on a single request, and this is not handled with the normal API calls. If an application is expecting multiple responses to a broadcast request, it must call :mod:`client.execute` and deal with the responses.
```
janiversen, line 194: used as broadcast (in accordance with the modbus standard)

Not "at least" it is "wait for one response".

doc/source/client.rst
```diff
@@ -191,14 +191,9 @@ The logical devices represented by the device is addressed with the :mod:`slave=
...
 
+If no response is expected to a broadcast or unicast write request, the :mod:`no_response_expected=True` argument can be used in the normal write API calls. However, these write operations will not return indication of failure or success, and will not respect the turnaround delay after sending a broadcast message, therefore, it is the application's responsibility to allow time for the slaves to process the broadcast request before the next request is made on the bus.
```
janiversen, line 196: to a slave=0 request (remove unicast, that I think is not defined in Modbus).

We should add the parameter to all calls and not just write calls, that way the API is uniform, when the pdu is refactored, we might support read with slave=0 in the standard API.

pymodbus/client/base.py
```diff
@@ -108,6 +108,10 @@ async def async_execute(self, request) -> ModbusResponse:
             async with self._lock:
                 req = self.build_response(request)
                 self.ctx.send(packet)
+                no_response_expected = hasattr(request, "no_response_expected") and request.no_response_expected is True
```
janiversen, line 111: hasattr is relative costly, please add no_response_expected to ModbusPDU so that it is always available.

pymodbus/client/base.py
```diff
@@ -121,7 +125,7 @@ async def async_execute(self, request) -> ModbusResponse:
                 f"ERROR: No response received after {self.retries} retries"
             )
 
-        return resp
+        return resp  # type: ignore[return-value]
```
janiversen, line 128: This is wrong, the problem is the typing of the method which should include "| None".

janiversen (changes requested):
A couple of smaller things and the request to extend it to all calls.

Otherwise fine.

I will later make some test cases.

janiversen:
Please rebase the branch to the newest dev, as pdu have had recent changes.

doc/source/client.rst
matesh: Will update. Just to make sure I'm not getting informed from the wrong source, I have been referring to the specification laid out in this document: https://www.modbus.org/docs/Modbus_over_serial_line_V1_02.pdf
which is labelled "Please use this for new MODBUS implementations."

If this is no longer applicable or valid, please let me know

pymodbus/client/base.py
matesh: Certainly

pymodbus/client/base.py
matesh: Will fix

matesh:
I will open a new PR for this, something went wrong on my end, dev branch didn't sync properly and this all got messy.

Issue #2368, opened 2024-10-10
struct.error: required argument is not an integer


MAKOMO:
### Versions

- Python: 3.12
- OS: macOS 15
- Pymodbus: 3.7.3
- Modbus Hardware (if used):

### Description

The type of the Client methods `write_register` (and `write_registers`) was changed in `client/mixin.py` (#2309). Now the value to be sent should be of type `bytes` instead of `int` as in the previous releases.

Converting the int `value` to bytes using `value.to_bytes(2, 'big')` makes mypy happy, but results in

```
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/pymodbus/client/base.py", line 104, in async_execute
    packet = self.ctx.framer.buildPacket(request)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/pymodbus/framer/base.py", line 85, in buildPacket
    data = message.function_code.to_bytes(1,'big') + message.encode()
                                                     ^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/pymodbus/pdu/register_write_message.py", line 42, in encode
    packet += struct.pack(">H", self.value)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
struct.error: required argument is not an integer
```

Sending the int value as before works, but makes mypy sad.

What to do?

janiversen:
You should be able to send bytes or int, seems I did a bad review on #2309.

Fix will be available in a moment.

MAKOMO:
Thanks for your quick reaction. Your fix looks good to me.

MAKOMO:
Thanks for providing v3.7.4 of your great lib!

However, I am not completely happy with the new type annotation of the values in mixin:write_registers which is now `list[bytes | int]`.

As [`list` is invariant](https://github.com/python/mypy/issues/3351) it is now not type correct to provide this function with a homogeneous list of either bytes or int.

For example on a `test.py` with

```
from typing import Sequence

def f(l:list[int | bytes]) -> None:
    return
def g(l:list[int] | list[bytes]) -> None:
    return
def h(l:Sequence[int | bytes]) -> None:
    return

l:list[int] = [1,2,3]

f(l) # type error
g(l) # type correct
h(l) # type correct
```

[...]

PR #2411, opened 2024-10-24
Improve type of parameter values of write_registers

MAKOMO:
To use write_registers type correctly with inhomogeneous lists of type list[int] or list[bytes] as values, we promote the type of the parameter values from list to Sequence (see Issue #2368)

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -2,9 +2,9 @@
 from __future__ import annotations
 
 import struct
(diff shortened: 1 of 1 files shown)


janiversen (approved):
LGTM, thanks

janiversen:
FYI, in the main directory you will find check_ci.sh a small utility that runs like CI, but locally and thereby allow ruff and others to automatically change the code, instead of just suggesting it.

PR #2415, opened 2024-10-25
Fixed handle local echo in serialserver

jorisplusplus:
Fixes issue with the serial server not handling local echo of an rs485 adapter (see  #2414 )


janiversen (approved):
LGTM, thanks

PR #2435, opened 2024-11-04
Add typing to examples.

janiversen:
Add test example for write_registers, to ensure the different lists are legal.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,7 @@ jobs:
       - name: mypy
         if: matrix.run_lint == true
         run: |
-          mypy pymodbus
+          mypy pymodbus examples
 
       - name: ruff
         if: matrix.run_lint == true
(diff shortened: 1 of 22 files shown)


janiversen:
@MAKOMO I had to change the write_register typing, to make it consistent with the internals, but to ensure it worked I added a test case in client_async_calls.py: async_write_registers_mypy just to check that mypy did not throw it out.

MAKOMO:
> @MAKOMO I had to change the write_registers typing, to make it consistent with the internals, but to ensure it worked I added a test case in client_async_calls.py: async_write_registers_mypy just to check that mypy did not throw it out.

Looks good to me! Thanks for those tests. They will be helpful in the future.

janiversen:
If you want other typing tests, then please just add them to client_async_call.py, maybe rename write_registers_mypy to something like check_mypy.

The disadvantage for you is a bit work upfront, but the bigger advantage is that the typing is checked in every PR.

PR #2448, opened 2024-11-09
Enforce keyword only parameters.

janiversen:

### DIFF
--- a/API_changes.rst
+++ b/API_changes.rst
@@ -7,6 +7,7 @@ API changes 3.8.0
 - ModbusSlaveContext, remove zero_mode parameter.
 - Remove skip_encode parameter.
 - rename ModbusExceptions enums to legal constants.
+- enforce client keyword only parameters (positional not allowed).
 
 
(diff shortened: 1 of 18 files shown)


ilka-schulz:
@janiversen Thank you for implementing this requirement. I think it makes sense here to require keyword-only parameters. However, as this introduced a breaking change, I would be super happy if next time a breaking change will only be introduced with a major version bump. This would make it much easier for software based on `pymodbus` to become more stable and not break unexpectedly. Thank you very much for your development and maintenance work!

janiversen:
Please read the README, our releases follow x.y.z

where
z is plugin
y means you need to make minor adjustments 
x means major adjustments 

This is a pattern used by many projects, including big ones like HomeAssistant and LibreOffice.

If breaking changes meant upping x, would mean we would be at something like 200.1.1 and there would be no need for y. This would have the bad side effect that you would not know if we just changed a simple parameter or you need to rewrite your app.

PR #2456, opened 2024-11-15
Deprecate BinaryPayloadDecoder / BinaryPayloadBuilder.

janiversen:

### DIFF
--- a/pymodbus/payload.py
+++ b/pymodbus/payload.py
@@ -38,6 +38,15 @@ class BinaryPayloadBuilder:
         payload = builder.build()
     """
 
+    @classmethod
+    def deprecate(cls):
+        """Log warning."""
+        Log.warning(
+            "BinaryPayloadBuilder is deprecated and will be removed in v3.9.0 !\n"
+            "Please use \"client.convert_from_registers()\" or \"client.convert_to_registers\"\n"
+            "See documentation: \"https://pymodbus.readthedocs.io/en/latest/source/client.html#pymodbus.client.mixin.ModbusClientMixin.convert_from_registers\""
+        )
+
     def __init__(
         self, payload=None, byteorder=Endian.LITTLE, wordorder=Endian.BIG, repack=False
     ):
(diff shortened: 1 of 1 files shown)


andreapignotti:
The new functions proposed as alternative do not include the ByteOrder and WordOrder!

janiversen:
Byte order and word order are handled internally.

andreapignotti:
Can I ask you how the library handle this?
 I have to read/write from/to different devices that can use Litte or Big Endian for byteorder and wordorder (this is device specific)

janiversen:
Of course you can, look at the source code.

janiversen:
It does, it defines big endian for address, count and registers, the modbus protocol do not (sadly) know about 32 bit integers etc.

The new functions are made to support the "de facto" standard used in many but not all modbus devices.

But no need to discuss, nobody are forced to use the convert functions supplied, it's an option.

farzadpanahi:
@janiversen I think I understand your point basically saying `pymodbus` does not need to be in the business of converting bytes and such which totally makes sense. 

but if that is the intention why even have this half-baked method [convert_from_registers](https://github.com/pymodbus-dev/pymodbus/blob/f5ede3eb08dab4e2079e19bf96e4778d4dfa91dd/pymodbus/client/mixin.py#L697), that makes assumptions about endians?
 
If the idea is to keep `pymodbus` clean and out of this, isn't it better to remove those methods all together and do not provide the half-baked utility/helper functionality which I think can cause more problems in future for novice users?

unix people had it this way ->  _Write programs that do one thing and do it **well**._

janiversen:
it's because I and a large number of users need exactly that.

If you do no not want to use it, fine by me, it's an option not a demand. with my life long experience this is a proper solution for most professional devices.

Please remember byte order is actually defined by the standard as big endian (otherwise there would have been a lot of problems with fields like address/count and all bit requests). Word order is a de facto standard defined by the largest manufacturers like siemens and schneider (which uses big endian), so yes there are companies not following the de facto standard....but these companies also often deviate from the standard in other ways.

Issue #2459, opened 2024-11-19
ModbusClientMixin.write_registers typing could be loosened


vfazio:
Note: all code is untested except by unit tests.

### Pymodbus Specific

- Client

### Description

The typed interface for `ModbusClientMixin.write_registers` has flip-flopped a few times in recent releases, but it's still slightly wrong IMO.

The current typing, as of the time of writing, is:

```python
    def write_registers(
        self,
        address: int,
        values: list[bytes | int],
        *,
        slave: int = 1,
        no_response_expected: bool = False
    ) -> T:
```

The problem with this typing is that any code calling into this function that is explicitly typed as `values: Sequence[int]` or `values: list[int]` will fail type checks.

It seems the motivation to migrate this from `list[int]` to `list[int] | list[bytes]` to `list[int|bytes]` is to denote that the input can be a mixture of ints or bytes and that the interface will handle conversions as necessary.

Unfortunately, because `list` is mutable, it is invariant. Type checkers, like mypy, assume because it's typed as `list[int|byte]` that the function could _potentially_ mutate this list to contain a type that a caller, who has typed their interfaces to be `list[int]` to be unaware of and would be unable to handle.

These interfaces _should not_ mutate the value, so can and probably should loosen their types as much as possible... with a few tweak, it could be as generic as:

```python
values: Iterable[int|bytes]
```

The underlying code really only needs it to be iterable (via `enumerate`). The only exception is:

```python
self.count: int = count or len(registers)
```

Which would mean it needs to be `Container` or `Sequence` however I think this could be avoided by simply doing:

```python
self.count: int = count or len(self.registers)
```

### Code and Logs

```diff
diff --git a/examples/client_async_calls.py b/examples/client_async_calls.py
index d2c118e5..2c119e8d 100755
--- a/examples/client_async_calls.py
+++ b/examples/client_async_calls.py
@@ -163,13 +163,15 @@ async def async_write_registers_mypy(client):
     await client.write_registers(1, regs1, slave=SLAVE)
     rr = await client.read_holding_registers(1, count=len(regs1), slave=SLAVE)
     assert not rr.isError()  # test that call was OK
+    assert rr.registers is not regs1
     assert rr.registers == regs1
 
     regs2: list[bytes] = [b'\x01\x02', b'\x03\x04']
     await client.write_registers(1, regs2, slave=SLAVE)
     rr = await client.read_holding_registers(1, count=len(regs2), slave=SLAVE)
[...]

PR #2464, opened 2024-11-22
Do not mutate caller's registers in ModbusPDU

vfazio:
Previously, ModbusPDU would assign the `registers` argument to a member
variable and subsequently convert `bytes` to `int`.

The registers argument is typed as `Sequence`, implying that these
values are immutable. However, the assignment and subsequent conversion
could corrupt a caller's data as the conversion affects the original
object reference. It could also potentially raise an error if the
sequence did not allow assignment (tuple, frozenset).

Now, the sequence is iterated through and a new list is created with
converted values (if necessary), leaving the caller's data unchanged.

Also fix a case in the examples that incorrectly relied on this data corruption.

Because the examples are not fully typed, it's not clear to example
writers what types the class attributes actually are.

In this case, `ModbusPDU.registers: list[int]` was being compared for
equality to `regs2: list[bytes]`. This previously worked because
`ModbusPDU` took ownership of the caller's registers and mutated them.
The mutated values were then copied into the response class and compared
against the caller's corrupted values.

Since `registers` should be treated as immutable due to `ModbusPDU`'s
constructor typing the value as `Sequence`, the example has been updated
to compare the list of registers to their converted type (int).


vfazio:
related #2459

examples/client_async_calls.py
```diff
@@ -169,7 +169,7 @@ async def async_write_registers_mypy(client):
     await client.write_registers(1, regs2, slave=SLAVE)
     rr = await client.read_holding_registers(1, count=len(regs2), slave=SLAVE)
     assert not rr.isError()  # test that call was OK
-    assert rr.registers == regs2
+    assert rr.registers == [int.from_bytes(i, byteorder="big") for i in regs2]
```
janiversen, line 172: This seems wrong, registers in responses are always int never bytes.

pymodbus/pdu/pdu.py
```diff
@@ -36,11 +35,14 @@ def __init__(self,
...
-        self.registers: list[int] = cast(list[int], registers)
-        for i, value in enumerate(registers):
+        self.registers: list[int] = []
+        # copy values so original sequence is not mutated
+        for value in registers:
             if isinstance(value, bytes):
-                self.registers[i] = int.from_bytes(value, byteorder="big")
-        self.count: int = count or len(registers)
+                value = int.from_bytes(value, byteorder="big")
+            self.registers.append(value)
```
janiversen, line 43: This is quite a costly operation, which is now done with every entry.

The old code fast copied the entire array and only update entries that contained bytes, ensuring a lot faster execution for well behaving calls.

janiversen (changes requested):
The changes are not incorrect, just not efficient, which is a problem when you make thousands of calls.

janiversen:
Maybe a note in de parameter description, would solve the whole problem.

janiversen:
If you really want to ensure that there are no changes, then a simple .copy() would solve it, and keep the efficiency

examples/client_async_calls.py
vfazio: Because the test was broken when the values were copied "appropriately". After the call, `regs2` is no longer a list of bytes, it's a list of ints. The reason the comparison works is because of that mutation. if regs2 is not mutated by the call, it continues to be a list of bytes, which you cannot compare to a list of ints.

Your latest commit now changes how all of this works by trying to inform the caller that their data will potentially be corrupted by this call. If they do not want that, they will have to send a copy of their data every time they make an API call which sounds expensive

pymodbus/pdu/pdu.py
vfazio: It's not as costly as it looks. `Sequence` does not have a `copy` method, so that was not an option. I could have used a slice `[:]` but since we have to iterate over the entire list to convert the bytes, we may as well only iterate _once_, An explicit copy would iterate once to get the elements and then a second time to convert the bytes.

I figured, at least this way there's only one iteration. When i get back to the office i could certainly do performance testing to see which is the quickest option.

However, the typing has now changed again so this entire MR is kind of pointless. It only mattered because you typed this as `Sequence` which implies immutability.

I do think, however, that if you were really concerned about performance, there would not be a loop inside this function to iterate through the entire list and convert bytes to ints. Since ints are the preferred type, and to keep the code strict and performant, it would seem like you'd want to make this strictly a `list[int]` (or, IMO, a `Sequence[int]`) so you're not responsible for the conversion (or suffer the penalty of looping over data to convert when there's no need to convert it) and put that on the callers of the API.

Since you're moving to 3.8.x and can allow API changes, this seems like something you could do here, and then retype all of your interfaces to take a `list[int]` and maybe for convenience, provide a helper function for users that used to send `list[bytes]` to convert their list to a `list[int]`.

janiversen:
Please point me to the documentation that tells Sequence may not be modified, I have not found it.

vfazio:
> Please point me to the documentation that tells Sequence may not be modified, I have not found it.

Sure, it's not straightforward.

https://docs.python.org/3/library/stdtypes.html#immutable-sequence-types
https://docs.python.org/3/glossary.html#term-sequence
https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence
https://github.com/python/cpython/blob/main/Lib/_collections_abc.py#L995

The type system that exists, since it's not a core part of the language, is... "tricky".

A `Sequence` is not a `list` but a `list` is a `Sequence`.

A `Sequence` as defined in `collections.abc.Sequence` defines a minimal interface for any sequence, like a tuple, a frozenset, etc

`list` is also a `MutableSequence` which is distinct and allows mutation.

https://github.com/python/cpython/blob/main/Lib/_collections_abc.py#L1131

So, saying you accept a `Sequence` implies that you can take an immutable sequence of values and use only the methods defined on all `Sequence` types.

Typeshed can sometimes help here:

https://github.com/python/typeshed/blob/main/stdlib/typing.pyi#L579

So, it has `__len__` and `__contains__` but not `copy` or any method by which to index and assign values like `__setitem__` or `append` or `clear`.

vfazio:
It would seem to me that it's possible that callers that are communicating with devices may well have fixed messages or fixed register values that they send.. like constants. The closest thing to a const array in Python would be immutable sequences via `tuple` or `frozenset` so that they can guarantee that those values aren't modified either by their own code or downstream APIs.

However in most cases this is fine, APIs don't need to be able to change values, they need to be able to get the length of the array or iterate through the values. So just because one caller uses `frozenset` and someone else uses `list`, from the API's perspective, it can take a `Sequence` and work with both types.

The "challenge" however is if you do need to modify the array of items _internally_, in which case it may make sense to copy the elements so you can perform your mutation, so, either creating a list from the sequence via `copy = list(sequence)` or whatever method you want to put it in the datastructure you prefer seems reasonable.

janiversen:
I see what you write, and my solution to this is simple, I should never have allowed list[bytes] that is essentially a problem for the app to handle.

PR follow soon.

vfazio:
This is why it's also not safe to do: `cast(list[int], sequence)` because cast doesn't actually morph the type or look into a different vtable or whatever. it just informs the type checker to now assume that going forward the variable is a list, even if it is actually a tuple.

The easiest way to prove this out is to construct a `ModbusPDU` and pass in `registers = (bytes(2),)`. I expect it will blow up when trying to modify the byte value to an int because you cannot assign into a tuple.

vfazio:
> Obsolete with #2469, that secures the array is not modified (even though allowed as it is typed list[int]).

Thanks for your work on this and sorry again about earlier. I really just wanted to make sure this API became stable.

If there's anything I can help out with, please lmk.

janiversen:
Pull requests are always welcome, and e.g. there are still a lot of typing and testing  to be done.

We are currently refactoring the whole library and so far: transport, pdu, framer and transaction are done, with typing and 100% test coverage. But e.g. datastore and server are a mess.

PR #2493, opened 2024-12-09
Remove all trailing zeroes during string decoding

wlcrs:
The current string decoding logic removes only a single trailing zero from the input it receives. In practice, multiple trailing zeroes can exist.

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -708,8 +708,12 @@ class ModbusClientMixin(Generic[T]):  # pylint: disable=too-many-public-methods
         for x in registers:
             byte_list.extend(int.to_bytes(x, 2, "big"))
         if data_type == cls.DATATYPE.STRING:
-            while byte_list[-1:] == b"\00":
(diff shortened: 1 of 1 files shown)


pymodbus/client/mixin.py
```diff
@@ -708,8 +708,12 @@ def convert_from_registers(
         for x in registers:
             byte_list.extend(int.to_bytes(x, 2, "big"))
         if data_type == cls.DATATYPE.STRING:
-            if byte_list[-1:] == b"\00":
-                byte_list = byte_list[:-1]
+            # remove trailing null bytes
+            trailing_nulls_begin = len(byte_list)
+            while trailing_nulls_begin > 0 and byte_list[trailing_nulls_begin - 1] == 0:
```
janiversen, line 713: You compare an int with a byte, this forces a conversion, please change to compare to a byte.

janiversen (changes requested):

janiversen (changes requested):
looks good except for CI error.

PR #2497, opened 2024-12-11
Fix struct.unpack when decode data is longer than data types

pjkundert:
o This is triggered in situations where we are decoding responses
  from other hosts (eg. RS485 multi-drop).
o It only worked for decoding requests, because the supplied data
  happened to not contain any extra bytes.

The decode will raise an exception because of too much data, instead of decoding the supplied data.

### DIFF
--- a/pymodbus/pdu/bit_message.py
+++ b/pymodbus/pdu/bit_message.py
@@ -23,7 +23,7 @@ class ReadCoilsRequest(ModbusPDU):
 
     def decode(self, data: bytes) -> None:
         """Decode a request pdu."""
-        self.address, self.count = struct.unpack(">HH", data)
+        self.address, self.count = struct.unpack(">HH", data[:4])
 
     def get_response_pdu_size(self) -> int:
         """Get response pdu size.
@@ -84,7 +84,7 @@ class WriteSingleCoilResponse(ModbusPDU):
 
     def decode(self, data: bytes) -> None:
         """Decode a write coil request."""
-        self.address, value = struct.unpack(">HH", data)
+        self.address, value = struct.unpack(">HH", data[:4])
         self.bits = [bool(value)]
 
 
@@ -123,7 +123,7 @@ class WriteMultipleCoilsRequest(ModbusPDU):
 
     def decode(self, data: bytes) -> None:
         """Decode a write coils request."""
-        self.address, count, _byte_count = struct.unpack(">HHB", data[0:5])
+        self.address, count, _byte_count = struct.unpack(">HHB", data[:5])
         self.bits = unpack_bitstring(data[5:])[:count]
 
     async def update_datastore(self, context: ModbusSlaveContext) -> ModbusPDU:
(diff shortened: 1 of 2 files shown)


pymodbus/pdu/bit_message.py
```diff
@@ -23,7 +23,7 @@ def encode(self) -> bytes:
 
     def decode(self, data: bytes) -> None:
         """Decode a request pdu."""
-        self.address, self.count = struct.unpack(">HH", data)
```
janiversen, line 26: According to the modbus standard, a read_coil request contains address and count, nothing more.

pymodbus support the standard, not non-standard messages !

if you receive a non-standard message (which happens in real life), then please modify it in the trace_packet hook and pass a legal packet.

janiversen (changes requested):
Be aware that the pymodbus server do not work on a multi drop line.

We have tried it a couple of times, but the problem have always been to detect the pause between packets (3.5 characters), which are non-detectable due to device drivers/usb devices handling of the pause.

If you have an alternate proper solution to divide frames on a multi-drop line, then the correct place to change that is in framer/rtu.py...due to the fact that many responses due not have a proper length field we have failed so far.

The core of pymodbus support the standard modbus as defined, we have seen many requests to support non-standard, actually often conflicting requests (like one that receives to few bytes and yours that receive too many). We simply cannot maintain all these deviations.

However we recognize the need to handle these messages, and that the biggest reason why trace_packet was implemented....with trace_packet you receive the whole byte stream (for recv and send), and can make any changes you want before returning the modified stream.

janiversen:
Did you consider a reason why you get extra data on a multidrop line, is because the server expects a request but actually receives a response sent from another device.

In order to use the server on a multidrop line, the server needs to monitor and remember the requests to other devices in order to dismiss the responses. This is quite complicated, the other approach is to dismiss frames with an unknown dev_id.

Titoflegit:
"Be aware that the pymodbus server do not work on a multi drop line"
Nice to see this stated here... digging digging after some troubles
As it is an obvious limitation of using Pymodbus server it should have been written somewhere in the documentation!
First I saw pymodbus replying to slaves not defined in the datastore, even corrupting the response from the slave
Then I found the "ignore_missing_slaves" option, but effectively when set to True, the server get lost trying to decode the other slaves answers... and as it retry, data accumulate and the next master frame is added to the buffer...
So what is the solution ? I will try to emulate the other slaves in the datastore, and have a second serial line to read the other slaves... And I will not use Pymodbus client to do it for sure !

janiversen:
ignore_slaves is documented, and is needed if you have a multidrop line, so that the server do not respond with "unknown slave".

It is not stated anywhere that the pymodbus work on a multidrop line, very few servers do so (unless they have special rs485 converters). It is stated that the pymodbus server is for testing purposes and not production. It is quite hard to document what a product do not support, in fact its impossible, so we do the second best, and document what is supported.

If something is not documented 
, you could use "discussions" to ask...just a small hint.

As I wrote earlier, the server gets request/responses to other slaves, and get lost in the responses due to wrong size calculations.

I hope I explained why python and no special hardware, makes it quite impossible to make a server that runs stable on a multidrop line...I really hope you show that I am wrong by making a couple of pull requests, because it is surely a feature that would be nice to have.

I am not sure why you do not want to use the pymodbus client, but that is clearly your choice.

janiversen:
Closing, because seems to be a misunderstand.

PR #2501, opened 2024-12-13
Bugfix: TCP auto-reconnect

alexis-care:
As ask in the [Discussion](https://github.com/pymodbus-dev/pymodbus/discussions/2499), here is a workaround.


janiversen (commented):
Lgtm, thanks.

PR #2502, opened 2024-12-13
Raise exception on no response in async client.

### LINKED ISSUE #2500: Async client functions return `None` on no response.
### Versions
- Pymodbus: 3.8.0

### Description

Previously, a timeout during a call to e.g. `client.read_coils` would log an error and return an exception response. As of 3.8.0, this instead returns `None` in the async client, which breaks the normal usage of PyModbus as given in the documentation:
```python
try:
    rr = await client.read_coils(1, 1, slave=1)
except ModbusException as exc:
    _logger.error(f"ERROR: exception in pymodbus {exc}")
    raise exc
if rr.isError():  # Error: NoneType has attribute 'isError'.
    _logger.error("ERROR: pymodbus returned an error!")
    raise ModbusException(txt)
```

This can be seen by comparing `TransactionManager`'s [`sync_execute`](https://github.com/pymodbus-dev/pymodbus/blob/efb02eb3a01dc623a205da1367e8934b7788e3ff/pymodbus/transaction/transaction.py#L120C1-L122C41):
```python
txt = f"No response received after {self.retries} retries, continue with next request"
Log.error(txt)
raise ModbusIOException(txt)
```
with [`execute`](https://github.com/pymodbus-dev/pymodbus/blob/efb02eb3a01dc623a205da1367e8934b7788e3ff/pymodbus/transaction/transaction.py#L156C1-L159C1):
```python
Log.error(f"No response received after {self.retries} retries, continue with next request")
self.response_future = asyncio.Future()
return None
```

philj56:
This makes the async client behave the same as the sync client again.

Fixes #2500.

### DIFF
--- a/pymodbus/transaction/transaction.py
+++ b/pymodbus/transaction/transaction.py
@@ -151,9 +151,10 @@ class TransactionManager(ModbusProtocol):
                     f"ERROR: No response received of the last {self.accept_no_response_limit} request, CLOSING CONNECTION."
                 )
             self.count_no_responses += 1
-            Log.error(f"No response received after {self.retries} retries, continue with next request")
             self.response_future = asyncio.Future()
-            return None
+            txt = f"No response received after {self.retries} retries, continue with next request"
+            Log.error(txt)
+            raise ModbusIOException(txt)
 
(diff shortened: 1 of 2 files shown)


janiversen (approved):
Lgtm, thanks.

PR #2506, opened 2024-12-17
Convert endianness

daanwtb:

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import struct
 from abc import abstractmethod
 from enum import Enum
-from typing import Generic, TypeVar, Literal
+from typing import Generic, Literal, TypeVar
 
 import pymodbus.pdu.bit_message as pdu_bit
 import pymodbus.pdu.diag_message as pdu_diag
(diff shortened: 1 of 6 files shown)


pymodbus/client/mixin.py
```diff
@@ -695,18 +695,19 @@ class DATATYPE(Enum):
...
     @classmethod
     def convert_from_registers(
-        cls, registers: list[int], data_type: DATATYPE
+        cls, registers: list[int], data_type: DATATYPE, endian: Literal["big", "little"] = "big"
     ) -> int | float | str | list[bool]:
         """Convert registers to int/float/str.
 
         :param registers: list of registers received from e.g. read_holding_registers()
         :param data_type: data type to convert to
+        :param endian: endianness of the byte order
```
janiversen, line 704: This is wrong, byte order is always big.

What is not defined by the standard is the word order.

pymodbus/client/mixin.py
```diff
@@ -695,18 +695,19 @@ class DATATYPE(Enum):
...
         :returns: int, float, str or list[bool] depending on "data_type"
         :raises ModbusException: when size of registers is not 1, 2 or 4
         """
         byte_list = bytearray()
         for x in registers:
-            byte_list.extend(int.to_bytes(x, 2, "big"))
+            byte_list.extend(int.to_bytes(x, 2, endian))
```
janiversen, line 710: this is wrong, a register is always "big".

pymodbus/client/mixin.py
```diff
@@ -725,12 +726,13 @@ def convert_from_registers(
 
     @classmethod
     def convert_to_registers(
-        cls, value: int | float | str | list[bool], data_type: DATATYPE
+        cls, value: int | float | str | list[bool], data_type: DATATYPE, endian: Literal["big", "little"] = "big"
```
janiversen, line 729: see above.

pymodbus/client/mixin.py
```diff
@@ -725,12 +726,13 @@ def convert_from_registers(
...
     ) -> list[int]:
         """Convert int/float/str to registers (16/32/64 bit).
 
         :param value: value to be converted
         :param data_type: data type to be encoded as registers
+        :param endian: endianness to encode bytes to
```
janiversen, line 735: see above.

pymodbus/client/mixin.py
```diff
@@ -749,7 +751,7 @@ def convert_to_registers(
         else:
             byte_list = struct.pack(f">{data_type.value[0]}", value)
         regs = [
-            int.from_bytes(byte_list[x : x + 2], "big")
+            int.from_bytes(byte_list[x : x + 2], endian)
```
janiversen, line 754: see above.

janiversen (changes requested):
Please read the standard:

'''
4.2 Data Encoding
 MODBUS uses a ‘big-Endian’ representation for addresses and data items. This means
that when a numerical quantity larger than a single byte is transmitted, the most significant
byte is sent first. So for example
Register size value
16 - bits 0x1234 the first byte sent is 0x12 then 0x34
'''

Your pull request would make sense, for int32 etc.

daanwtb:
You are right about the modbus protocol stating each register should be big-endian, but we have encountered situations where it was not. I like the new implementation convert_from_registers/convert_to_registers but we can only use it if we can also use little endian byte_order. Proposing a new solution and like to hear your opinion.

janiversen:
Pymodbus implemented the modbus standard NOT all the non-standard implementations!

If you want a different byte order then please convert the bytes in the app.

pymodbus/client/mixin.py
```diff
@@ -695,18 +695,25 @@ class DATATYPE(Enum):
...
     def convert_from_registers(
-        cls, registers: list[int], data_type: DATATYPE
+        cls, registers: list[int], data_type: DATATYPE, byte_order: Literal["big", "little"] = "big",
+            word_order: Literal["big", "little"] = "big"
     ) -> int | float | str | list[bool]:
         """Convert registers to int/float/str.
 
         :param registers: list of registers received from e.g. read_holding_registers()
         :param data_type: data type to convert to
+        :param byte_order: Literal[big, small] order of to bytes in 16 bit register encoding (ALMOST always big)
```
janiversen, line 705: Byte order is fixed by the standard.

pymodbus/client/mixin.py
```diff
@@ -695,18 +695,25 @@ class DATATYPE(Enum):
...
+                                                                                                        (default is big)
         :returns: int, float, str or list[bool] depending on "data_type"
         :raises ModbusException: when size of registers is not 1, 2 or 4
         """
         byte_list = bytearray()
         for x in registers:
-            byte_list.extend(int.to_bytes(x, 2, "big"))
+            byte_list.extend(int.to_bytes(x, 2, byte_order))
+        if word_order == "little":
+            byte_list = byte_list[::-1]
```
janiversen, line 716: This does not convert the order of all registers.

janiversen (changes requested):
You miss test cases that shows these conversions actually works.

janiversen (approved):

PR #2545, opened 2025-01-23
Added keyword arguments to remote.py

hjkarssies:
Got the following error:
TypeError: ModbusClientMixin.read_holding_registers() takes 2 positional arguments but 3 positional arguments (and 1 keyword-only argument) were given

Fixed it by adding keywords to callbacks in __build_mapping(self)

### DIFF
--- a/pymodbus/datastore/remote.py
+++ b/pymodbus/datastore/remote.py
@@ -74,42 +74,42 @@ class RemoteSlaveContext(ModbusBaseSlaveContext):
             params["slave"] = self.slave
         self.__get_callbacks = {
             "d": lambda a, c: self._client.read_discrete_inputs(
-                a, c, **params
+                address=a, count=c, **params
             ),
             "c": lambda a, c: self._client.read_coils(
-                a, c, **params
+                address=a, count=c, **params
             ),
             "h": lambda a, c: self._client.read_holding_registers(
-                a, c, **params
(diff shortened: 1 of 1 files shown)


janiversen (approved):
LGTM, thanks

janiversen (changes requested):
Ups seems you broke our test harness, please fix that.

janiversen (changes requested):
I was really too fast, address and count are positional not optional parameters.

janiversen:
Are you using an older version of pymodbus but have the current examples, that would explain your problems, because earlier it accepted named parameters for positional parameters

hjkarssies:
I have pymodbus 3.8.3 installed, and I'm not running any of the examples. The error arises when I make a Modbus TCP to RTU gateway like this:

```
async def modbus_gateway(self):
    store = RemoteSlaveContextWithFix(self.rtu_client, self.rtu_slave)
    context = ModbusServerContext(slaves=store, single=True)
    await StartAsyncTcpServer(context=context, address=("", self.tcp_port))
```

About breaking the test harness, I have no idea what went wrong and how I can fix it to be honest, but I would be happy to take a look if you can point me in the right direction!

hjkarssies:
I see, in that case I'll just close the pull request. Thanks a lot for your your feedback!

PR #2550, opened 2025-01-25
Typo arround `no_response_expected`

MrWaloo:
Documentation typo and advanced modbus message call example.
No breaking change, only cosmetic.

### DIFF
--- a/pymodbus/client/mixin.py
+++ b/pymodbus/client/mixin.py
@@ -37,10 +37,10 @@ class ModbusClientMixin(Generic[T]):  # pylint: disable=too-many-public-methods
     Advanced modbus message call::
 
         request = ReadCoilsRequest(1,10)
-        response = client.execute(no_response_expected=False, request=request)
+        response = client.execute(False, request)
         # or
         request = ReadCoilsRequest(1,10)
-        response = await client.execute(no_response_expected=False, request=request)
+        response = await client.execute(False, request)
 
     .. tip::
         All methods can be used directly (synchronous) or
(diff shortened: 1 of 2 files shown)


pymodbus/client/mixin.py
```diff
@@ -37,10 +37,10 @@ class ModbusClientMixin(Generic[T]):  # pylint: disable=too-many-public-methods
     Advanced modbus message call::
 
         request = ReadCoilsRequest(1,10)
-        response = client.execute(request)
+        response = client.execute(no_response_expected=False, request=request)
```
janiversen, line 40: That is the default, so no need to add it.

pymodbus/client/mixin.py
```diff
@@ -37,10 +37,10 @@ class ModbusClientMixin(Generic[T]):  # pylint: disable=too-many-public-methods
...
         # or
         request = ReadCoilsRequest(1,10)
-        response = await client.execute(request)
+        response = await client.execute(no_response_expected=False, request=request)
```
janiversen, line 43: That is the default, so no need to add it.

janiversen (changes requested):
Looks good, apart from the added arguments, please correct.

MrWaloo:
Indeed, it is positional:
- mixin:
https://github.com/pymodbus-dev/pymodbus/blob/f20930de1a5d53a16dbfd9829c494359aac8fe2c/pymodbus/client/mixin.py#L54

- base sync:
https://github.com/pymodbus-dev/pymodbus/blob/f20930de1a5d53a16dbfd9829c494359aac8fe2c/pymodbus/client/base.py#L80

- base async:
https://github.com/pymodbus-dev/pymodbus/blob/f20930de1a5d53a16dbfd9829c494359aac8fe2c/pymodbus/client/base.py#L191

The order is correct in the PR and there is no default value (yet) in the existing code. So the correction is legitimate.
Unless I missed something or you want to change the existing code. But, I did not find anything pointing this fact in the roadmap.

janiversen (approved):
LGTM, thanks.

PR #2551, opened 2025-01-26
Fix exception text in ModbusPDU.validateAddress

MrWaloo:
According to the test, the text should be `0 < address {address} < 65535 !` instead of `9 < address {address} < 65535 !`.


janiversen (commented):
Good catch, thanks.

PR #2581, opened 2025-02-21
Flush recv_buffer before each transaction write.

jameshilliard:
Ensure we don't have any non-relevant data in the buffer.

Haven't tested yet but I think this is the expected behavior based on [this](https://github.com/pymodbus-dev/pymodbus/issues/2580#issuecomment-2675297181) comment.

### DIFF
--- a/pymodbus/transaction/transaction.py
+++ b/pymodbus/transaction/transaction.py
@@ -137,6 +137,7 @@ class TransactionManager(ModbusProtocol):
             request.transaction_id = self.getNextTID()
             count_retries = 0
             while count_retries <= self.retries:
+                self.recv_buffer = b""
                 self.response_future = asyncio.Future()
                 self.pdu_send(request)
                 if no_response_expected:


pymodbus/transaction/transaction.py
```diff
@@ -152,6 +152,7 @@ async def execute(self, no_response_expected: bool, request: ModbusPDU) -> Modbu
                         )
                     return response
                 except asyncio.exceptions.TimeoutError:
+                    self.recv_buffer = b""
```
janiversen, line 155: This is the wrong place, since it can cause conflicts with the active callback.

It would be better placed in line 140, which is before the callback is active.

janiversen (changes requested):

pymodbus/transaction/transaction.py
jameshilliard: Ah, yeah I guess purging the buffer before doing the write makes more sense.

janiversen (approved):

PR #2582, opened 2025-02-21
Don't set_result on completed futures.

### LINKED ISSUE #2580: asyncio.exceptions.InvalidStateError: invalid state
Seeing this invalid state error [here](https://github.com/pymodbus-dev/pymodbus/blob/v3.8.3/pymodbus/transaction/transaction.py#L183), guessing we need to check if the future already has a result set?:
```python
Exception response 254 / 0
Exception in callback SerialTransport.intern_read_ready()
handle: <Handle SerialTransport.intern_read_ready()>
  File "/usr/lib/python3.12/asyncio/events.py", line 88, in _run
  File "/usr/lib/python3.12/site-packages/pymodbus/transport/serialtransport.py", line 134, in intern_read_ready
  File "/usr/lib/python3.12/site-packages/pymodbus/transport/transport.py", line 304, in data_received
  File "/usr/lib/python3.12/site-packages/pymodbus/transport/transport.py", line 338, in datagram_received
  File "/usr/lib/python3.12/site-packages/pymodbus/transaction/transaction.py", line 183, in callback_data
asyncio.exceptions.InvalidStateError: invalid state
```

jameshilliard:
We need to guard self.response_future.set_result with a `done()` check to prevent `asyncio.exceptions.InvalidStateError: invalid state` errors.

Similar to #2581 I think we also need to flush the `recv_buffer` if the transaction is cancelled by the caller(say by the caller executing a transaction inside of another `asyncio.wait_for` that times out before the `asyncio.wait_for` inside of the `execute` call), in this case however we should re-raise the exception after the flush so that no retries are attempted.

Since `self.response_future` is cancelled in this case we should also guard the `self.response_future.set_result` call.

Should hopefully fix #2580 in combination with #2581.

fixes #2580

### DIFF
--- a/pymodbus/transaction/transaction.py
+++ b/pymodbus/transaction/transaction.py
@@ -195,7 +195,10 @@ class TransactionManager(ModbusProtocol):
                     raise ModbusIOException(
                         f"ERROR: request ask for id={self.request_dev_id} but got id={pdu.dev_id}, CLOSING CONNECTION."
                     )
-                self.response_future.set_result(self.last_pdu)
+                if self.response_future.done():
+                    Log.warning("received unexpected pdu: {}", pdu, ":str")
+                else:
+                    self.response_future.set_result(self.last_pdu)
         return used_len
 
     def getNextTID(self) -> int:
--- a/pymodbus/transaction/transaction.py
+++ b/pymodbus/transaction/transaction.py
@@ -154,6 +154,8 @@ class TransactionManager(ModbusProtocol):
                     return response
                 except asyncio.exceptions.TimeoutError:
                     count_retries += 1
+                except asyncio.exceptions.CancelledError as exc:
+                    raise ModbusIOException("Request cancelled outside pymodbus.") from exc
             if self.count_until_disconnect < 0:
                 self.connection_lost(asyncio.TimeoutError("Server not responding"))
                 raise ModbusIOException(
@@ -196,9 +198,8 @@ class TransactionManager(ModbusProtocol):
                         f"ERROR: request ask for id={self.request_dev_id} but got id={pdu.dev_id}, CLOSING CONNECTION."
                     )
                 if self.response_future.done():
-                    Log.warning("received unexpected pdu: {}", pdu, ":str")
-                else:
-                    self.response_future.set_result(self.last_pdu)
+                    raise ModbusIOException(f"received pdu: {pdu:str} without a corresponding request")
+                self.response_future.set_result(self.last_pdu)
         return used_len
 
     def getNextTID(self) -> int:


pymodbus/transaction/transaction.py
```diff
@@ -153,6 +153,9 @@ async def execute(self, no_response_expected: bool, request: ModbusPDU) -> Modbu
                     return response
                 except asyncio.exceptions.TimeoutError:
                     count_retries += 1
+                except asyncio.exceptions.CancelledError:
+                    self.recv_buffer = b""
```
janiversen, line 157: see other PR.

janiversen (changes requested):

pymodbus/transaction/transaction.py
```diff
@@ -194,7 +197,8 @@ def callback_data(self, data: bytes, addr: tuple | None = None) -> int:
                     raise ModbusIOException(
                         f"ERROR: request ask for id={self.request_dev_id} but got id={pdu.dev_id}, CLOSING CONNECTION."
                     )
-                self.response_future.set_result(self.last_pdu)
+                if not self.response_future.done():
```
janiversen, line 200: I think it would be more correct to notify the user by raising a ModbusIOException, because this is caused by a cancelation from outside pymodbus, and the app knows how to handle it correctly.

janiversen (changes requested):

pymodbus/transaction/transaction.py
jameshilliard: Ah, yeah, I guess if we purge the buffer before each send we don't need to clean up on exceptions.

pymodbus/transaction/transaction.py
jameshilliard: > notify the user by raising a ModbusIOException

If the future is already completed due to the future being cancelled I don't think we can set another exception like `ModbusIOException` on it. The `execute` function already raises a `ModbusIOException` after all retries are exhausted so I'm not sure we need to raise an exception from here.

> this is caused by a cancelation from outside pymodbus

I'm not sure that's always the case, I think the pymodbus retry loop `asyncio.wait_for` also cancels the `response_future` when the timeout triggers.

If the app cancels the operation a `ModbusIOException` doesn't seem to make sense to me, on cancellation I think the normal `CancelledError`(which should AFAIU happen automatically from within the `execute` codepath waiting on the future even when cancelled externally by the app) is what is expected to be emitted since cancellation is not neccesarially indicative of any actual modbus error but may simply be the app abandoning the operation for whatever reason.

janiversen: I think we cannot just silently ignore a real error! if this happens it means that client handles 2 requests in the same call an that is. serious bug.

jameshilliard: Maybe we should just log something?  I mean if there isn't an active transaction I'm not sure what else we can do here as we don't have a future we can use to return an exception to right?

jameshilliard: This can also happen if a slave writes garbage on the line without a request right?

janiversen: The style in pymodbus is to raise a ModbuIOException, when something like this happens....typical users do not look in the log.

If a slave (server) writes data without being asked to, there are no request waiting, the loop is not active and data will be removed with next request.

PR #2584, opened 2025-02-23
Ensure CancelledError is re-raised properly

jameshilliard:
Python's asyncio cancellation propagation requires `CancelledError` to be re-raised by any exception handlers catching it.

I'm pretty sure catching and replacing a `CancelledError` with a `ModbusIOException` is not correct as python's cancellation mechanism requires the `CancelledError` to propagate back to the cancellation initiator, instead what we could do is attach the `ModbusIOException` as exception context to the `CancelledError`.

Alternatively we could just remove this exception handler entirely as I don't think it's all that important to catch the `CancelledError` here unless we actually need to do some sort of cleanup operation.

Once pymodbus requires a minimum python version of 3.11 we could alternatively raise an `ExceptionGroup` with both the `CancelledError` and `ModbusIOException` in this exception handler.

Python 3.11+ `ExceptionGroup` syntax:
```python
except asyncio.exceptions.CancelledError as exc:
    raise ExceptionGroup("Request cancelled", [exc, ModbusIOException("Request cancelled outside pymodbus.")])
```

### DIFF
--- a/pymodbus/transaction/transaction.py
+++ b/pymodbus/transaction/transaction.py
@@ -155,7 +155,8 @@ class TransactionManager(ModbusProtocol):
                 except asyncio.exceptions.TimeoutError:
                     count_retries += 1
                 except asyncio.exceptions.CancelledError as exc:
-                    raise ModbusIOException("Request cancelled outside pymodbus.") from exc
+                    exc.__context__ = ModbusIOException("Request cancelled outside pymodbus.")
+                    raise
             if self.count_until_disconnect < 0:
                 self.connection_lost(asyncio.TimeoutError("Server not responding"))
                 raise ModbusIOException(


janiversen (changes requested):
ModbusIOException is just as good as CancelledError. Asyncio do not check for its own cancellation at this level.

Pymodbus needs to inform the app that something incorrect happened and that is done with our own exception. Please be aware this is a situation that should not happen in a well behaved app.

Catching the cancelled exception is actually important because this is the app stack, whereas the callback runs in its own stack space, so the user might not see the relation to the app.

janiversen:
Python 3.11 as baseline is unlikely for the next 5-10 years, we are still at 3.9 with test and the library works with 3.8

jameshilliard:
> ModbusIOException is just as good as CancelledError. Asyncio do not check for its own cancellation at this level.

It's not true that `ModbusIOException` will behave the same as `CancelledError` in this case, `asyncio.wait_for()` for explicitly catches `CancelledError` [here](https://github.com/python/cpython/blob/v3.13.2/Lib/asyncio/tasks.py#L503). Using `ModbusIOException` would result in the wrong exception being raised in the application code if the application does a modbus operation that times out inside an `asyncio.wait_for()` that is stricter than the pymodbus timeout.

> Pymodbus needs to inform the app that something incorrect happened and that is done with our own exception. Please be aware this is a situation that should not happen in a well behaved app.

Cancelling a request is arguably not "incorrect" behavior in many cases(pymodbus itself does this for timeouts), virtually all asyncio python libraries are expected to handle request cancellations in this way AFAIU.

> Catching the cancelled exception is actually important because this is the app stack

Catching the cancelled exception in and of itself is fine, that's how cleanup on cancellation is typically done, the issue is that the cancellation exception must be re-raised for the application code that initiated the cancellation to work correctly.

jameshilliard:
> Pymodbus needs to inform the app that something incorrect happened and that is done with our own exception.

If we want to raise our own pymodbus specific exception one option would be to create say a `ModbusCancelledError` exception or similar that extends the asyncio `CancelledError`, the important part here is that we need to raise an exception that will get caught by asyncio's own `CancelledError` exception handlers as `CancelledError` is a treated by asyncio as a special type of exception.

janiversen:
You are entitled to your opinion, however it is quite normal that a library raises something to alert the app that the state of the request is unknown.

wait_for works nicely with the ModbusIOException, which by the way also changes the exception, so changing is allowed.

I believe the current solution better than the bug you reported in your original issue.

jameshilliard:
> it is quite normal that a library raises something to alert the app that the state of the request is unknown

But it's not really unknown in the event of a `CancelledError` as `CancelledError` has a specific meaning, the state of the request is that it was cancelled before a response was received.

> wait_for works nicely with the ModbusIOException

I don't see how it could work properly here. Erasing the `CancelledError` exception like you are doing actually prevents the application from functioning correctly when chaining `asyncio.wait_for` calls.

> which by the way also changes the exception, so changing is allowed

The `asyncio.wait_for` [`except exceptions.CancelledError as exc:`](https://github.com/python/cpython/blob/v3.13.2/Lib/asyncio/tasks.py#L503) handler will not catch a `ModbusIOException`, in fact this changing of the `CancelledError` exception to `TimeoutError` is required for `asyncio.wait_for` to function correctly(i.e. for application code to determine if an `asyncio.wait_for` triggered a timeout), hence why pymodbus needs to raise an exception that will get caught by the `CancelledError` exception handler in `asyncio.wait_for`. Library code in general should not change a `CancelledError` to a different exception as it would prevent an application side `asyncio.wait_for` from raising a `TimeoutError` properly.

> I believe the current solution better than the bug you reported in your original issue.

This isn't really even related to the original bug AFAIU.

janiversen:
It is related to your original issue, because that started the whole discussion.

jameshilliard:
> It is related to your original issue, because that started the whole discussion.

I had originally caught a `asyncio.CancelledError` exception so I could do cleanup there and then re-raise the same exception, my implementation was using the `asyncio.CancelledError` exception handler in the [documented way](https://docs.python.org/3/library/asyncio-task.html#task-cancellation)(i.e. catch -> cleanup -> re-raise). The way you implemented the `asyncio.CancelledError` exception handling does not follow the [documentation](https://docs.python.org/3/library/asyncio-task.html#task-cancellation) and will break application code.

> But as with other libraries pymodbus catches many different exceptions, some are handled and others are transformed that is how the library works, always re-raising the original exception would be quite a change.

So `asyncio.CancelledError` is very special and can not be treated like other exceptions as asyncio standard library code has explicit special case handling for it, extra care must be taken to ensure that cancellation exception propagation is done properly, especially the part where the exception gets re-raised if caught.

> Unless there is a violation of a PEP or similar, I consider the current code valid and most importantly consistent.

The expectations regarding catching an `asyncio.CancelledError` in an exception handler are documented [here](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError) as well, re-raising the `asyncio.CancelledError` exception is effectively mandatory in most situations:
> This exception can be caught to perform custom operations when asyncio Tasks are cancelled. **In almost all situations the exception must be re-raised.**

janiversen:
And where is this "demand" documented? I have seen no PEP about this, but of course I can easily have missed something.

To me it sounds more like it is your interpretation, and anyhow the app should handle ModbusIOException.

jameshilliard:
> And where is this "demand" documented?

In the official python docs as I [linked to](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError) is where it's most explicitly stated that we must re-raise the exception in virtually all cases, there's some [additional info in PEP-3156](https://peps.python.org/pep-3156/#tasks) regarding why we really need to re-raise the `asyncio.CancelledError` exception if caught as it will break task cancellation essentially.:
> Cancelling a task that’s not done yet throws an asyncio.CancelledError exception into the coroutine. If the coroutine doesn’t catch this (or if it re-raises it) the task will be marked as cancelled (i.e., cancelled() will return True); but if the coroutine somehow catches and ignores the exception it may continue to execute (and cancelled() will return False).

> "almost all" is a suggestion not a demand

My interpretation of this is that we must re-raise the `asyncio.CancelledError` exception unless we have a really good reason not to.

janiversen:
My interpretation is different, pymodbus detected an unexpected problem and the consistency of the connection is no longer guaranteed (since cleanup etc are not done), so the recommendation is to restart the client.

Your app does something not foreseen, which is your choice, but the app must handle the consequences of this.

It does not sound logical to me to have different timeouts for different types of requests to the same device, however if there are a valid reason (which I assume you have), the correct solution would be to add timeout= at request level for both sync and async of course with corresponding tests.

jameshilliard:
> pymodbus detected an unexpected problem and the consistency of the connection is no longer guaranteed (since cleanup etc are not done)

This isn't an accurate description of the meaning of the `asyncio.CancelledError` here, when the `asyncio.CancelledError` is thrown into the coroutine it is effectively a signal from the app to cancel the request, it's very much not a consistency error originating from the connection side, and cleanup is now actually being done here since we now unconditionally clean the buffer before each request(so cancellation no longer inhibits the necessary cleanup operation). If we didn't unconditionally clean the buffer before the request we could do something like a catch -> cleanup -> re-raise pattern that I had originally used here to ensure cleanup is done.

> the recommendation is to restart the client.

Even pymodbus doesn't restart the client until multiple requests timeout, so handling the app initiated coroutine cancellation is not really any different, we simply need to propagate up the `asyncio.CancelledError` to not break python's asyncio cancellation expectations.

> Your app does something not foreseen, which is your choice, but the app must handle the consequences of this.

Task/coroutine cancellation is hardly some unforeseen scenario, it's a rather fundamental feature in python's asyncio implementation(it's used for shutting down applications for example and pymodbus uses it internally as well), it's expected that python libraries handle cancellation properly, deliberately breaking asyncio cancellation capabilities should not be done without a very good reason.

> It does not sound logical to me to have different timeouts for different types of requests to the same device

Different requests have different response times based on how much data they return among other reasons and also the current timeout implementation in pymodbus is not per device, it is per client.

> however if there are a valid reason (which I assume you have), the correct solution would be to add timeout= at request level for both sync and async of course with corresponding tests.

I think I mentioned I could try and add this as a feature in the past, however this does not negate the necessity for an application to handle task/coroutine cancellation properly according to normal python asyncio expectations.

janiversen:
I am not sure what we actually are discussing here, the PR is merged a while ago, and the code works.

You are entitled to your opinion but so am I and I see things differently than you. This do not make your opinion or my opinion wrong but merely different solutions to a problem, please accept that.

If you do not like how pymodbus works, then do not upgrade.

janiversen:
It is considered a illegal action in pymodbus and reported correctly. A request is considered an atom that should not be interrupted.

You can set a timeout on the client, which is valid for all requests, that is considered enough....but if you really need different timeouts for different requests the correct solution is to add timeout= at all requests

jameshilliard:
> It is considered a illegal action in pymodbus and reported correctly.

Why? It's literally something the code effectively has to support already due to the retry logic having the same requirements in regard to buffer cleanups. Not supporting asyncio task cancellation will make pymodbus incompatible with many async frameworks as those often use task cancellation for various purposes like task reaping on shutdown.

janiversen:
Because the effects are unpredictable, it depends a lot on where the code pointer is at the time of cancellation, and the cleanup would be totally missing. If you e,g. cancel while asyncio is sending, you will need to restart the client and your devices might f.x. wait minutes for the rest of the message and thus block communication.

Actually there are a number of places where pymodbus catches cancellation and handles it internally, because it was started internally and the side effects are known.

You are only focusing on one specific case, but please do not forget all the others.

janiversen:
If you say so, but the send is handled by asyncio and actually happens after pdu_send()

jameshilliard:
> Actually there are a number of places where pymodbus catches cancellation and handles it internally, because it was started internally and the side effects are known.

This looks problematic/unnecessary, we should do something like #2588 instead I think.

janiversen:
> > Actually there are a number of places where pymodbus catches cancellation and handles it internally, because it was started internally and the side effects are known.
> 
> 
> 
> This looks problematic/unnecessary, we should do something like #2588 instead I think.

Raise that PR with the asyncio library, this is how asyncio works.

jameshilliard:
> this is how asyncio works.

I really don't see from where you've gotten this idea that it's normal to fully suppress asyncio cancellation exceptions, especially since the official python documentation quite strongly recommends against doing so.

janiversen:
I was talking about send()

PR #2586, opened 2025-02-25
Add transaction task cancellation test

jameshilliard:
This test demonstrates why suppression of `asyncio.CancelledError` is broken.

### DIFF
--- a/pymodbus/transaction/transaction.py
+++ b/pymodbus/transaction/transaction.py
@@ -154,8 +154,6 @@ class TransactionManager(ModbusProtocol):
                     return response
                 except asyncio.exceptions.TimeoutError:
                     count_retries += 1
-                except asyncio.exceptions.CancelledError as exc:
-                    raise ModbusIOException("Request cancelled outside pymodbus.") from exc
             if self.count_until_disconnect < 0:
                 self.connection_lost(asyncio.TimeoutError("Server not responding"))
                 raise ModbusIOException(
--- a/test/transaction/test_transaction.py
+++ b/test/transaction/test_transaction.py
@@ -173,6 +173,26 @@ class TestTransaction:
             await asyncio.sleep(0.1)
             assert response == await resp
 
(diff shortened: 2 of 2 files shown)


janiversen:
It basically shows nothing! you cancel a request in process which is not supported in pymodbus (independent what other libraries do)!

And by coincidence I have just been debugging a problem in aiohttp which turned out to be caused by a cancel minutes earlier.

jameshilliard:
> It basically shows nothing! you cancel a request in process which is not supported in pymodbus (independent what other libraries do)!

Without removing the `asyncio.CancelledError` suppression this test does not pass for multiple reasons, this test is demonstrating standard asyncio task cancellation behavior. Before 74d1dcf774ba36badad1e2abb1696fc834c62ebe task cancellation worked fine(I'm not sure why the `asyncio.CancelledError` suppression change that broke asyncio task cancellation was merged with the `callback_data` bug fix as that's an unrelated change).

> And by coincidence I have just been debugging a problem in aiohttp which turned out to be caused by a cancel minutes earlier.

FYI this `asyncio.CancelledError` suppression will likely also cause issues if pymodbus is used with aiohttp along with many other asyncio frameworks due to the [shutdown code expecting task cancellation to not be broken](https://github.com/aio-libs/aiohttp/blob/v3.11.13/aiohttp/web.py#L443-L464).

janiversen:
Actually it is not unrelated, cancelation can cause a side effect in the callback, it was the only way I could reproduce that problem.

jameshilliard:
> cancelation can cause a side effect in the callback,

The future in the callback could have been cancelled by the retry handler as well, it's not strictly limited to application side task cancellation.

> it was the only way I could reproduce that problem

This seems strange to me, can you share the test application?

> my test app continued after the cancel as if everything was normal

Why shouldn't it continue? Cancellation of a request by the calling application is not indicative of a modbus protocol error.

janiversen:
it is the retry handler! there are no higher level wait in pymodbus.

PR #2587, opened 2025-02-25
Document task cancellation brokenness

jameshilliard:
Document the broken task cancellation behavior described [here](https://github.com/pymodbus-dev/pymodbus/pull/2584#issuecomment-2682693129) as being intentional.

### DIFF
--- a/API_changes.rst
+++ b/API_changes.rst
@@ -2,6 +2,10 @@ API changes
 ===========
 Versions (X.Y.Z) where Z > 0 e.g. 3.0.1 do NOT have API changes!
 
+API changes 3.8.7
+-----------------
+- Drop support for asyncio task cancellation
+
 API changes 3.8.0
 -----------------
 - ModbusSlaveContext, removed zero_mode parameter.
--- a/doc/source/client.rst
+++ b/doc/source/client.rst
@@ -10,6 +10,9 @@ The client is NOT thread safe, meaning the application must ensure that calls ar
 This is only a problem for synchronous applications that use multiple threads or for asynchronous applications
 that use multiple :mod:`asyncio.create_task`.
 
+The client no longer supports task cancellation, please don't use pymodbus with asyncio frameworks like aiohttp that rely on task cancellation.
+Note that pymodbus is deliberately NOT safe to use with code that relies on task cancellation.
+
 It is allowed to have multiple client objects that e.g. each communicate with a TCP based device.
 
 


API_changes.rst
```diff
@@ -2,6 +2,10 @@ API changes
 ===========
 Versions (X.Y.Z) where Z > 0 e.g. 3.0.1 do NOT have API changes!
 
+API changes 3.8.7
+-----------------
+- Drop support for asyncio task cancellation
```
janiversen, line 7: This is not correct, it was a problem earlier see your own callback problem.

There are no API changes.

doc/source/client.rst
```diff
@@ -10,6 +10,9 @@ The client is NOT thread safe, meaning the application must ensure that calls ar
 This is only a problem for synchronous applications that use multiple threads or for asynchronous applications
 that use multiple :mod:`asyncio.create_task`.
 
+The client no longer supports task cancellation, please don't use pymodbus with asyncio frameworks like aiohttp that rely on task cancellation.
```
janiversen, line 13: Could you possibly be more negative.

janiversen (changes requested):
It is never wrong to document conventions, but it needs to be to be done in a positive tone!

janiversen:
And just for the record it is not broken...it is intentional to solve a problem that was undetected before.

jameshilliard:
> And just for the record it is not broken...it is intentional to solve a problem that was undetected before.

You quite clearly stated that task cancellation is not supported and changed the API semantics to ensure code that uses task cancellation can not function properly at all(previously it would work most of the time), I'm simply documented that pymodbus having broken task cancellation is an intentional design decision here.

doc/source/client.rst
jameshilliard: It's pretty hard to put a positive spin on dropping support for widely used asyncio features like task cancellations IMO.

janiversen: Ok then let's close it instead.

jameshilliard: So it's better to keep users unaware that the library doesn't follow normal asyncio task cancellation conventions so that they get tricked into using it with incompatible applications? Our pymodbus integration is by far the most painful library integration I've ever had to maintain due to pymodbus not following typical python conventions/expectations. We have had to put wrappers around virtually every pymodbus operation in our code to get even somewhat sane exceptions being thrown.

janiversen: As I wrote earlier it is good to document conventions, but you language was not inline with the rest of the documentation and you wrote you could not write it less negative hence the close.

PR #2588, opened 2025-02-25
Refactor reconnect_task so that asyncio.CancelledError is not suppressed

jameshilliard:
Suppressing task cancellation in the reconnect_task is not necessary here and is likely to cause issues with application shutdown.

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -468,9 +468,8 @@ class ModbusProtocol(asyncio.BaseProtocol):
             Log.debug("Wait 1s before reopening listener.")
             await asyncio.sleep(1)
             await self.listen()
-        except asyncio.CancelledError:
-            pass
-        self.reconnect_task = None
+        finally:
+            self.reconnect_task = None
 
     async def do_reconnect(self) -> None:
         """Handle reconnect as a task."""
@@ -490,8 +489,10 @@ class ModbusProtocol(asyncio.BaseProtocol):
                     self.comm_params.reconnect_delay_max,
                 )
         except asyncio.CancelledError:
-            pass
-        self.reconnect_task = None
+            self.reconnect_delay_current = self.comm_params.reconnect_delay or 0.0
+            raise
+        finally:
+            self.reconnect_task = None
 
     # ----------------- #
     # The magic methods #
--- a/test/transport/test_protocol.py
+++ b/test/transport/test_protocol.py
@@ -305,7 +305,8 @@ class TestTransportProtocol2:
         assert client.reconnect_delay_current == client.comm_params.reconnect_delay * 2
         assert not client.reconnect_task
         client.connect.side_effect = asyncio.CancelledError("stop loop")
-        await client.do_reconnect()
+        with pytest.raises(asyncio.CancelledError):
+            await client.do_reconnect()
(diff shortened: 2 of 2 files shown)


pymodbus/transport/transport.py
```diff
@@ -468,9 +468,8 @@ async def do_relisten(self) -> None:
             Log.debug("Wait 1s before reopening listener.")
             await asyncio.sleep(1)
             await self.listen()
-        except asyncio.CancelledError:
-            pass
-        self.reconnect_task = None
+        finally:
+            self.reconnect_task = None
```
janiversen, line 472: This is wrong, because it breaks to code, self.reconnect_delay_current is never reset and the listeners are not popped.

pymodbus/transport/transport.py
```diff
@@ -490,8 +489,10 @@ async def do_reconnect(self) -> None:
                     self.comm_params.reconnect_delay_max,
                 )
         except asyncio.CancelledError:
-            pass
-        self.reconnect_task = None
+            self.reconnect_delay_current = self.comm_params.reconnect_delay or 0.0
```
janiversen, line 492: This is wrong.

test/transport/test_protocol.py
```diff
@@ -305,7 +305,8 @@ async def test_do_reconnect(self, client):
         assert client.reconnect_delay_current == client.comm_params.reconnect_delay * 2
         assert not client.reconnect_task
         client.connect.side_effect = asyncio.CancelledError("stop loop")
-        await client.do_reconnect()
+        with pytest.raises(asyncio.CancelledError):
+            await client.do_reconnect()
```
janiversen, line 309: The as such if correct, with the changes, but is shows very well the problem with the PR. 

This PR leaves the asyncio task with an uncaught exception, which is depending on the python version will be reported...and at least shows up in my code debugger. The old code secured that the task terminated gracefully and correctly.

janiversen (changes requested):
I am really not sure this is a good idea, mainly because it introduces an uncaught exception in the code and it really do not improve the code.

pymodbus/transport/transport.py
jameshilliard: > This is wrong.

Should this be in the finally block instead?

> I am also surprised that you do not re-raise the exception....you continue suppressing.

But I do re-raise the expression, right below this line...

test/transport/test_protocol.py
jameshilliard: > This PR leaves the asyncio task with an uncaught exception, which is depending on the python version will be reported...and at least shows up in my code debugger. The old code secured that the task terminated gracefully and correctly.

When an asyncio task/coroutine is cancelled(in this case by the side_effect on the connect call) it's expected that the cancellation exception propagates upwards, although in this case the cancellation is being triggered a bit differently than normal since the cancellation is originating from the client.connect call.

> and the close() have been and are tested thoroughly due to the fact that we had some nasty bugs 2 years ago, so I am not sure what problem there are with close().

I recall this code was breaking in weird ways before we started inhibiting the reconnects entirely more recently than that. This code is likely to cause problems for task reapers that kill all tasks in an application on shutdown.

janiversen:
Seems abandoned.

PR #2589, opened 2025-02-25
Flush the send buffer before each send

jameshilliard:
We need to ensure the send buffer is flushed before calling `pdu_send` based on [this](https://github.com/pymodbus-dev/pymodbus/pull/2584#issuecomment-2683017353) comment.


janiversen (changes requested):
ANot needed code change, that probably will not be called in a lot of cases.

asyncio send have a non-synchronous mechanism and we should not interfere with that.

PR #2611, opened 2025-03-16
Simplify ruff config

alexrudd2:
Use `extend-exclude` instead of `exclude` to take advantage of the default exclusions.


janiversen (approved):
Nice to see you back, thanks.

PR #2618, opened 2025-03-28
Fix exception error message for decoding response

Szewcson:
When decoding exception frame Error message is showing always exception code as 0. Also in error message I think showing Exception code in human readable format will be more helpful, as well as showing failing FC instead FC with exception bit set.

### DIFF
--- a/pymodbus/pdu/decoders.py
+++ b/pymodbus/pdu/decoders.py
@@ -102,7 +102,7 @@ class DecodePDU:
         """Decode a frame."""
         try:
             if (function_code := int(frame[0])) > 0x80:
-                pdu_exp = base.ExceptionResponse(function_code & 0x7F, int(frame[1]))
+                pdu_exp = base.ExceptionResponse(function_code & 0x7F)
                 pdu_exp.decode(frame[1:])
                 return pdu_exp
             if not (pdu_type := self.lookup.get(function_code, None)):
--- a/pymodbus/pdu/decoders.py
+++ b/pymodbus/pdu/decoders.py
@@ -102,7 +102,7 @@ class DecodePDU:
         """Decode a frame."""
         try:
             if (function_code := int(frame[0])) > 0x80:
-                pdu_exp = base.ExceptionResponse(function_code & 0x7F)
+                pdu_exp = base.ExceptionResponse(function_code & 0x7F, int(frame[1]))
                 pdu_exp.decode(frame[1:])
                 return pdu_exp
             if not (pdu_type := self.lookup.get(function_code, None)):
--- a/pymodbus/pdu/pdu.py
+++ b/pymodbus/pdu/pdu.py
@@ -133,7 +133,6 @@ class ExceptionResponse(ModbusPDU):
         return (
             f"{self.__class__.__name__}("
             f"dev_id={self.dev_id}, "
-            f"transaction_id={self.transaction_id}, "
             f"function_code={self.function_code}, "
             f"exception_code={self.exception_code})"
         )
--- a/pymodbus/pdu/pdu.py
+++ b/pymodbus/pdu/pdu.py
@@ -59,7 +59,7 @@ class ModbusPDU:
             raise ValueError(f"0 < address {address} < 65535 !")
 
     def __str__(self) -> str:
-        """Build a representation of an exception response."""
+        """Build a representation of a Modbus response."""
         return (
             f"{self.__class__.__name__}("
             f"dev_id={self.dev_id}, "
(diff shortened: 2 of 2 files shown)


Szewcson:
Also maybe exception code and function code should be added to __str__?

pymodbus/pdu/pdu.py
```diff
@@ -118,6 +118,20 @@ class ExceptionResponse(ModbusPDU):
     GATEWAY_PATH_UNAVIABLE = 0x0A
     GATEWAY_NO_RESPONSE = 0x0B
 
+    exception_code_dict = {
+        0x00: "None",
```
janiversen, line 122: Please use the constants defined above, to avoid mismatch between constant and value.

pymodbus/pdu/decoders.py
```diff
@@ -102,7 +102,7 @@ def decode(self, frame: bytes) -> base.ModbusPDU | None:
         """Decode a frame."""
         try:
             if (function_code := int(frame[0])) > 0x80:
-                pdu_exp = base.ExceptionResponse(function_code & 0x7F)
+                pdu_exp = base.ExceptionResponse(function_code & 0x7F, int(frame[1]))
```
janiversen, line 105: please do not convert to int, if a conversion is needed it should be done where needed.

pymodbus/pdu/pdu.py
janiversen: This is a constant meaning the name should be capitalized

pymodbus/pdu/pdu.py
```diff
@@ -137,3 +152,4 @@ def encode(self) -> bytes:
     def decode(self, data: bytes) -> None:
         """Decode a modbus exception response."""
         self.exception_code = int(data[0])
+        self.exception_name = self.exception_code_dict.get(self.exception_code, f"Unknown Exception value: {self.exception_code}")
```
janiversen, line 155: Not sure what this does, it really seems not needed.

pymodbus/pdu/pdu.py
```diff
@@ -128,7 +142,8 @@ def __init__(
         super().__init__(transaction_id=transaction, dev_id=device_id)
         self.function_code = function_code | 0x80
         self.exception_code = exception_code
-        Log.error(f"Exception response {self.function_code} / {self.exception_code}")
+        self.exception_name = self.exception_code_dict.get(self.exception_code, f"Unknown Exception value: {self.exception_code}")
```
janiversen, line 145: no need to make an object variable, remove self.

pymodbus/pdu/pdu.py
```diff
@@ -128,7 +142,8 @@ def __init__(
...
+        Log.error(f"Exception response FC{self.function_code & 0x7F} / {self.exception_name}")
```
janiversen, line 146: no need to add 0x7f just use the function_code parameter.

Just showing the name, makes it more difficult for log parsers so the code should still be presented.

Showing the exception name makes sense for people who do not know the modbus standard, but the same people to not know the function codes either....so it would be better to either use names for both or not at all.

janiversen (changes requested):
The idea is good, but there are a number of comments.

pymodbus/pdu/pdu.py
Szewcson: Ok, so which approach here will be the best?

janiversen: make \_\_str\_\_

Szewcson: > in general instead of logging, it would be better as \_\_str\_\_ allowing the app to do what it want.

For now __str__ only shows base pdu variables. Which can be confusing if app will use __str__ on exception.

Szewcson: And with the log?

janiversen: no

Szewcson:
But this log was there probably for opposite situation - generation of the exception. Maybe we need some flag then.

janiversen:
the generation of the exception is used in the server.

pymodbus/pdu/pdu.py
```diff
@@ -128,7 +127,21 @@ def __init__(
         super().__init__(transaction_id=transaction, dev_id=device_id)
         self.function_code = function_code | 0x80
         self.exception_code = exception_code
-        Log.error(f"Exception response {self.function_code} / {self.exception_code}")
+
+    def __str__(self) -> str:
+        """Build a representation of an exception response."""
+        return (
+            f"{self.__class__.__name__}("
```
janiversen, line 134: This is far too complicated !!! you are adding a lot of fields that are not used and a lot more that are only used with some function codes.

It is a representation of the "exception response" NOT every possible response.

janiversen (changes requested):

janiversen (changes requested):

Issue #2621, opened 2025-04-01
Default ModbusSlaveContext will share the same dataBlocks for every instance


maximeLeurent:
Hello,

For my project, I used several Modbus Server in one unique process, each server in their own thread.

I realised that that each device was sharing the same memory : when two devices are writing at the same register number, with different values, they would rewrite the values of the other device.

I found the cause of this sharing memory in the definition of the __init__ function of the class ModbusSlaveContext

```
class ModbusSlaveContext(ModbusBaseSlaveContext):
    def __init__(self, *_args,
                    di=ModbusSequentialDataBlock.create(),
                    co=ModbusSequentialDataBlock.create(),
                    ir=ModbusSequentialDataBlock.create(),
                    hr=ModbusSequentialDataBlock.create(),
                    zero_mode=False):
        """Initialize the datastores."""
        self.store = {}
        self.store["d"] = di
        self.store["c"] = co
        self.store["i"] = ir
        self.store["h"] = hr
        self.zero_mode = zero_mode
```
I wasn't calling ModbusSlaveContext with any di, co, ir, or hr variable
therefore all self.store["d"]  was using the same instance of ModbusSequentialDataBlock

it is the same problem of default argument of this piece of code:
```
def f(l =[]):
    l.append(3)
    print(l)

f()
>>>[3]
f()
>>>[3, 3]
```

I propose you the following code

```
class ModbusSlaveContext(ModbusBaseSlaveContext):
    """Create a modbus data model with data stored in a block.

    :param di: discrete inputs initializer ModbusDataBlock
    :param co: coils initializer ModbusDataBlock
    :param hr: holding register initializer ModbusDataBlock
    :param ir: input registers initializer ModbusDataBlock
    :param zero_mode: Not add one to address

        When True, a request for address zero to n will map to
        datastore address zero to n.

        When False, a request for address zero to n will map to
        datastore address one to n+1, based on section 4.4 of
        specification.

        Default is False.

    """

    def __init__(self, *_args,
                    di: ModbusSequentialDataBlock=None,
                    co: ModbusSequentialDataBlock=None,
                    ir: ModbusSequentialDataBlock=None,
                    hr: ModbusSequentialDataBlock=None,
                    zero_mode=False):
        """Initialize the datastores."""
        self.store = {}
        self.store["d"] = di or ModbusSequentialDataBlock.create()
        self.store["c"] = co or ModbusSequentialDataBlock.create()
[...]

PR #2622, opened 2025-04-01
Correction  : default argument of ModbusDeviceContext.__init__ was creating shared ModbusSequentialDataBlock between instances

maximeLeurent:
PR for issue #2621 
`ModbusSlaveContext.__init__` was using default argument that are mutable, creating shared instance bug when using multiple ModbusSlaveContext

Adding a test that is failling in previous version of the code

### DIFF
--- a/pymodbus/datastore/context.py
+++ b/pymodbus/datastore/context.py
@@ -83,10 +83,10 @@ class ModbusDeviceContext(ModbusBaseDeviceContext):
     """
 
     def __init__(self, *_args,
-                    di: ModbusSequentialDataBlock = None,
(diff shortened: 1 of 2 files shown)


maximeLeurent:
code 
`di or ModbusSequentialDataBlock.create()`
will not use `ModbusSequentialDataBlock.create()` if `di` can be evale to False, like an empty list.

I saw this kind of code in the function `ModbusDeviceContext.register`. Let me know if a ```if di is None:``` is more suitable.

janiversen (approved):
Thanks.

PR #2638, opened 2025-04-14
Don't bool-test explicit datastores

smurfix:
Empty data stores tend to be false-y when tested,
which is not the intent here.

Should probably be backported to master

### DIFF
--- a/pymodbus/datastore/context.py
+++ b/pymodbus/datastore/context.py
@@ -90,10 +90,10 @@ class ModbusDeviceContext(ModbusBaseDeviceContext):
                 ):
         """Initialize the datastores."""
         self.store = {}
-        self.store["d"] = di or ModbusSequentialDataBlock.create()
-        self.store["c"] = co or ModbusSequentialDataBlock.create()
-        self.store["i"] = ir or ModbusSequentialDataBlock.create()
-        self.store["h"] = hr or ModbusSequentialDataBlock.create()
+        self.store["d"] = di if di is not None else ModbusSequentialDataBlock.create()
+        self.store["c"] = co if di is not None else ModbusSequentialDataBlock.create()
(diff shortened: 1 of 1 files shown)


pymodbus/datastore/context.py
```diff
@@ -90,10 +90,10 @@ def __init__(self, *_args,
...
         """Initialize the datastores."""
         self.store = {}
-        self.store["d"] = di or ModbusSequentialDataBlock.create()
-        self.store["c"] = co or ModbusSequentialDataBlock.create()
-        self.store["i"] = ir or ModbusSequentialDataBlock.create()
-        self.store["h"] = hr or ModbusSequentialDataBlock.create()
+        self.store["d"] = di if di is not None else ModbusSequentialDataBlock.create()
+        self.store["c"] = co if di is not None else ModbusSequentialDataBlock.create()
+        self.store["i"] = ir if di is not None else ModbusSequentialDataBlock.create()
+        self.store["h"] = hr if di is not None else ModbusSequentialDataBlock.create()
```
chkr-private, line 96: The last 3 lines of the new code look like cut'n'paste mistakes.

e.g. the original code
`self.store["c"] = co or ModbusSequentialDataBlock.create()`
tested variable `co` and conditionally assigned it

in the new code, `co` is conditionally assigned but `di` was tested:
`self.store["c"] = co if di is not None else ModbusSequentialDataBlock.create()`

smurfix: Ugh. Obviously. Thanks, will fix ASAP.

janiversen: @smurfix I thought you had tested the code, if you open a new PR please add tests to ensure it works correctly.

But you have time, because unless a serious bug is found there will be no 3.9 updates before late summer.

smurfix:
Yeah I did test the code, but unfortunately I did it with `di`. Surprise …

PR #2655, opened 2025-04-30
fixed string type handling in simulator

Lucas-Morais-Freire:
Hello. I was playing with the string functionality when I tried to allocate 3 16-bit registers to hold a 6 character string (which should be ok) and an exception was raised. I then noticed that the number of registers that the string allocates was being computed as `stop - start`, which actually undercounts the number of registers allocated by one. I just changed that calculation where necessary. I tested it and it worked okay, I hope I didn't miss anything. Cheers!

### DIFF
--- a/pymodbus/datastore/simulator.py
+++ b/pymodbus/datastore/simulator.py
@@ -198,14 +198,14 @@ class Setup:
 
     def handle_type_string(self, start, stop, value, action, action_parameters):
         """Handle type string."""
-        regs = stop - start
+        regs = stop - start + 1
         reg_len = regs * 2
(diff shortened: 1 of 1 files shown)


janiversen:
Thanks your code looks ok, but it seems you broke a lot of tests, please fix.

janiversen:
It might be you just broke the standard configuration and not the tests themselves.

Lucas-Morais-Freire:
yeah... I'll check those tests and see what I've missed.

Lucas-Morais-Freire:
I figured it out. It was indeed an error in my setup. Sorry for the inconvenience

Issue #2660, opened 2025-05-03
Serial forwarder demo doesnt work with write single register


sthome-cz:
Steps to reproduce

* Start `serial_forwarder` from <https://github.com/pymodbus-dev/pymodbus/blob/dev/examples/contrib/serial_forwarder.py>
* Send `modbus` command to `write_single_register` - for Goodwe inverter
* The response on `RTU` is ok
* The response on `TCP` side is corrupted

Evidences

* Communication on `RTU` side is as expected, enable (on) the switch
  * Request: `0xf7 0x6 0xb9 0xec 0x0 0x1 0xb8 0x35`
  * Response: `0xf7 0x6 0xb9 0xec 0x0 0x1 0xb8 0x35`
  * Critical section is tuple `0x0 0x1` starting on the 5th byte
* Communication on `TCP` side is corrupted, as the packet is encoded like
  * Request: `0x82 0xcc 0x0 0x0 0x0 0x6 0xf7 0x6 0xb9 0xec 0x0 0x1`
  * Response: `0x82 0xcc 0x0 0x0 0x0 0x6 0xf7 0x6 0xb9 0xec 0x0 0x0`
  * Critical section is also `0x0 0x1` in request, two bytes from the end of the packet
  * And the corrupted section is `0x0 0x0` in the response, the last two bytes

The reason (in the code), why `pymodbus` works like this is in the code, where the condition makes to always `return [0]`.

https://github.com/pymodbus-dev/pymodbus/blob/b4cecc36e72c03645178320d516d53a3a58ee092/pymodbus/datastore/remote.py#L34-L41

Datastore is called during the execution of the `execute` in `WriteSingleRegisterRequest` object.

https://github.com/pymodbus-dev/pymodbus/blob/b4cecc36e72c03645178320d516d53a3a58ee092/pymodbus/pdu/register_message.py#L181-L194

Due to the condition above, the returned value after the `WriteSingleRegisterRequest` is always `0` and this will fail the command in the management application.

What is the correct way to fix this issue ? I have removed the condition for the test and the command is now executed successfully. Unfortunately I dont think, this is correct, because the program now executes the command to write the register value, then read its state back. AFAIK this shouldnt be necessary and the current state of the register is contained in the response on the `RTU` side.

Am I missing something, or is it a bug ?

janiversen:
please add a debug log (using the pymodbus call) so we can see what happens.

[...]

PR #2663, opened 2025-05-09
Fix pymodbus TCP / RTU forwarder

sthome-cz:
Getting response of WriteSingleRegister is read correctly.

Problem described at <https://github.com/pymodbus-dev/pymodbus/issues/2660>

### DIFF
--- a/pymodbus/datastore/remote.py
+++ b/pymodbus/datastore/remote.py
@@ -33,11 +33,10 @@ class RemoteDeviceContext(ModbusBaseDeviceContext):
 
     def getValues(self, fc_as_hex, _address, _count=1):
         """Get values from remote device."""
-        if fc_as_hex in self._write_fc:
-            return [0]
-        group_fx = self.decode(fc_as_hex)
-        func_fc = self.__get_callbacks[group_fx]
-        self.result = func_fc(_address, _count)
+        if not fc_as_hex in self._write_fc:
+            group_fx = self.decode(fc_as_hex)
+            func_fc = self.__get_callbacks[group_fx]
+            self.result = func_fc(_address, _count)
         return self.__extract_result(self.decode(fc_as_hex), self.result)
 
     def setValues(self, fc_as_hex, address, values):
@@ -116,7 +115,11 @@ class RemoteDeviceContext(ModbusBaseDeviceContext):
             if fc_as_hex in {"d", "c"}:
                 return result.bits
             if fc_as_hex in {"h", "i"}:
-                return result.registers
+                if result.registers:
+                    return result.registers
+                else:
+                    # WriteSingleRegisterResponse don't have registers value set
+                    return [result.value]
         else:
             return result
         return None


pymodbus/datastore/remote.py
```diff
@@ -33,11 +33,10 @@ def reset(self):
 
     def getValues(self, fc_as_hex, _address, _count=1):
         """Get values from remote device."""
-        if fc_as_hex in self._write_fc:
-            return [0]
-        group_fx = self.decode(fc_as_hex)
-        func_fc = self.__get_callbacks[group_fx]
-        self.result = func_fc(_address, _count)
+        if not fc_as_hex in self._write_fc:
```
janiversen, line 36: This means that all requests that not write are called twice, that sounds wrong.

pymodbus/datastore/remote.py
```diff
@@ -116,7 +115,11 @@ def __extract_result(self, fc_as_hex, result):
             if fc_as_hex in {"d", "c"}:
                 return result.bits
             if fc_as_hex in {"h", "i"}:
-                return result.registers
```
janiversen, line 119: Why change this ?
apart from that it is wrong...it is perfectly legal to return 0 registers.

janiversen (changes requested):
It would be nice if you tested the pull request, before submitting it...you should have seen the double read.

pymodbus/datastore/remote.py
sthome-cz: As you said earlier, everything is within this one file, or not ?

I don't know all the internals of this project. It seems to me, that the object `WriteSingleRegisterResponse` works a little bit differently, than the others objects. The internal value `registers` is not set there, that's the reason, why I had to change it to `return [result.value].

I don't know, whether this is correct or not, it's highly dependent on the internal design.

janiversen:
Why on earth does poetry need to be newer ????

Just make a venv, which helps you keep your projects separate.

pymodbus/datastore/remote.py
janiversen: WriteSingleRegisterResponse do not work differently, it sets registers, just look in the decode method.

pymodbus/datastore/remote.py
janiversen: As I see it, it is far from the same, you call __extract_result, without being sure that self.result is filled...that is an accident waiting to happen.

janiversen (changes requested):

janiversen:
Why not consider just to change setValues.

You also need to solve the CI problem, you can run the checks locally.

sthome-cz:
> Why not consider just to change setValues.
> 
> You also need to solve the CI problem, you can run the checks locally.

I will solve the CI problem, for sure. Just waiting what the other changes are gonna be.

> Why not consider just to change setValues.

I have to test the changes on version 3.7.0. The implementation of `WriteSingleRegisterResponse:decode` is a little bit different there. So you are probably right, that part of this change is not necessary for newer versions.

pymodbus/datastore/remote.py
sthome-cz: Is it possible, that `getValues` with write command will be called on this object, before `setValues` ?

In that case, exception should be raised as `return [0]` is not correct either. What is the correct way to raise exception ? Should it look similar to `raise ValueError(f"setValues() called with an non-write function code {fc_as_hex}")` ?

janiversen:
3.7.0 is not supported and actually have quite a lot of known bugs.

sthome-cz:
> Why on earth does poetry need to be newer ????

Poetry has changed a lot since the version 2.0 - <https://github.com/python-poetry/poetry/releases/tag/2.0.0>. Syntax of pyproject.toml and changes in CLI.

> Just make a venv, which helps you keep your projects separate.

I'm not so fluent in python and it's ecosystem. I know some basic stuff with `poetry`, but it doesn't work now, so I can't create the virtual environment. 

There is just `source .venv/bin/activate   <-- Activate the virtual environment` in the main `README`, which doesn't help me much.

Any help would be appreciated.

janiversen:
Search for information, it is really basic python knowledge. we do not have bandwidth to help with python education

PR #2671, opened 2025-05-22
transaction: add number of retries attempted to return calls

mdeneen:
If a transaction is successfully processed, return result along with the number of retries.  This can be useful for application layer logic which may want to know that retries are happening.

### DIFF
--- a/pymodbus/client/base.py
+++ b/pymodbus/client/base.py
@@ -14,7 +14,7 @@ from pymodbus.transaction import TransactionManager
 from pymodbus.transport import CommParams
 
 
-class ModbusBaseClient(ModbusClientMixin[Awaitable[ModbusPDU], Awaitable[int]]):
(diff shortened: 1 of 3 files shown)


mdeneen:
Ah, I see that I need to adjust the return types.  Give me a few minutes.

mdeneen:
Can you offer some guidance on the python typing?  I think that you will understand what I am going for here.

janiversen (changes requested):
This really breaks a lot, as you can see in the CI run. 

We do not want to return a tuple, because it would be a lot easier just to add a field in the response, with the added benefit that it is backwards compatible.

mdeneen:
I see that this will impact existing code too much, so I am going to close this request.

mdeneen:
Hey, Jan, I will look at adding it to the response.

PR #2672, opened 2025-05-22
Add retries to ModbusPDU class

mdeneen:
I tried doing this before by adding another return value, but it would break many things.

This time I have added an integer representing the number of retries to the ModbusPDU class and setting that value in both the async and sync transaction execute code path.

### DIFF
--- a/pymodbus/pdu/pdu.py
+++ b/pymodbus/pdu/pdu.py
@@ -38,6 +38,7 @@ class ModbusPDU:
         self.status: int = status
         self.exception_code: int = 0
         self.fut: asyncio.Future
+        self.retries: int = 0
 
     def isError(self) -> bool:
         """Check if the error is a success or failure."""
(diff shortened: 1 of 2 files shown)


mdeneen:
Hey, Jan, one downside to this is that if you enable logging it will show retries as zero with the string representation of the PDU, since the decoder action happens _before_ the retry is set.

mdeneen:
Other things, like device id, are also incorrect in the trace.  I guess I am okay with the way it currently is.

janiversen:
device id is correctly 0 in the response, signaling broadcast... but in general it might be better to remove all variables and just return the class name.

mdeneen:
You're right, I am using them solely from the client's perspective, my apologies.

In my case, the device I am communicating with has a device id of 2.  When I read holding registers I must set device_id (or slave, for older versions) to 2.  The debug trace shows 0, but the actual returned object is set to 2.

Am I look at this incorrectly?

janiversen:
In the client you know what you requested (since you can only do one request at a time) so it does not really matter what is being returned.

PR #2686, opened 2025-07-14
DummyProtocol is not async

alexrudd2:
I found this tiny fix on an old laptop; not sure what tool found it :)


janiversen (approved):
Good catch.

PR #2720, opened 2025-08-08
Simplify code flow for broadcast requests

alexrudd2:
A small simplification - return early in the broadcast case and don't bother with `response`.

(`pyright` had a false positive here, but it leads to simpler code)


alexrudd2:
> What was wrong with the old way ??? it seems more efficient to use asyncio.gather

I'm not sure, but it was causing the test failure so there must be some race condition?  (My local `pytest` was fine...)

I can investigate in a separate PR.

PR #2722, opened 2025-08-08
(CI only) run CI with asyncio.gather and broadcast change

alexrudd2:
https://github.com/pymodbus-dev/pymodbus/pull/2721

NOT TO BE MERGED.

### DIFF
--- a/pymodbus/server/requesthandler.py
+++ b/pymodbus/server/requesthandler.py
@@ -92,11 +92,8 @@ class ServerRequestHandler(TransactionManager):
             if self.server.broadcast_enable and not self.last_pdu.dev_id:
                 # if broadcasting then execute on all device contexts,
                 # note response will be ignored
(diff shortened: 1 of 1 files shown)


janiversen:
Just for info, you can run CI manually on a branch via the actions menu...saves some typing.

alexrudd2:
> Just for info, you can run CI manually on a branch via the actions menu...saves some typing.

Yes, I tried that - but the way things are setup that still keeps the fail draft job which fails everything.

Anyways, I have some other ideas for CI in progress.

janiversen:
The fail draft job was there for a reason (you too know why), and it might be time that we retire it.

I look forward to see if you have more changes than using uv, which is a nice idea.

alexrudd2:
The original failing test doesn't reproduce, but also... I don't care hahah

PR #2723, opened 2025-08-08
Switch from venv+pip to uv

alexrudd2:
[`uv` ](https://docs.astral.sh/uv/) is to `pip` and `venv` what `ruff` is to `pyflakes` and `black` - faster and better in every way.

It greatly speeds up dependency resolution and install, and eliminates the fuss of activating venv on Windows (which I remember was quite `fun` to setup for @janiversen ... )

The dependency setup with a cache miss is <2s, which is nearly as fast as a cache hit with the existing solution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,9 +69,9 @@ jobs:
         run: |
           uv run codespell --skip "*.lock"
 
-      # - name: documentation
-      #   if: matrix.run_doc == true
-      #   run: cd doc && ./build_html
+      - name: documentation
+        if: matrix.run_doc == true
+        run: cd doc && ./build_html
 
       - name: pylint
         if: matrix.run_lint == true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,11 +54,6 @@ jobs:
       - name: Checkout repo from github
         uses: actions/checkout@v4.2.2
 
-      - name: Set up Python ${{ matrix.python }}
-        uses: actions/setup-python@v5.4.0
-        with:
-          python-version: ${{ matrix.python }}
-
       - name: Install uv
         uses: astral-sh/setup-uv@v6
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,11 @@ jobs:
 
       - name: documentation
         if: matrix.run_doc == true
-        run: cd doc && uv ./build_html
+        run: |
+          source .venv/bin/activate
+          cd doc 
+          ./build_html
+        # this won't work on Windows, but we run_doc == False on Windows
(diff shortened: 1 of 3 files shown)


janiversen (commented):
Looks like a nice change, but it seems uv is not ready to meet codespell.

janiversen:
Feel free to exclude lock files in pyproject.toml.

.github/workflows/ci.yml
```diff
@@ -72,7 +72,7 @@ jobs:
       - name: codespell
         if: matrix.run_doc == true
         run: |
-          uv run codespell
+          uv run codespell --skip "*.lock"
```
janiversen, line 70: This should be in pyproject.toml, otherwise check_ci.sh will not work identical to CI, which is kind of it´s purpose.

janiversen (changes requested):

doc/build_html
```diff
@@ -5,4 +5,10 @@ zip -9 doc/source/_static/examples.zip examples/*
 tar -czf doc/source/_static/examples.tgz examples
 popd
 
-sphinx-build -M html "." "../build/html" 
+if command -v uv &> /dev/null; then
```
janiversen, line 8: This change could be made a lot easier in ci.yml line 74 without any side effects.

janiversen (changes requested):

doc/build_html
alexrudd2: (1) No, since the call to `sphinx-build` is in `build_html`.  It would fail without `uv run` (or activating the venv)
(2) By doing it this way it can work either with or without `uv`.  

I'm not ready to propose switching over the entire project yet.  (e.g. `python -m build` -> `uv build` and some `pyproject.toml` changes)

janiversen: if you out uv in front of build_html it should activate the venv which will then be handed to sphinx.

janiversen: Yeah I expected you to try before writing, however if I run "workon" before calling build_html it uses the environment assigned, so I do not understand what the problem is with running "uv" before the build_html command. venv is normally inherited across shell sessions.

your python_env function have one floor, I have at the moment 8 different venv, and when I login to my vm, I start by "workon <current project>".

alexrudd2: Interesting, this is my first hearing of `workon`.

For the CI, `uv sync` _creates_ a venv, but does not activate it.  So I had to activate it.

I think this is resolved to your satisfaction?  `build_html` is untouched compared to `dev`.

janiversen (approved):
Thanks a lot, this seems to be a very GOOD solution.

I am sorry, that I had to "correct" you, your PR´s and comments are HIGHLY valued, but I also need to think of the consequences.

I am sure you will find an elegant solution for build_html as well.

PR #2730, opened 2025-08-09
Ignore lockfile

alexrudd2:
I'm using `uv` but don't want to push that on everybody else.  (yet :)  )

By doing `*.lock` instead of `uv.lock` it might also help someone using e.g. `poetry` or `rye`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ build/html
 venv
 downloaded_files/
 pymodbus.log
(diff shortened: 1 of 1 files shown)


alexrudd2:
> you need quite a lot of convincing before the scripts get converted.

Soon!  I was convinced within one hour of trying it  :)  `pip` has no chance.

PR #2731, opened 2025-08-09
Stricter types with pyright

alexrudd2:
`pyright` is another type-checker that finds some things that `mypy` does not.

Currently there are 42 errors in `dev`, which this PR reduces to ~6~.  EDIT: 32.

Several of them are false positives from imports and `byte`/`bytearray` confusion.
However, many are mismatched type signatures between classes and subclasses that are easy enough to improve.

I have split different types of changes into different commits so they can be cherry-picked or reverted if desired.


pymodbus/datastore/simulator.py
```diff
@@ -581,16 +585,16 @@ def validate(self, func_code, address, count=1):
         fx_write = func_code in self._write_func_code
         return self.loop_validate(real_address, real_address + count, fx_write)
 
-    def getValues(self, func_code, address, count=1) -> list[int] | list[bool] | ExcCodes:
+    def getValues(self, fc_as_hex, address, count=1) -> list[int] | list[bool] | ExcCodes:
```
janiversen, line 588: I do believe func_code is an int, not a hex string.

janiversen: This is general for the rest of the file.

pymodbus/pdu/diag_message.py
```diff
@@ -65,7 +65,7 @@ def get_response_pdu_size(self) -> int:
         """
         return 1 + 2 + 2
 
-    async def update_datastore(self, _context: ModbusDeviceContext) -> ModbusPDU:
+    async def update_datastore(self, context: ModbusDeviceContext) -> ModbusPDU:
```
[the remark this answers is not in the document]
janiversen: If pyright complains that the variable have a different name, then that should be silenced, because it is perfectly legal.

pymodbus/server/requesthandler.py
```diff
@@ -40,23 +40,23 @@ def __init__(self, owner, trace_packet, trace_pdu, trace_connect):
             trace_connect,
         )
 
-    def callback_new_connection(self) -> ModbusProtocol:
+    def callback_new_connection(self):
```
janiversen, line 43: This is wrong, the callback returns ModbusProtocol.

pymodbus/transport/serialtransport.py
```diff
@@ -5,10 +5,13 @@
 import contextlib
 import os
 import sys
+from typing import TYPE_CHECKING
 
 
 with contextlib.suppress(ImportError):
     import serial
+if TYPE_CHECKING:
```
janiversen, line 13: This seems wrong, serial is already imported.

pymodbus/transport/transport.py
```diff
@@ -604,7 +604,7 @@ def close(self) -> None:
         if self.protocol:
             self.protocol.connection_lost(None)
 
-    def sendto(self, data: bytes, _addr: Any = None) -> None:
+    def sendto(self, data: bytes, addr: Any = None) -> None:
```
janiversen, line 607: addr is unused hence _addr.

pymodbus/client/serial.py
```diff
@@ -17,6 +18,8 @@
 
 with contextlib.suppress(ImportError):
     import serial
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
alexrudd2: `pyright` is stricter than `mypy` and considers that the `ImportError` could be suppressed.
`/home/alex/git/pymodbus/pymodbus/transport/serialtransport.py:29:28 - error: "serial" is possibly unbound (reportPossiblyUnboundVariable)`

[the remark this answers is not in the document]
janiversen: But that is exactly the functionality we want ! 

Either pyright suggest an elegant solution or please disable reportPossiblyUnboundVariable in this case.

alexrudd2:
I fully expected some pushback on the `_unused_arg` / `unused_arg` so they're in separate commits.

Alternatives:
```python

class Base:
    def method(arg):
        return arg

# this PR - keep identical signatures
class Derived:
    def method(arg):
        pass

# explicit suppress
class Derived_ignore:   # type:ignore[reportIncompatibleMethodOverride]   
    def method(_arg):   
        pass

# fake usage
class Derived_assign:
    def method(arg):
        _ = arg

# unpacking
class Derived_kwargs:
   def method(*args, **kwargs):
      pass
```

pymodbus/server/requesthandler.py
alexrudd2: I tried that, but there is one test that uses what should be an abstract method.  Actually, I think it may be possible to entirely remove this method but that's for another PR.

janiversen:
But this is wrong !! it does not make sense to demand that positional parameters have the same name in inherited classes....I would consider that a general false positive.

pymodbus/client/serial.py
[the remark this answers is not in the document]
alexrudd2: It's about a dozen places to suppress, unfortunately...

I thought the `if TYPE_CHECKING` _was_ an elegant solution, haha.  Perhaps there is some other way.

alexrudd2:
Anyways, I have removed the contentious commits for another day.  Do you agree with [c3ed1b1](https://github.com/pymodbus-dev/pymodbus/pull/2731/commits/c3ed1b100ccffe7c4208a0364c1839bbca3cb210) or should I revert that as well?
https://github.com/pymodbus-dev/pymodbus/pull/2731#discussion_r2264933956

pymodbus/server/requesthandler.py
janiversen: Totally remove this method is interesting, I do not really see how since the content depend on the type of transport.

You could make it abstract and modify that one test.

alexrudd2: For another day; reverted.

janiversen:
I think you made that commit a separate PR, which I have approved.

janiversen:
Let me take a look.

janiversen (approved):
LGTM, thanks.

PR #2732, opened 2025-08-09
Hypothesis property-based testing demo

alexrudd2:
Just a demo for now; I want to find some more complex testing that `hypothesis` would be good at.

[`hypothesis`](https://hypothesis.readthedocs.io/en/latest/index.html) is a property-based testing library.  It does semi-automatic property-based testing to find edge cases easier than manual test cases.  I think this would be especially useful for testing encoding/decoding or various message roundtrips.

### DIFF
--- a/test/hypothesis_test.py
+++ b/test/hypothesis_test.py
@@ -1,0 +1,38 @@
+"""Property-based tests for Modbus payload round-trip encoding/decoding with randomized word order."""
+import pytest
+from hypothesis import given, settings
+from hypothesis import strategies as st
+
+from pymodbus.client.mixin import ModbusClientMixin as mixin
+
+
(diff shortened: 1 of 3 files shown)


test/hypothesis_test.py
```diff
@@ -0,0 +1,38 @@
+"""Property-based tests for Modbus payload round-trip encoding/decoding with randomized word order."""
+import pytest
+from hypothesis import given, settings
+from hypothesis import strategies as st
+
+from pymodbus.client.mixin import ModbusClientMixin as mixin
+
+
+word_order_strategy = st.one_of(st.just("big"), st.just("little"))
```
alexrudd2, line 9: similar to `@pytest.mark.parameterize` but it will mix in the big/little with testing a range of values.

PR #2734, opened 2025-08-09
Use asyncio.Event to deterministically ensure simulator start

alexrudd2:
TIL about `asyncio.Event`.  In this case, it's both faster and more deterministic than `asyncio.sleep(1)`


janiversen (approved):
LGTM, thanks.

PR #2748, opened 2025-08-23
Missing coma in string representation of ModbusPDU

MrWaloo:
Current output of 
```python
print(f"{anyFunctionInstance}")
```
is (for example):
```
ReadHoldingRegistersRequest(dev_id=1, transaction_id=0, address=12503, count=7, bits=[], registers=[], status=1retries=0)
```
A coma is missing between `status` and `retries`, it should be:
```
ReadHoldingRegistersRequest(dev_id=1, transaction_id=0, address=12503, count=7, bits=[], registers=[], status=1, retries=0)
```

This humble PR corrects it.


janiversen (approved):
LGTM, thanks

Issue #2768, opened 2025-09-09
reconnect_delay: seconds or milliseconds ?


wlcrs:
The reconnect logic is broken due to a mixup of units: in the constructors of the clients you document it as being in milliseconds:

https://github.com/pymodbus-dev/pymodbus/blob/8c9c152ff14b32c5f5f6ae41f0cadc815987d2ae/pymodbus/client/tcp.py#L31-L32

However, you interpret it as being in seconds in the transport:

https://github.com/pymodbus-dev/pymodbus/blob/8c9c152ff14b32c5f5f6ae41f0cadc815987d2ae/pymodbus/transport/transport.py#L464-L480

Side question: why is `reconnect_delay_current` not just a local variable to the `do_reconnect` function? This would lead to cleaner code imho.

janiversen:
"in the constructors of the clients you document it as being in milliseconds:" that is not true!

The signature states:
```
:param reconnect_delay: Minimum delay in seconds.milliseconds before reconnecting.
    :param reconnect_delay_max: Maximum delay in seconds.milliseconds before reconnecting.
```
Which is correct seconds.milliseconds, that so 1.3 is 1 second and some milliseconds. 

I do not see where the code is broken, however if you think it is broken, then feel free to submit a PR.

Having reconnect_delay_current as a local variable, assures it cannot be used between do_reconnect calls, however if you want it to work differently, then feel free to submit a PR.

PR #2769, opened 2025-09-09
Clarify documentation on reconnect_delay

wlcrs:
A fix for #2768: the syntax of `seconds.milliseconds` just does not make sense. As the arguments are typed as floats it is clear that you can pass fractional values.

### DIFF
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -35,9 +35,9 @@ class AsyncModbusSerialClient(ModbusBaseClient):
     :param stopbits: Number of stop bits 1, 1.5, 2.
     :param handle_local_echo: Discard local echo from dongle.
     :param name: Set communication name, used in logging
-    :param reconnect_delay: Minimum delay before reconnecting, in seconds.
-    :param reconnect_delay_max: Maximum delay before reconnecting, in seconds.
-    :param timeout: Timeout for connecting and receiving data, in seconds.
+    :param reconnect_delay: Minimum delay when reconnecting, in seconds (use decimals for milliseconds).
+    :param reconnect_delay_max: Maximum delay when reconnecting, in seconds (use decimals for milliseconds).
+    :param timeout: Timeout for connecting and receiving data, in seconds (use decimals for milliseconds).
     :param retries: Max number of retries per request.
     :param trace_packet: Called with bytestream received/to be sent
     :param trace_pdu: Called with PDU received/to be sent
--- a/pymodbus/client/tcp.py
+++ b/pymodbus/client/tcp.py
@@ -28,9 +28,9 @@ class AsyncModbusTcpClient(ModbusBaseClient):
     :param port: Port used for communication
     :param name: Set communication name, used in logging
     :param source_address: source address of client
-    :param reconnect_delay: Minimum delay before reconnecting, in seconds.
-    :param reconnect_delay_max: Maximum delay before reconnecting, in seconds.
-    :param timeout: Timeout for connecting and receiving data, in seconds.
+    :param reconnect_delay: Minimum delay when reconnecting, in seconds (use decimals for milliseconds).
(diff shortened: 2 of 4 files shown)


pymodbus/client/serial.py
```diff
@@ -35,8 +35,8 @@ class AsyncModbusSerialClient(ModbusBaseClient):
     :param stopbits: Number of stop bits 1, 1.5, 2.
     :param handle_local_echo: Discard local echo from dongle.
     :param name: Set communication name, used in logging
-    :param reconnect_delay: Minimum delay in seconds.milliseconds before reconnecting.
-    :param reconnect_delay_max: Maximum delay in seconds.milliseconds before reconnecting.
+    :param reconnect_delay: Minimum delay in seconds before reconnecting.
```
janiversen, line 38: This does not say that fractions are accepted ! We have had issues with a very similar text, where users said it was not clear that milliseconds could be used.

janiversen: I do not mind a new wording, but it needs to state what is possible, not everybody make the connection between seconds and a float.

janiversen (changes requested):

pymodbus/client/serial.py
janiversen: Ups forgot this change should be done for all client types, since it is the same parameter.

wlcrs: This change brought it in line with how the `timeout` parameter is documented. I've also never seen it being documented differently in another library, hence my confusion when I misread it the way it is currently written.

I've now added (float) after each parameter description, including the 'timeout' value where this also was not specified.

pymodbus/client/serial.py
```diff
@@ -35,9 +35,9 @@ class AsyncModbusSerialClient(ModbusBaseClient):
     :param stopbits: Number of stop bits 1, 1.5, 2.
     :param handle_local_echo: Discard local echo from dongle.
     :param name: Set communication name, used in logging
-    :param reconnect_delay: Minimum delay in seconds before reconnecting.
-    :param reconnect_delay_max: Maximum delay in seconds before reconnecting.
-    :param timeout: Timeout for connecting and receiving data, in seconds.
+    :param reconnect_delay: Minimum delay before reconnecting, in seconds (float).
```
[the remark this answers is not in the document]
wlcrs: The current wording and the one that you are proposing now are both incorrect: the part behind the comma is a fraction of a second.

If you interpret the current wording `seconds.milliseconds` literally, 5 milliseconds can be written as `0.5` and 25 milliseconds as `0.25` , which is of course incorrect.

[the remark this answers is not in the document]
wlcrs: >, but (float) does not add useful information.

At my university, data types are one of the first concepts covered in the Python Programming 101 course. "float" is a concept that every Python programmer should be familiar with: it clearly tells the programmer that the value is a floating-point number, or to be mathematically precise: a *real* number.

> 10 times (fraction of a second) is millisecond

I don't understand what you mean. This sentence does not make any sense.

> but I am not trying to be precise, just give users a hint, that milliseconds can be used.

they can use any fraction of a second, not only milliseconds. Your 'hint' is currently confusing at best. 

> Your issue was a good example you thought it was only milliseconds even though you knew it was a float.

That is not a fair example. I started using this feature in my own library when migrating to pymodbus 3.0.0 . The documentation then read:

https://github.com/pymodbus-dev/pymodbus/blob/d7834aca6d80cd93d2b97ce00e867e52ee64ff61/pymodbus/client/base.py#L34C5-L34C82

Your API_CHANGES.rst never mentioned that this unit had been changed!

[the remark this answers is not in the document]
janiversen: But as you pointed out, the type is float, so adding (float) does not give extra information. You could still describe the decimal part, which is milliseconds, I leave wording to you.

PR #2826, opened 2025-12-08
Fix monitoring of functions

alexrudd2:
Caught by `pyright` :)

### DIFF
--- a/pymodbus/server/simulator/http_server.py
+++ b/pymodbus/server/simulator/http_server.py
@@ -7,6 +7,7 @@ import dataclasses
 import importlib
 import json
 import os
+from typing import TYPE_CHECKING
 
 
 with contextlib.suppress(ImportError):
@@ -24,6 +25,9 @@ from pymodbus.server.server import (
     ModbusUdpServer,
 )
 
+
+if TYPE_CHECKING:
+    from aiohttp import web
 
 MAX_FILTER = 1000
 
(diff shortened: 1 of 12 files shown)


pymodbus/server/simulator/http_server.py
```diff
@@ -372,10 +372,10 @@ def build_html_calls(self, params: dict, html: str) -> str:  # pragma: no cover
         for function in DecodePDU(True).list_function_codes():
             selected = (
                 "selected"
-                if function.function_code == self.call_monitor.function
+                if function == self.call_monitor.function
                 else ""
             )
-            function_codes += f"<option value={function.function_code} {selected}>function code name</option>"
+            function_codes += f"<option value={function} {selected}>function code name</option>"
```
janiversen: This is wrong, you are now building a list of object pointers and not function codes !

pymodbus/server/simulator/http_server.py
```diff
@@ -372,10 +372,10 @@ def build_html_calls(self, params: dict, html: str) -> str:  # pragma: no cover
         for function in DecodePDU(True).list_function_codes():
             selected = (
                 "selected"
-                if function.function_code == self.call_monitor.function
+                if function == self.call_monitor.function
```
janiversen: This seems wrong, it is a list of function codes, not object pointers.

janiversen (changes requested):
This one seems wrong.

PR #2830, opened 2025-12-17
Fix Modbus TCP protocol ID validation in FramerSocket

starwarsivan:
### Summary

This PR adds validation for the Modbus TCP Protocol ID in `FramerSocket`
and ensures only compliant frames (Protocol ID = 0) are accepted.

### Changes

- Added a Protocol ID check in `FramerSocket` to reject non-zero values
- Updated existing test vectors to use Protocol ID = 0, per the Modbus TCP specification
- Added a test to verify that invalid (non-zero Protocol ID) messages produce empty results

### Rationale

According to the Modbus TCP specification, the Protocol ID field must be
set to `0`. Previously, non-zero values were accepted and some tests used
invalid protocol IDs, which could mask malformed input.

This change enforces spec compliance and improves robustness when handling
invalid Modbus TCP frames.

### Testing

- Updated unit tests pass
- New test confirms invalid Protocol ID frames are rejected

### DIFF
--- a/test/framer/test_framer.py
+++ b/test/framer/test_framer.py
@@ -473,7 +473,6 @@ class TestFramerType:
 
     def test_invalid_protocol_id_for_framer_socket(self):
         """Test that ModbusSocketFramer rejects an invalid Protocol ID."""
-        from pymodbus.framer.socket import FramerSocket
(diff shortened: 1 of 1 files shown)


janiversen (approved):
Thanks.

PR #2842, opened 2026-01-19
di is for discrete input

MrWaloo:
"direct input" is not correct


MrWaloo:
Somehow the line end was different on my repo clone, it should be solved now

PR #2853, opened 2026-01-29
Simulator DataBlock docstring corrections

MrWaloo:
Some corrections according to the code

### DIFF
--- a/pymodbus/transport/transport.py
+++ b/pymodbus/transport/transport.py
@@ -329,6 +329,8 @@ class ModbusProtocol(asyncio.BaseProtocol):
             if not data:
                 return
         Log.transport_dump(Log.RECV_DATA, data, self.recv_buffer)
+        if len(self.recv_buffer) > 1024:
+            self.recv_buffer = b''
         self.recv_buffer += data
         cut = self.callback_data(self.recv_buffer, addr=addr)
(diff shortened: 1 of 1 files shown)


pymodbus/simulator/simdevice.py
```diff
@@ -20,7 +20,7 @@ class SimDevice:
 
     Some old devices uses 4 distinct blocks instead of 1 block, to
     support these devices, define 1 large block consisting of the
-    4 blocks and use the offset_*= parameters.
+    4 blocks and use the offset_address= parameter.
```
janiversen: That is not the name of the parameter.

pymodbus/simulator/simdevice.py
```diff
@@ -37,11 +37,7 @@ class SimDevice:
         SimDevice(
             id=1,
             registers=[SimData(...)],
-            non_shared_mode=True,
-            offset_coil=0,
-            offset_discrete=10,
-            offset_holding=20,
-            offset_input=30,
+            offset_address=(0, 10, 20, 30),
```
janiversen: That is not very user friendly, since which block is which parameter is only indirectly documented.

janiversen (approved):
My bad, I currently have another version active.

LGTM, Thanks.

Issue #2872, opened 2026-02-09
[dev] Fix premature shutdown in simulator example and update README instructions


FedericoMusa:
Versions
Python: 3.13

OS: Windows 10 (MINGW64)

Pymodbus: 4.0.0.dev8 (branch dev)

Modbus Hardware: None (Simulator)

Pymodbus Specific
Server: tcp - async (Simulator)

Client: tcp - async (Simulator)

Description
I am reporting an inconsistency in the current dev branch regarding the simulator's usability as a standalone tool.

Premature Shutdown in Example: The examples/simulator.py script is currently structured as a one-shot test. As seen in the attached code snippet, it calls await task.stop() immediately after the test calls, which triggers a graceful shutdown of both the Modbus and HTTP servers.

Web UI Inaccessibility: Because of this immediate shutdown, the Web UI (listening on port 8081) becomes unreachable for manual interaction.

Documentation Gap: The README.rst still points to the deprecated v3_simulator.py script, which no longer exists in the examples/ directory.

Expected behavior: The simulator example should remain persistent to allow users to interact with the Web interface. I have verified that replacing task.stop() with a persistent wait allows the Web UI to function correctly.

Code and Logs
Execution Log (Premature Shutdown):

Plaintext
2026-02-09 17:20:29,987 INFO  base:89 Server listening.
2026-02-09 17:20:30,180 INFO  http_server:226 Stopping modbus server
2026-02-09 17:20:30,181 INFO  base:93 Server graceful shutdown.
2026-02-09 17:20:30,181 INFO  http_server:232 Modbus server Stopped

Code causing the issue (examples/simulator.py):

<img width="657" height="140" alt="Image" src="https://github.com/user-attachments/assets/eee2003d-6ba5-41ce-85df-24d6a7e5b81c" />

<img width="811" height="384" alt="Image" src="https://github.com/user-attachments/assets/39dda3de-f644-4dfe-80a9-3f5df394541c" />

<img width="738" height="435" alt="Image" src="https://github.com/user-attachments/assets/6a029363-0b74-42c3-a08c-ff38ad0cea08" />

janiversen:
Thanks for the report.

I have now reread the README.rst and cannot find a link to any simulator script, can you please point out where the link is so we can fix it.

Examples are just that, they are meant to be adapted to your needs ! simulator.py shows how to start the simulator, connect a client, and stop it...full circle.

The installation adds pymodbus.simulator as a script, which you can run with your own configuration.

[...]

PR #2873, opened 2026-02-10
Docs/simulator official entrypoint

FedericoMusa:
This PR addresses the documentation gap discussed in issue #2872. It clarifies the use of the official entry point for a persistent simulator experience.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -268,6 +268,14 @@ so that others can benefit from the results.
 
 `Example documentation <https://pymodbus.readthedocs.io/en/dev/source/examples.html>`_
 
+Ready to go simulator
+^^^^^^^^^^^^^^^^^^^^^
+To start the simulator and explore the Web UI, run the following command from the root directory::
+
+    python examples/simulator.py
+
+Once started, the Modbus server will listen on port 5020 and the Web UI will be available at http://localhost:8081.
+
 Troubleshooting the dev branch
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 If you encounter errors while running examples, please check:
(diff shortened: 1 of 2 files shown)


janiversen:
Just checked it does start (somebody added a default configuration).

I believe the best way to document it are to show the allowed parameters in README.rst, otherwise it is easy to see the next issue "why does it not work when my client request data".

README.rst
```diff
@@ -268,6 +268,14 @@ so that others can benefit from the results.
 
 `Example documentation <https://pymodbus.readthedocs.io/en/dev/source/examples.html>`_
 
+Ready to go simulator
+^^^^^^^^^^^^^^^^^^^^^
+The simulator can be started directly using the installed entry point::
+
+    pymodbus.simulator
+
+Once started, the Modbus server will listen on port 5020 and the Web UI will be available at http://localhost:8081. This command provides a persistent instance for manual testing.
```
janiversen: But you need to add configuration to start it properly. It actually surprises me that it starts without configuration.

janiversen (changes requested):

README.rst
FedericoMusa: I've updated the README with the parameter table and a note about the default behavior.

Regarding the 'silent start' surprise: I investigated the source and found that in pymodbus/server/simulator/main.py (line 86), the --json_file parameter defaults to an internal setup.json within the package directory. This explains why it starts even without a local config.
Beyond documentation, should we 'armour' this entry point? I could add a logic check in that same main.py to log a heavy warning or even exit if a user-defined configuration isn't explicitly found.

Let me know what you think of the new docs and this potential code fix!
image


image


image

janiversen (approved):
LGTM, thanks.

PR #2894, opened 2026-02-24
Fix ReportDeviceIdResponse

pnb990:
refs #2891

### DIFF
--- a/pymodbus/pdu/other_message.py
+++ b/pymodbus/pdu/other_message.py
@@ -196,10 +196,9 @@ class ReportDeviceIdResponse(ModbusPDU):
     def encode(self) -> bytes:
         """Encode the response."""
         status = ID_ON if self.status else ID_OFF
-        length = len(self.identifier) + 1
-        packet = struct.pack(">B", length)
+        length = len(self.identifier) + 2
+        packet = struct.pack(">BBB", length, dev_id, status)
         packet += self.identifier  # we assume it is already encoded
-        packet += struct.pack(">B", status)
         return packet
 
     def decode(self, data: bytes) -> None:
@@ -208,10 +207,11 @@ class ReportDeviceIdResponse(ModbusPDU):
         Since the identifier is device dependent, we just return the
         raw value that a user can decode to whatever it should be.
         """
-        self.byte_count = int(data[0])
-        self.identifier = data[1 : self.byte_count + 1]
-        status = int(data[-1])
+        self.byte_count, self.dev_id, status = struct.unpack(">BBB",
+                                                                  data[:3])
         self.status = status == ID_ON
+        self.identifier = data[3:3 + self.byte_count - 1]
+
 
 DecodePDU.add_pdu(ReadExceptionStatusRequest, ReadExceptionStatusResponse)
 DecodePDU.add_pdu(GetCommEventCounterRequest, GetCommEventCounterResponse)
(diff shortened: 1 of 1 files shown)


janiversen (changes requested):
LGTM, thanks.

However it seems you did not change the test cases, and therefore CI fails, please correct.

pymodbus/pdu/other_message.py
```diff
@@ -196,10 +196,9 @@ def __init__(self, identifier: bytes = b"\x00", status: bool = True, dev_id: int
     def encode(self) -> bytes:
         """Encode the response."""
         status = ID_ON if self.status else ID_OFF
-        length = len(self.identifier) + 1
-        packet = struct.pack(">B", length)
+        length = len(self.identifier) + 2
+        packet = struct.pack(">BBB", length, self.dev_id, status)
```
janiversen, line 200: This do NOT correspond to the modbus standard:

which defines:
- 1 byte count
- x bytes Server id (device dependent, here it is self.identifier
- 1 byte status
- Additional data (the pymodbus server do not have extra data)

self.dev_id is the byte before the function code in all modbus messages.

pymodbus/pdu/other_message.py
```diff
@@ -208,10 +207,11 @@ def decode(self, data: bytes) -> None:
         Since the identifier is device dependent, we just return the
         raw value that a user can decode to whatever it should be.
         """
-        self.byte_count = int(data[0])
-        self.identifier = data[1 : self.byte_count + 1]
-        status = int(data[-1])
+        self.byte_count, self.dev_id, status = struct.unpack(">BBB",
```
janiversen, line 210: This do NOT respond to the modbus standard, see above.

The old code allowed for a variable Server ID, which is demanded by the standard (Page 31).

janiversen (changes requested):
Sorry my previous approve was not correct.

Some of your changes do not correspond to the modbus standard, which pymodbus strictly adheres to.

the old line 213 which referred to data[-1] is actually correct since it is the last data byte.

Once you have made the needed changes, please do not forget to update the test harness as well, we cannot merge without a Green light from CI.

janiversen:
Since the length of Server ID is "device dependent" and there can be additional data after the status, it is impossible to determine the location of status. So the most correct way to handle it would be:

to let decode() simply return all bytes, and let the app determine how to interpret the data.

encode() represent the pymodbus server, meaning we define the bytes and have defined that
there are no "additional data", hence status is the last byte.

pnb990:
Yes, I was read to quickly, Name is "Slave ID" I believed it was like Address (slave, device, unit)  and it is of one byte.
But your right, it is strange for modbus standard that leave this device specific size before a standard value...
So i f you want keep something in my PR tell-me otherwise close it.

janiversen:
I gave you guidelines to what I think would be the best solution.

But you might want to do it differently....

If you submit something I will review it.

Btw device id == dev_id == slave_id

but device ident is typically a string with the name of the device.

janiversen:
If you do not want to work further on this PR, then please close it. In FOSS it is considered bad practice to close PR/Issues from others.

As I wrote suggestions are still more than welcome, and if this is based on a real problem, then please submit a debug log so we can analyze it.

GitHub CI closes old abandoned PRs and then reports it, which might have a negative effect on the issuer.

pnb990:
I m using home made device, so I will fix my device to match modbus specification

PR #2898, opened 2026-02-25
Fix Read Fifo Query  RTU Frame Size

brambo123:
Found a small calculation error in the Read FIFO Query function.
This should fix it, unless a test fails again.

### DIFF
--- a/pymodbus/pdu/file_message.py
+++ b/pymodbus/pdu/file_message.py
@@ -260,7 +260,7 @@ class ReadFifoQueueResponse(ModbusPDU):
         """Calculate the size of the message."""
         hi_byte = int(data[2])
         lo_byte = int(data[3])
-        return ((hi_byte << 16) + lo_byte) * 2 + 6
+        return ((hi_byte << 16) + lo_byte) + 6
 
     def __init__(self, values: list[int] | None = None, dev_id: int = 1, transaction_id:int  = 0) -> None:
         """Initialize a new instance."""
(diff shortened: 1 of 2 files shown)


janiversen (changes requested):
LGTM, thanks.

Please correct the test harness as we cannot merge a PR without a greenwich CI.

brambo123:
https://github.com/pymodbus-dev/pymodbus/blob/54d35b4e48e7e2fa3f1ed8d9b68a386d8a3934c4/test/pdu/test_file_message.py#L23
https://github.com/pymodbus-dev/pymodbus/blob/54d35b4e48e7e2fa3f1ed8d9b68a386d8a3934c4/test/pdu/test_file_message.py#L47-L65

Can someone explain to me what input calculateRtuFrameSize should actually have?
I think it should have the full frame, including the device ID and function code.
But TEST_MESSAGE seems to only be the output of encode(), which I don't think is a full frame?

Edit:
I made a quick change to make it a full frame, and that seems to work.
But I'm not sure if it's correct, and it might not be the most prettiest way to do it.

janiversen:
I am confused about your question, check the code:

pymodbus/framer/rtu.py line 98.

why would you assume then encode generates what decode wants..it is used in 2 very different situations...but again just look at the code.

janiversen (approved):
LGTM, Thanks.

janiversen:
Just saw that the test file have a copy/paste error...all the references to bit/coil/discrete input should clearly have been to file...

This includes the class name.

Feel free to submit a follow up PR if you want to.

PR #2903, opened 2026-03-12
Fix usage docs in examples: add missing option --baudrate

KevinWMatthews:
The usage text in several examples is missing the option `--baudrate`, even though:
- the option is still available as a CLI argument
- the explanation for this option is still present


janiversen (approved):
LGTM, thanks.

PR #2921, opened 2026-04-14
fix: guard short frames and validate mbap pid in TLS framer

MrAlaskan:
## Root Cause
FramerTLS.decode() had two parsing issues: it did not validate MBAP PID (so non-zero PID frames were still treated as normal), and it accessed fixed positions (like data[6]) without checking minimum frame length first, which caused an index error on short input.

## Fix
 I updated FramerTLS.decode() to add basic frame validation before decoding: it now checks minimum length first, validates that MBAP PID is 0 , and verifies the MBAP length before reading frame fields. If data is incomplete or invalid, it returns an empty decode result instead of continuing to parse.

### DIFF
--- a/pymodbus/framer/tls.py
+++ b/pymodbus/framer/tls.py
@@ -20,9 +20,18 @@ class FramerTLS(FramerBase):
 
     def decode(self, data: bytes) -> tuple[int, int, int, bytes]:
         """Decode MDAP+payload."""
+        if (data_len := len(data)) < self.MIN_SIZE:
+            return 0, 0, 0, self.EMPTY
         tid = int.from_bytes(data[0:2], 'big')
+        if int.from_bytes(data[2:4], 'big'):
+            return 0, 0, 0, self.EMPTY
+        msg_len = int.from_bytes(data[4:6], 'big') + 6
         dev_id = int(data[6])
-        return len(data), dev_id, tid, data[7:]
+        if data_len < msg_len:
+            return 0, 0, 0, self.EMPTY
+        if msg_len == 8 and data_len == 9:
+            msg_len = 9
+        return msg_len, dev_id, tid, data[7:msg_len]
 
     def encode(self, payload: bytes, device_id: int, tid: int) -> bytes:
         """Encode MDAP+payload."""
--- a/pymodbus/framer/tls.py
+++ b/pymodbus/framer/tls.py
@@ -23,15 +23,8 @@ class FramerTLS(FramerBase):
         if (data_len := len(data)) < self.MIN_SIZE:
             return 0, 0, 0, self.EMPTY
         tid = int.from_bytes(data[0:2], 'big')
-        if int.from_bytes(data[2:4], 'big'):
-            return 0, 0, 0, self.EMPTY
-        msg_len = int.from_bytes(data[4:6], 'big') + 6
         dev_id = int(data[6])
-        if data_len < msg_len:
-            return 0, 0, 0, self.EMPTY
-        if msg_len == 8 and data_len == 9:
-            msg_len = 9
-        return msg_len, dev_id, tid, data[7:msg_len]
+        return data_len, dev_id, tid, data[7:]
 
     def encode(self, payload: bytes, device_id: int, tid: int) -> bytes:
         """Encode MDAP+payload."""
--- a/pymodbus/framer/tls.py
+++ b/pymodbus/framer/tls.py
@@ -21,7 +21,7 @@ class FramerTLS(FramerBase):
     def decode(self, data: bytes) -> tuple[int, int, int, bytes]:
         """Decode MDAP+payload."""
         if (data_len := len(data)) < self.MIN_SIZE:
-            return 0, 0, 0, self.EMPTY
+            return data_len, 0, 0, self.EMPTY
         tid = int.from_bytes(data[0:2], 'big')
         dev_id = int(data[6])
         return data_len, dev_id, tid, data[7:]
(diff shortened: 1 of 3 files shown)


pymodbus/framer/tls.py
```diff
@@ -20,9 +20,18 @@ class FramerTLS(FramerBase):
 
     def decode(self, data: bytes) -> tuple[int, int, int, bytes]:
         """Decode MDAP+payload."""
+        if (data_len := len(data)) < self.MIN_SIZE:
+            return 0, 0, 0, self.EMPTY
         tid = int.from_bytes(data[0:2], 'big')
+        if int.from_bytes(data[2:4], 'big'):
```
janiversen: This seems wrong, why should we demand the PID be 0 ?

A lot of implementation use the pid for special purposes.

pymodbus/framer/tls.py
```diff
@@ -20,9 +20,18 @@ class FramerTLS(FramerBase):
...
+            return 0, 0, 0, self.EMPTY
+        msg_len = int.from_bytes(data[4:6], 'big') + 6
         dev_id = int(data[6])
-        return len(data), dev_id, tid, data[7:]
+        if data_len < msg_len:
+            return 0, 0, 0, self.EMPTY
+        if msg_len == 8 and data_len == 9:
+            msg_len = 9
+        return msg_len, dev_id, tid, data[7:msg_len]
```
janiversen: This is wrong it should be data_len, the TLS library delivers packets not a byte stream, so there are not "leftovers", between calls.

pymodbus/framer/tls.py
```diff
@@ -20,9 +20,18 @@ class FramerTLS(FramerBase):
...
+            return 0, 0, 0, self.EMPTY
         tid = int.from_bytes(data[0:2], 'big')
+        if int.from_bytes(data[2:4], 'big'):
+            return 0, 0, 0, self.EMPTY
+        msg_len = int.from_bytes(data[4:6], 'big') + 6
         dev_id = int(data[6])
-        return len(data), dev_id, tid, data[7:]
+        if data_len < msg_len:
+            return 0, 0, 0, self.EMPTY
+        if msg_len == 8 and data_len == 9:
```
janiversen: this seems wrong, data_len == 9, means a UID without a modbus message, that is not defined in the protocol.

janiversen (changes requested):
There are a couple of comments below, please resolve those.

Short frames is basically not possible with the TLS library, because  the sender encrypt the whole packet, and the receiver do not decrypt until the whole packet is received.

However your check do not hurt, and can catch ill-implemented devices.

janiversen:
you need to add tests test/framer/test_framer.py to ensure the new code is tested.

your PR reduces the test coverage from 100% to 99,95% !! We are proud to have 100% test coverage and want to keep it this way.

pymodbus/framer/tls.py
MrAlaskan: Thanks for the comment.

pymodbus/framer/tls.py
MrAlaskan: I agree with you. That requirement doesn’t make sense, and I’ve revised it.

pymodbus/framer/tls.py
```diff
@@ -20,9 +20,11 @@ class FramerTLS(FramerBase):
 
     def decode(self, data: bytes) -> tuple[int, int, int, bytes]:
         """Decode MDAP+payload."""
+        if (data_len := len(data)) < self.MIN_SIZE:
+            return 0, 0, 0, self.EMPTY
```
janiversen: This will accumulate data between packets, which is not allowed in TLS (I think I wrote that in another comment).

Returning 0, causes the data to be appended to the next packet, which then will fail due to the mba being wrong.

janiversen (changes requested):

pymodbus/framer/tls.py
MrAlaskan: My mistake, thanks for the correction! Chatting with you gives me a much better grasp of how the Modbus TLS works.

janiversen (approved):
LGTM, Thanks.

MrAlaskan:
Based on your comments, my understanding is twofold:1) we do not need to strictly enforce the PID to be 0; and 2) returning `0` as the first value from from `decode()` upon encountering malformed data causes erroneous data accumulation.
However, upon reviewing the current implementation of the `decode()` method in another class, `FramerSocket (pymodbus/framer/socket.py)`, I noticed the following:
```c
if (pid := int.from_bytes(data[2:4], 'big')):
      Log.error("Invalid Modbus protocol id: {}", pid)
      return 0, 0, 0, self.EMPTY
```
This appears to contradict the principle you mentioned, as it returns `0, 0, 0, self.EMPTY` when an invalid PID is detected. Could you help clarify this discrepancy? I look forward to your insights!

janiversen:
The pid test in socket is a leftover that should have been removed long time ago.

With regards to returning 0 or data_len:

Socket is a bytestream protocol where TLS is a packet protocol  (in terms of what is sent from and received).

With socket pymodbus might need to read several times to get a complete package, hence "leftovers" are allowed (and often used). With TLS the asyncio transport delivers a frame with each read (I think this is due to the fact that decryption is not done as a stream).

MrAlaskan:
It makes sense that pymodbus retains "leftover" data, as socket connections often require multiple reads to complete a packet. However, in the case of the PID check, no matter how many times subsequent data is read, the invalid PID will always stay at the head of the buffer. This causes the code to re-enter this branch and return 0 over and over again, creating an infinite loop.

**Not a big deal. Deleting this useless PID check easily solves the problem.** If you don't mind, I can submit a follow-up PR later to address this.

janiversen:
pull requests are always welcome.

The pid will not stay in the start of data forever, after a while with different factors it will be removed.

Removing the tls encode() decode() would also make maintenance easier.

PR #2923, opened 2026-04-14
Test CI will fail.

janiversen:


janiversen:
In the current implementations the tls.decode() encode() could probably be removed, with slight modifications to the socket decode()/encode()

They were implemented differently for historical reasons.

PR #2924, opened 2026-04-20
Fix reg_count calc

rubenmendes-eaton:
The current formula computes the number of 16-bit registers needed based only on the bit count, ignoring where within the first register the read starts. It breaks when bit_offset + count > 16, i.e., the requested bits actually span a register boundary.

Also, when address is a multiple of 16, the formula over-allocates by 1 register.

The first part of the following "if" also does not seem correct.


janiversen (approved):
LGTM, thanks

PR #2926, opened 2026-04-23
Fix address bounds check in get_reg_block

foolip:
Because offset can be negative, register_count <= offset < 0 doesn't work.

It's specifically `rt.async_getValues(0x03, 8, 1)` that would previously
incorrectly pass the bounds check and return a value (15) due to negative list
indexing. Going even further out of bounds would instead throw an IndexError.

Discovered while SimDevice/SimData for a simple static test server.

### DIFF
--- a/pymodbus/simulator/simruntime.py
+++ b/pymodbus/simulator/simruntime.py
@@ -72,7 +72,7 @@ class SimRuntime:
         """Handle holding registers and input registers."""
         start_address, register_count, registers, _ = self.block[block_id]
         offset = address - start_address
-        if register_count <= offset < 0 or offset + count > register_count:
+        if offset < 0 or offset + count > register_count:
             return ExcCodes.ILLEGAL_ADDRESS
         if (result := await self.__check_block(func_code, block_id, address, count, offset, values)):
             return result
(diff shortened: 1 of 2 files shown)


pymodbus/simulator/simruntime.py
```diff
@@ -72,7 +72,7 @@ async def get_reg_block(self, block_id: str, func_code: int, address: int, count
         """Handle holding registers and input registers."""
         start_address, register_count, registers, _ = self.block[block_id]
         offset = address - start_address
-        if register_count <= offset < 0 or offset + count > register_count:
+        if offset < 0 or offset + count > register_count:
```
janiversen, line 75: This seems to be the wrong fix, it would be more logical to test that address >= start_address.

This should also be done in set_values

janiversen (changes requested):
Good catch, but needs some small changes.

janiversen (approved):

pymodbus/simulator/simruntime.py
foolip: Did you mean to merge this without any further changes?

janiversen: yes, the PR solved the problem which is enough.

Issue #2928, opened 2026-04-28
Swap pyserial for serialx


balloob:
### Versions

- Python: 3.14
- OS: Mac
- Pymodbus: 3.11.2
- Modbus Hardware (if used):

### Pymodbus Specific

- Server: rtu async
- Client: rtu async

### Description

The Open Home Foundation has released a new serial library `serialx` that is made with asyncio as a first class citizen. It is actively maintained (vs pyserial, [last release was 2020](https://pypi.org/project/pyserial/#history)).

Would pymodbus be interested in migrating? It will work faster for asyncio use cases (like Home Assistant).

Migration is straight forward as it's meant to be a drop-in replacement. More information about `serialx` and how to migrate can be found here: https://developers.home-assistant.io/blog/2026/04/27/pyserial-to-serialx/

Issues like this could be resolved: https://github.com/pymodbus-dev/pymodbus/blob/8a97aa0a12fb47a0fb9077a5e7843eec7495f3c7/pymodbus/client/serial.py#L240-L242

janiversen:
Just had a quick look, sound like a nice library.

I could not find anything about RS485, something we have been fighting with for a long time in pyserial.

pymodbus contains its own async-serial (due to the lack of maintenance of pyserial), so it is more than just plug and pray.

A couple of objectives to replace pyserial:
- serialx-compat is not a way to go !
- pymodbus/client/serial.py should be completely removed since it just a async interface
- Test cases must be added (seems serialx do not provide a solution like pyserial that allows use of sockets)

If you are up to the task, then feel free to submit pull requests. We can offer to make a new branch (serial) where your changes can be merged, without breaking current dev.

janiversen:
@balloob welcome to the pymodbus project......seems our ways continue to intertwine.

puddly:
Hello! I'm the author of serialx. If there are any API gaps, I would be more than happy to talk about missing features. 

- We support connecting to a TCP server via the `socket://` URI handler (in addition to `esphome://`, `rfc2217://`, and a few more). I wasn't able to find any places within pymodbus where you pass around raw sockets or file descriptors so I think this should be a simple object rename.
- `pymodbus/client/serial.py` and `pymodbus/transport/serialtransport.py` can effectively be removed, since we provide a fully cross-platform sync and async serial transport (including Windows).
- > serialx-compat is not a way to go !
   
[...]

PR #2929, opened 2026-04-29
Migrate from pyserial and in-tree async serial transport to serialx

puddly:
Implementation of #2928.

This PR migrates from pyserial to serialx. As part of the change, I've dropped the in-tree asyncio adapter in favor of serialx's native serial code. serialx does not use polling on any platform or for either API so I've removed a few tests that referenced it.

Two major bugfixes:
1. The in-tree async serial transport called `Serial.close()` [within the event loop](https://github.com/pymodbus-dev/pymodbus/blob/8a97aa0a12fb47a0fb9077a5e7843eec7495f3c7/pymodbus/transport/serialtransport.py#L59). This is a blocking operation for the `socket://` URI scheme and can block almost indefinitely with `os.close()` on Linux, especially at low baudrates. For code running in a shared event loop (like in Home Assistant), this will cause issues. https://github.com/home-assistant-libs/pyserial-asyncio-fast/pull/16 has more context.
2. Windows support no longer relies on any sort of polling. Serialx uses overlapped IO on Windows and hooks into the proactor event loop for async.

Unrelated, but I think a codepaths using [`await asyncio.sleep(0.1)`](https://github.com/pymodbus-dev/pymodbus/blob/8a97aa0a12fb47a0fb9077a5e7843eec7495f3c7/pymodbus/client/base.py#L62) may be due to a race between opening a transport and `connection_made` being called.


pymodbus/client/serial.py
```diff
@@ -16,7 +14,13 @@
...
 
 with contextlib.suppress(ImportError):
-    import serial
+    import serialx
+
+
+# Buffer size for a single `read()` syscall when the caller didn't specify how many
+# bytes to read. Larger than any Modbus frame, so one syscall returns whatever's
+# currently in the kernel buffer without truncation.
+DEFAULT_RECV_SIZE = 4096
```
janiversen: This is way too big !! this must be synchronized with the cutoff limit in transport.py. The cutoff was added to close huge frames as an attack vector.

pymodbus/client/serial.py
```diff
@@ -227,19 +226,18 @@ def connect(self) -> bool:
...
                 baudrate=self.comm_params.baudrate,
                 parity=self.comm_params.parity,
                 exclusive=True,
+                inter_byte_timeout=self.inter_byte_timeout,
             )
-            self.socket.inter_byte_timeout = self.inter_byte_timeout
-        # except serial.SerialException as msg:
-        # pyserial raises undocumented exceptions like termios
-        except Exception as msg:  # pylint: disable=broad-exception-caught
+            self.socket.open()
```
janiversen, line 239: Where is the inter_byte_timeout ? this is used to detect broken frames.

pymodbus/client/serial.py
```diff
@@ -250,53 +248,25 @@ def close(self):
...
         """Read data from the underlying descriptor."""
         if not self.socket:
             raise ConnectionException(str(self))
-        if size is None:
-            size = self._wait_for_data()
-        if size > self._in_waiting():
-            self._wait_for_data()
-        result = self.socket.read(size)
-        return result
+        return self.socket.read(size if size else DEFAULT_RECV_SIZE)
```
janiversen, line 269: This seems to ignore the whole idea of waiting for data, something which is important with low BPS.

This can return 0, which meaning the wait for data must be handled at the calling level, which is not acceptable....all serial special handling must be done at this level.

test/client/test_client_sync.py
```diff
@@ -293,11 +293,11 @@ def test_sync_serial_client_instantiation(self):
             FramerRTU,
         )
 
-    @mock.patch("serial.Serial")
+    @mock.patch("serialx.serial_for_url")
```
janiversen, line 296: it looks as if, this is not testing a serial line, but the socket part. If correct the tests should be duplicated to ensure it works for both.

Same goes for the next tests.

test/transport/test_comm.py
```diff
@@ -153,10 +152,10 @@ async def test_split_serial_packet(self, client, server, use_port):
         test_data = b"abcd"
 
         self.serial_write = (  # pylint: disable=attribute-defined-outside-init
-            client.transport.sync_serial.write
+            client.transport.serial.write
```
janiversen, line 155: client.transport provides to different serial implementations, a sync and an async, not sure if this change just combined them.

pyproject.toml
```diff
@@ -69,7 +69,6 @@ development = [
     "ruff>=0.15.0",
     "twine>=6.2.0",
     "types-Pygments",
-    "types-pyserial",
```
janiversen, line 72: This cannot be true, does serialx not have a typed API ???

janiversen (changes requested):
To be very fair and based on experience, that the tests pass is good and like sufficient for the async code.

But for the sync code, we have over time had so many issues stemming from real life devices, that we must be very careful to throw out that code. Your code removes all the wait for data as well as inter byte timeouts which scares me.

pymodbus/client/serial.py
puddly: It moved into the constructor, just a few lines up:
```diff
                parity=self.comm_params.parity,
                exclusive=True,
+               inter_byte_timeout=self.inter_byte_timeout,
            )
-           self.socket.inter_byte_timeout = self.inter_byte_timeout
```

pymodbus/client/serial.py
puddly: I don't think behavior has changed.

The old function read all of the buffered data, if there is any. If there was none, it busy polled for up to `self.comm_params.timeout_connect` seconds until some data arrived. If nothing arrived, it returned an empty string. This is exactly how `serial.read(max_size)` works if you pass `read_timeout=self.comm_params.timeout_connect,` to the serial class constructor.

pyproject.toml
puddly: Complete types are provided in the `serialx` package, there's no need to install a second one.

test/client/test_client_sync.py
puddly: `serial_for_url` is the dispatcher for all serial backends. Directly creating `Serial` objects isn't recommended because it is the platform-native serial constructor and will not work for other types of serial ports (e.g. RFC2217, ESPHome, TCP).

pymodbus/client/serial.py
puddly: I also think this may be worth noting: `inter_byte_timeout` in POSIX has a resolution of 0.1s so I think this feature always effectively sets `inter_byte_timeout = 0`, since `t0` would be less than 0.1s for any baudrate higher than 66 baud.

On Windows, it has a higher resolution and work as intended.

test/transport/test_comm.py
puddly: `sync_serial` what the `serialtransport.py` called the `Serial` object attached to the transport. In serialx, we expose it as just `serial`, for backwards compatibility. Ideally, the test should not touch the sync serial API on the async transport. We only expose it for backwards compatibility.

pymodbus/client/serial.py
puddly: The old implementation effectively called `.read(None)`, which returned _everything_ in the kernel buffer, so the old code technically had no limit at all. I've reduced it to 1024.

pymodbus/client/serial.py
puddly: If it's simpler or if I'm not understanding how `read()` is supposed to work, I can just restore the old behavior exactly, all of the functions are the same.

pymodbus/client/serial.py
janiversen: Well pymodbus is used on quite a number of platforms, so maybe it does not work on POSIX but helps elsewhere.

janiversen (approved):
This PR seems to work with Async apps, however edge cases have not been tested.

The sync part works with the standard serial interface at least in good weather situations, however due to the big number of RS485-USB converters it is not possible to ensure it works in all cases. 

The sync version of pymodbus is heavily used especially by people testing their new devices and exactly that fact means a change from pyserial to serialx is a very  critical move.

Changing from pyserial to serialx do have an effect on all apps when upgrading, and at least one production system would have problems (because they use a patched pyserial).

This PR is however correct, and should be merged.

janiversen:
The code is approved and can be merged once marked "ready for review".

However please change the target to pymodbus-dev:serialx, to allow it to coexist with the production version, giving users a change to verify it.

Since this is a major change, it will be part of an upcoming v4.0 version.

Thanks for the PR.

pyproject.toml
janiversen: Python documentation tells something different:

--> Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing. This marker applies recursively: if a top-level package includes it, all its sub-packages MUST support type checking as well.

https://typing.python.org/en/latest/spec/distributing.html

puddly: Yeah, we provide a `py.typed` file: https://github.com/puddly/serialx/blob/dev/serialx/py.typed

PR #2009, opened 2024-02-13
Move repl to a seperate repo

dhoomakethu:
* REPL code moved here https://github.com/pymodbus-dev/repl
* `pymodbus.console` and `pymodbus.server` are still available once installed as before
* Installation of optional `repl` feature is still the same `pip install ".[repl]"`

This way we can keep `pymodbus` core library separate from all the helpers and utils.

## TODO
* Tests and code checks on `repl` repo to be used
* `console` and `server` commands use different approach for `commands` and `hints`, to be moved to one common lib.

**NOTE**

This is just an idea, we can discuss the best coure of action and take action with this PR accordingly,

PR #1288, opened 2023-01-24
Fix #1258 - Repl client

dhoomakethu:
Fix #1258 Avoid showing unit as a seperate command line argument
Screenshot 2023-01-24 at 3 04 09 PM

PR #2932, opened 2026-05-21
docs(CHANGELOG): fix 'seperate' -> 'separate' (4 occurrences)

taiman724:
## Summary
Pure historical-changelog typo fix in `CHANGELOG.rst` for four release sections:

- 3.6.5: \"Move repl to a seperate repo (#2009)\"
- 3.1.2: \"Avoid showing unit as a seperate command line argument (#1288)\"
- 1.4.0: \"Asynchronous Server could now be stopped when running on a seperate thread\"
- 1.3.2: \"ModbusSerialServer could now be stopped when running on a seperate thread.\"

No code changes; PR rebased on `dev` per the PR template.

## Verification
- `grep -n seperate CHANGELOG.rst` → 0 matches after the change.
- All four lines are inside historical version sections, so the substitution is locally faithful to the original meaning.

## Note
Co-developed with AI assistance; reviewed and tested by submitter.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -557,7 +557,7 @@ Version 3.6.5
 * Allow slave=0 in serial communication. (#2023)
 * Client package test tool. (#2022)
 * Add REPL documentation back with links to REPL repo (#2017)
-* Move repl to a seperate repo (#2009)
+* Move repl to a separate repo (#2009)
 * solve more mypy issues with client (#2013)
 * solve more mypy issues with datastore (#2010)
 * Remove useless. (#2011)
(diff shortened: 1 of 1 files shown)


janiversen (changes requested):
We only add to change.log we do not alter old releases. Your change should really be changed in the commit text, calling for a 2 new releases.

But the spelling error is likely in the source or docs, since the PR surely were documented.

taiman724:
Thanks for the review @janiversen — that policy makes sense, and I should have checked first whether the typo lived anywhere outside the changelog before submitting.

I just grepped the dev branch for "seperate" across all non-CHANGELOG files (`*.py`, `*.md`, `*.rst`, `*.txt`, `*.yml`) and there are no current source/docs occurrences — only the four historical entries in `CHANGELOG.rst`. So under the append-only changelog policy, there's nothing left to do here.

Closing in respect of the policy. Apologies for the noise.

Note: Co-developed with AI assistance; reviewed and tested by submitter.

Issue #2939, opened 2026-06-23
`DATATYPE_STRUCT` maps `DataType.REGISTERS` to (signed) INT16 instead of UINT16


km-64:
I ran into some problems when using `pymodbus` 3.13.x while using the new SimData/SimDevice features. I narrowed it down to the `DataType.REGISTERS` type and a seemingly incorrect mapping to signed shorts rather than unsigned shorts.

In the [docs](https://pymodbus.readthedocs.io/en/latest/source/simulator/datamodel.html#pymodbus.simulator.DataType.REGISTERS) and the comment near the definition of `DataType.REGISTERS`, it states that `REGISTERS` is identical to `UINT16` (unsigned short).

https://github.com/pymodbus-dev/pymodbus/blob/a7a0a0d855a431a3800007b4a11328717f97926e/pymodbus/simulator/simutils.py#L48-L49

However, looking at the `DATATYPE_STRUCT` mapping between `DataType` and a tuple containing [format characters](https://docs.python.org/3/library/struct.html#format-characters), `DataType.REGISTERS` maps to format character `h` which corresponds to a signed short/int16 as seen below.

https://github.com/pymodbus-dev/pymodbus/blob/a7a0a0d855a431a3800007b4a11328717f97926e/pymodbus/simulator/simutils.py#L54-L67

Shouldn't the format character used for `REGISTERS` match that of `UINT16` rather than that of `INT16`? This could be fixed by replacing `h` with `H` for `REGISTERS`. 

The current behaviour doesn't look intentional. There aren't any tests checking that `REGISTERS` can be assigned values that are within the bounds of unsigned shorts.

**More details**
This is where the `DATATYPE_STRUCT` dict is used:
https://github.com/pymodbus-dev/pymodbus/blob/a7a0a0d855a431a3800007b4a11328717f97926e/pymodbus/simulator/simdata.py#L194-L210

Here is a simplified snippet that demonstrates what happens when I try to assign a value that should work for the `REGISTERS` data type.

```python
import struct

struct.pack(">H", 0xFFFF)
# works fine uint 16 max value

struct.pack(">h",  0xFFFF)
# struct.error: 'h' format requires -32768 <= number <= 32767
```

km-64:
Commit 99c88bd99eac0af62adb1af621dae5dbe5dbe062 introduced this new way using `struct.pack` starting in v3.12.0

janiversen:
Solved in #2940.

PR #2940, opened 2026-06-23
Use correct format character for `DataType.REGISTERS` 'h' -> 'H'

km-64:
Addresses  Issue #2939


janiversen (approved):
Good catch, thanks.

PR #2942, opened 2026-06-24
Support configured ASCII input delimiter

MrAlaskan:
# Summary

This PR fixes `FramerAscii` so that `Change ASCII Input Delimiter` actually affects future ASCII messages.

Previously, the delimiter value was updated in the control block, but the ASCII framer still used a hardcoded `CRLF` terminator for both encoding and decoding.

According to the Modbus ASCII specification, the `Change ASCII Input Delimiter` diagnostic command replaces the default `LF` character used at the end of future messages.

# Change

- `FramerAscii` now builds its end-of-frame marker from the control block delimiter instead of always using a hardcoded `CRLF`
- `decode()` now looks for the configured delimiter
- `encode()` now emits the configured delimiter
- The control block default delimiter is set to `LF`, so the default end-of-frame remains `CRLF`

# Rationale

This aligns the framer behavior with both the protocol and the existing diagnostic command implementation.

Without this change, `Change ASCII Input Delimiter` updates internal state but does not affect actual frame parsing or frame generation, so the command appears to succeed while the wire behavior remains unchanged.

### DIFF
--- a/pymodbus/framer/ascii.py
+++ b/pymodbus/framer/ascii.py
@@ -12,6 +12,7 @@ from ..logging import Log
 from ..pdu.device import ModbusControlBlock
 from .base import FramerBase
 
+
 _MCB = ModbusControlBlock()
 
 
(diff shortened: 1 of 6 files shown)


janiversen (approved):
LGTM, thanks.

PR #2944, opened 2026-06-24
Fix `ModbusServerContext` initialization with dict `devices` parameter

### LINKED ISSUE #2943: `ModbusServerContext` incorrectly initializes `self.simdevices` when given a dict for `devices`
There is a problem in the new way `ModbusServerContext` initializes its `simdevice` instance attribute caused by changes available in 3.13.0. Referring to the docs, it specifically mentions that beginning with v3.13.0, the internal functionality of the old datastores will be 'adapted' to use SimData/SimDevice. So effectively the old datastores are just wrappers in v3.13 and above.

> The old datastores:
> - ModbusSequentialDataBlock,
> - ModbusSparseDataBlock,
> - ModbusSimulatorContext
> 
> are due to be removed in v4.0.0, starting with v3.13.0 the internal functionality are replaced by SimData/SimDevice.

-- https://pymodbus.readthedocs.io/en/latest/source/upgrade_40.html#convert-to-simdata-simdevice

I narrowed down the change that causes the problem to a2e69f600c44b22d2cb894ada168874ccdda8e23. 

As can be seen in `ModbusServerContext.__init__`, the constructor handles either a single device, or a dict of devices (there is a problem when using dict).

https://github.com/pymodbus-dev/pymodbus/blob/347979fc8d5661536da52b7d09a1b13c7ca785f0/pymodbus/datastore/context.py#L81-L104

Notice that the `simdevices` instance attribute is given a type annotation `self.simdevices: list[SimDevice] = []`. At runtime this attribute is assigned an incorrect value when a dict is passed as the `devices` parameter.
```python
from pymodbus.datastore import ModbusServerContext, ModbusDeviceContext
from pymodbus.simulator import SimDevice


# Using a single device (works fine)
device_context = ModbusDeviceContext()
server_context = ModbusServerContext(devices=device_context)
assert all(map(lambda simdevice: isinstance(simdevice, SimDevice))), "simdevices aren't all instances of SimDevice"
# PASSES
print(server_context.simdevices)
# [SimDevice(id=0, simdata=([SimData(address=0, count=1, values=[False], datatype=<DataType.BITS: 11>, string_encoding='utf-8', readonly=False)], [SimData(address=0, count=1, values=[False], datatype=<DataType.BITS: 11>, string_encoding='utf-8', readonly=False)], [SimData(address=0, count=1, values=0, datatype=<DataType.REGISTERS: 12>, string_encoding='utf-8', readonly=False)], [SimData(address=0, count=1, values=0, datatype=<DataType.REGISTERS: 12>, string_encoding='utf-8', readonly=False)]), use_bit_addressing=True, identity=None, action=None)]


# Using a dict (initialization problem)
device_context = ModbusDeviceContext()
server_context = ModbusServerContext(devices={0: device_context})
assert all(map(lambda simdevice: isinstance(simdevice, SimDevice))), "simdevices aren't all instances of SimDevice"
# FAILS
print(server_context.simdevices)
# [<pymodbus.datastore.context.ModbusDeviceContext object at 0x7f2d9fa7a7b0>]
# this is an instance of something other than SimDevice
```

Currently the `zuban` type checker can't find any errors because it doesn't know the type of `entry` (see like 97). Since it looks like the only time these lines would run is when `entry` is an instance of `ModbusDeviceContext`, it could be assumed that entry will always be of type `ModbusDeviceContext`, however the type checker is not able to determine this information and defaults to `Any`.
https://github.com/pymodbus-dev/pymodbus/blob/347979fc8d5661536da52b7d09a1b13c7ca785f0/pymodbus/datastore/context.py#L98-L100

Adding an assertion statement as follows gives the type checker enough information to narrow down the type and report errors:
```python
for dev_id, entry in devices.items():
    if not isinstance(entry, ModbusSimulatorContext):
        typing.reveal_type(entry)  # note: Revealed type is "Any"
        assert isinstance(entry, ModbusDeviceContext)  # narrowing assertion
        typing.reveal_type(entry)  # note: Revealed type is "pymodbus.datastore.context.ModbusDeviceContext"
        
        entry.id = dev_id  # error: "ModbusDeviceContext" has no attribute "id"  [attr-defined]
        self.simdevices.append(entry)  #  error: Argument 1 to "append" of "list" has incompatible type "ModbusDeviceContext"; expected "SimDevice"  [arg-type]
```

It looks like `entry.simdevice` should be used instead. This is actually what the non-dict version does:
https://github.com/pymodbus-dev/pymodbus/blob/347979fc8d5661536da52b7d09a1b13c7ca785f0/pymodbus/datastore/context.py#L101-L104

**Solution**
If it is safe to assume that I haven't broken any intentional functionality, the following solution fixes the type errors from the previous example.
```python
if not isinstance(entry, ModbusSimulatorContext):
    entry.simdevice.id = dev_id
    self.simdevices.append(entry.simdevice)
```
Basically I just used the `simdevice` property of `entry` instead of `entry` itself. By looking at the examples and tests, I assume that this branch will only run when entry is an instance of `ModbusDeviceContext`. Is this assumption valid?

**Possible type-checking improvements**
`ModbusServerContext.__init__` could be given some more type information, particularly for the `devices` parameter. This would give `zuban` the information it needs to determine the type of `entry`.

km-64:
Addresses the solution described in issue #2943

fixes #2943

### DIFF
--- a/pymodbus/datastore/context.py
+++ b/pymodbus/datastore/context.py
@@ -96,8 +96,9 @@ class ModbusServerContext:
             self._devices = devices
             for dev_id, entry in devices.items():
                 if not isinstance(entry, ModbusSimulatorContext):
-                    entry.id = dev_id
-                    self.simdevices.append(entry)
+                    simdevice = entry.simdevice
+                    simdevice.id = dev_id
+                    self.simdevices.append(simdevice)
         else:
             self._devices = {0: devices}
             if not isinstance(devices, ModbusSimulatorContext):
--- a/pymodbus/datastore/context.py
+++ b/pymodbus/datastore/context.py
@@ -78,7 +78,14 @@ class ModbusServerContext:
     device contexts.
     """
 
-    def __init__(self, devices=None, single=True):
+    def __init__(
+        self,
+        devices: dict[int, ModbusDeviceContext | ModbusSimulatorContext]
+        | ModbusDeviceContext
+        | ModbusSimulatorContext
+        | None = None,
+        single=True,
+    ) -> None:
         """Initialize a new instance of a modbus server context.
 
         :param devices: A dictionary of client contexts
@@ -90,7 +97,7 @@ class ModbusServerContext:
         _ = single
         if not devices:
             raise TypeError("devices= cannot be None")
-        self._devices: dict[int, ModbusDeviceContext]
+        self._devices: dict[int, ModbusDeviceContext | ModbusSimulatorContext]
         self.simdevices: list[SimDevice] = []
         if isinstance(devices, dict):
             self._devices = devices
@@ -108,7 +115,7 @@ class ModbusServerContext:
                     "Please convert to SimData/SimDevice.\n"
                     "Please read https://pymodbus.readthedocs.io/en/dev/source/upgrade_40.html#convert-to-simdata-simdevice")
 
-    def __get_device(self, device_id: int) -> ModbusDeviceContext:
+    def __get_device(self, device_id: int) -> ModbusDeviceContext | ModbusSimulatorContext:
         """Return device object."""
         if device_id in self._devices:
             return self._devices[device_id]


km-64:
Added type information mentioned in issue #2943 in [`263c9d9` (this PR)](https://github.com/pymodbus-dev/pymodbus/pull/2944/changes/263c9d91b4508be903a483df39ab86a2b540ab58)

km-64:
Also more generally, I noticed that this file isn't formatted with `ruff`. Quite a few files show up as unformatted when I run `ruff format --check`: `112 files would be reformatted, 12 files already formatted`.

Maybe at some point `uv run ruff format --check` could be added as a step in `ci.yml`? It would fail when there are changes ruff would make. Someone would have to make a big formatting commit.

janiversen:
> Also more generally, I noticed that this file isn't formatted with `ruff`. Quite a few files show up as unformatted when I run `ruff format --check`: `112 files would be reformatted, 12 files already formatted`.
> 
> Maybe at some point `uv run ruff format --check` could be added as a step in `ci.yml`? It would fail when there are changes ruff would make. Someone would have to make a big formatting commit.

Thanks for catching this, adding a check step seems like a good idea, CI run ruff on every PR, so it is a bit hard to explain how this could have happened.

janiversen (approved):
LGTM, thanks.

km-64:
> Thanks for catching this, adding a check step seems like a good idea, CI run ruff on every PR, so it is a bit hard to explain how this could have happened.

Right now the only ruff related step in ci is `ruff check .` which just checks the codebase against lint rules. It doesn't check formatting though.
https://github.com/pymodbus-dev/pymodbus/blob/2bad9357d0882191799b6654a25d017b150c0a9f/.github/workflows/ci.yml#L82-L85

To check the formatting you would have to use the ruff formatter with the `--check` option like this `ruff format --check`. It will fail if there are files that would be formatted.

> Meanwhile, ruff format --check exits with the following status codes:
> 
> 0 if Ruff terminates successfully, and no files would be formatted if --check were not specified.
> 1 if Ruff terminates successfully, and one or more files would be formatted if --check were not specified.
> 2 if Ruff terminates abnormally due to invalid configuration, invalid CLI options, or an internal error.
-- https://docs.astral.sh/ruff/formatter/#exit-codes


Adding `ruff format --check` would prevent PRs from being merged when they conflict with the formatter. I think that might explain why unformatted code entered the repo.

> CI run ruff on every PR, so it is a bit hard to explain how this could have happened.

PR #2948, opened 2026-06-25
Format test_client.py with ruff

km-64:
Fixed problems in test_client.py that cause zuban to report errors after formatting with ruff.

The cause was that ruff moved the ignore comment, so it had to be manually moved so that zuban would ignore the right argument.


janiversen (approved):
LGTM, thanks

PR #2958, opened 2026-07-15
Apply docstring fix in sync serial client from #1850

km-64:
The `stopbits` parameter typo introduced in #1820 was fixed in #1850, but only the `AsyncModbusSerialClient` docstring was updated. 
https://github.com/pymodbus-dev/pymodbus/blob/e3bc332b182a8df916477412ba2c8529f0b52a9c/pymodbus/client/serial.py#L37

This PR applies that same typo fix to the (sync) `ModbusSerialClient`.

### DIFF
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -130,7 +130,7 @@ class ModbusSerialClient(ModbusBaseSyncClient):
     :param baudrate: Bits per second.
     :param bytesize: Number of bits per byte 7-8.
     :param parity: 'E'ven, 'O'dd or 'N'one
-    :param stopbits: Number of stop bits 0-2.
+    :param stopbits: Number of stop bits 1, 1.5, 2.
     :param handle_local_echo: Discard local echo from dongle.
     :param name: Set communication name, used in logging
     :param reconnect_delay: Not used in the sync client
--- a/pymodbus/client/serial.py
+++ b/pymodbus/client/serial.py
@@ -33,7 +33,7 @@ class AsyncModbusSerialClient(ModbusBaseClient):
     :param baudrate: Bits per second.
     :param bytesize: Number of bits per byte 7-8.
     :param parity: 'E'ven, 'O'dd or 'N'one
-    :param stopbits: Number of stop bits 1, 1.5, 2.
+    :param stopbits: Number of stop bits 1, 1.5, 2. (Note: 1.5 falls back to 2 on POSIX systems)
     :param handle_local_echo: Discard local echo from dongle.
     :param name: Set communication name, used in logging
     :param reconnect_delay: Minimum delay when reconnecting, in seconds (use decimals for milliseconds).
@@ -130,7 +130,7 @@ class ModbusSerialClient(ModbusBaseSyncClient):
     :param baudrate: Bits per second.
     :param bytesize: Number of bits per byte 7-8.
     :param parity: 'E'ven, 'O'dd or 'N'one
-    :param stopbits: Number of stop bits 1, 1.5, 2.
+    :param stopbits: Number of stop bits 1, 1.5, 2. (Note: 1.5 falls back to 2 on POSIX systems)
     :param handle_local_echo: Discard local echo from dongle.
     :param name: Set communication name, used in logging
     :param reconnect_delay: Not used in the sync client
--- a/pymodbus/server/server.py
+++ b/pymodbus/server/server.py
@@ -229,7 +229,7 @@ class ModbusSerialServer(ModbusBaseServer):
         :param framer: The framer strategy to use, default FramerType.RTU
         :param identity: An optional identify structure
         :param port: The serial port to attach to
-        :param stopbits: The number of stop bits to use
+        :param stopbits: The number of stop bits to use 1, 1.5, 2. (Note: 1.5 falls back to 2 on POSIX systems)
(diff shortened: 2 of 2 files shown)


janiversen (approved):
LGTM Thanks.

km-64:
On the topic of stopbits, I see that the type for `stopbits` in `__init__` and `CommParams` is `int`, even though 1.5 is one of the possible values. Would it be more suitable to make it's type `int | float` to accomodate 1.5?

It could also just be made `float` due to type heuristics:
> int is duck type compatible with float and complex.
-- https://mypy.readthedocs.io/en/stable/duck_type_compatibility.html

>  [PEP 3141](https://peps.python.org/pep-3141/) defines Python’s numeric tower, and the stdlib module numbers implements the corresponding ABCs (Number, Complex, Real, Rational and Integral). There are some issues with these ABCs, but the built-in concrete numeric classes complex, float and int are ubiquitous (especially the latter two :-).
> 
> Rather than requiring that users write import numbers and then use numbers.Float etc., this PEP proposes a straightforward shortcut that is almost as effective: **when an argument is annotated as having type float, an argument of type int is acceptable**; similar, for an argument annotated as having type complex, arguments of type float or int are acceptable. This does not handle classes implementing the corresponding ABCs or the fractions.Fraction class, but we believe those use cases are exceedingly rare.
-- https://peps.python.org/pep-0484/#the-numeric-tower

I looked into `types-pyserial` and found that it accepts type `float` in the method that pymodbus uses.
https://github.com/python/typeshed/blob/9f9e8a76654cde06fa3b15c805d5864ab140eba9/stubs/pyserial/serial/__init__.pyi#L15-L30

It compares them against these constants (documented [here](https://pyserial.readthedocs.io/en/latest/pyserial_api.html#constants)):
https://github.com/pyserial/pyserial/blob/a5c48d445fbc1943d4fabf8d9090a50fda3172fd/serial/serialutil.py#L80
and will raise a type error if it isn't one of the supported values:
https://github.com/pyserial/pyserial/blob/a5c48d445fbc1943d4fabf8d9090a50fda3172fd/serial/serialutil.py#L340-L347

`pyserial` also states in its documentation that using stopbits 1.5 on POSIX systems isn't supported, and that it will fall back to using 2 stopbits.
> Note that 1.5 stop bits are not supported on POSIX. It will fall back to 2 stop bits.
 -- https://pyserial.readthedocs.io/en/latest/pyserial_api.html#constants
 
It's probably worth relaying this information to `pymodbus` users as they will likely not be familiar with the internal workings of `pyserial`, especially since the fallback happens silently.
https://github.com/pyserial/pyserial/blob/a5c48d445fbc1943d4fabf8d9090a50fda3172fd/serial/serialposix.py#L460-L467

km-64:
It looks like some tests are failing randomly. I don't think it's caused by my changes since they're only docstring related.

janiversen:
Making it float does not seem correct, it would be more logical to mame it a set of literales.

janiversen:
CI sometimes does strange things, restarted the job.

km-64:
> Making it float does not seem correct, it would be more logical to mame it a set of literales.

I had that thought too, but there's a problem with using literals with floats:
```python
stopbits: Literal[1, 1.5, 2]
                     ^^^
# Zuban/mypy
# error: Parameter 2 of Literal[...] cannot be of type "float"  [valid-type]

# Pyright
# error: Type arguments for "Literal" must be None, a literal value (int, bool, str, or bytes), or an enum value (reportInvalidTypeForm)
```

> 
> The following are provisionally disallowed for simplicity. We can consider allowing them in the future.
> 
> **Floats: e.g. Literal[3.14]. Representing Literals of infinity or NaN in a clean way is tricky; real-world APIs are unlikely to vary their behavior based on a float parameter.**
> 
> Any: e.g. Literal[Any]. Any is a type, and Literal[...] is meant to contain values only. It is also unclear what Literal[Any] would actually semantically mean.
-- https://typing.python.org/en/latest/spec/literal.html#illegal-parameters-for-literal-at-type-check-time

janiversen:
Well seems life as usual is not that simple.

Since int does the job, and furthermore 1.5 is not supported on Posix (see comment in pyserial) and thus pr definition not supported by  pymodbus

janiversen:
param stopbit doc is updated to clarify what is supported and what might work, meaning that declaring stopbit an int is correct.

km-64:
Having `stopbits` typed as `int` doesn't stop you from passing 1.5 anyway, it'll just give a type warning which can just be ignored. At runtime 1.5 gets passed all the way down to `pyserial` which can handle it properly on non-POSIX systems.

I guess literals would work for `parity`. Currently `pymodbus` only documents 'N', 'E', and 'O', but there is also 'M' and 'S' supported by `pyserial`. So instead of using `str` something like this could work:
```python
parity: Literal["N", "O", "E", "M", "S"]
```

Something similar could be done for `bytesize` because it's a specific range of ints defined in the [pyserial docs](https://pyserial.readthedocs.io/en/latest/pyserial_api.html#constants)
```python
bytesize: Literal[5, 6, 7, 8]
```

janiversen:
I never said it stops anybody (did you read the doc string comment, which actually says you can pass 1.5).

Yes there are many changes that can be done.....but it is not really worth the time just discussing them, if you want a change feel free to submit a PR.

PR #2959, opened 2026-07-15
Fix pack_bitstring input mutation

huhujiajia1:
## Summary

Prevent `pack_bitstring()` from modifying the list supplied by the caller.

The function previously assigned the input list directly to `t_bits` and
then extended it with padding bits. This caused non-aligned input lists to
be modified during encoding.

## Changes

- Copy the input list before appending padding bits.
- Add regression tests for byte and word alignment.
- Update the bit response encoding test to explicitly account for padding
  instead of relying on mutation of the original input list.

## Testing

- `python -m pytest test/pdu/test_pdu.py -n 0`
- `python -m pytest test/pdu -n 0`
- `./check_ci.sh`

Full test result:

- 1907 passed
- 5 skipped
- 99.97% coverage

### DIFF
--- a/pymodbus/pdu/utils.py
+++ b/pymodbus/pdu/utils.py
@@ -19,7 +19,7 @@ def pack_bitstring(bits: list[bool], align_byte=True) -> bytes:
     """
     ret = b""
     i = packed = 0
-    t_bits = bits
+    t_bits = bits.copy()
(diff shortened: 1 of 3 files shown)


janiversen (approved):
LGTM, thanks

PR #2962, opened 2026-07-16
Cleanup `ModbusSerialServer.__init__` kwargs

km-64:
`ModbusSerialServer`'s `__init__` method describes all kwargs, but some of them are stored in `**kwargs`. This can be made clearer with type information by giving them explicit types and default values.


pymodbus/server/server.py
```diff
@@ -219,6 +219,14 @@ def __init__(
         trace_pdu: Callable[[bool, ModbusPDU], ModbusPDU] | None = None,
         trace_connect: Callable[[bool], None] | None = None,
         custom_pdu: list[type[ModbusPDU]] | None = None,
+        baudrate: int = 19200,
+        reconnect_delay: float = 2,
```
km-64: In `CommParams` this can also be `None`, should `None` also be allowed here?
https://github.com/pymodbus-dev/pymodbus/blob/e3c1b6518076f2ec23cf7a059458350a6914551f/pymodbus/transport/transport.py#L86

pymodbus/server/server.py
```diff
@@ -246,19 +254,18 @@ def __init__(
...
-        baudrate = kwargs.get("baudrate", 19200)
         params = CommParams(
             comm_type=CommType.SERIAL,
             comm_name="server_listener",
-            reconnect_delay=kwargs.get("reconnect_delay", 2),
+            reconnect_delay=reconnect_delay,
             reconnect_delay_max=0.0,
-            timeout_connect=kwargs.get("timeout", 3),
+            timeout_connect=timeout,
             source_address=(kwargs.get("port", 0), 0),
```
km-64: I wasn't sure how to type the `port` parameter. `CommParams.source_address` expects `tuple[str, int]`, but `kwargs.get("port", 0)` would fall back to int 0. Is this intentional?

https://github.com/pymodbus-dev/pymodbus/blob/e3c1b6518076f2ec23cf7a059458350a6914551f/pymodbus/transport/transport.py#L91

The most faithful way to reproduce the current behaviour would be to do this:
```python
port: str = 0,  # type: ignore[assignment]
```
It could also be left without a default value
```python
port: str,
```

pymodbus/server/server.py
janiversen: No , CommParams are common to all transports.

pymodbus/server/server.py
```diff
@@ -219,7 +219,15 @@ def __init__(
...
-        **kwargs,
+        baudrate: int = 19200,
+        reconnect_delay: float = 2,
+        timeout: float = 3,
+        bytesize: int = 8,
+        parity: str = "N",
+        stopbits: int = 1,
+        handle_local_echo: bool = False,
+        allow_multiple_devices: bool = False,
+        port: str = 0,  # type: ignore[assignment]
```
janiversen: We try to avoid ignoring type checking !

It is probably hard to find a device that is present on all supported platforms, which could be given as default....so maybe an empty string.

janiversen (changes requested):
Please also update your dev, you are a couple of commits behind.

pymodbus/server/server.py
janiversen: kwargs seems unused.

km-64: > We try to avoid ignoring type checking !
> 
> It is probably hard to find a device that is present on all supported platforms, which could be given as default....so maybe an empty string.

There doesn't have to be a default value, it can also be left as just `port: str,`. If you try calling without passing `port` it will fail (`missing 1 required keyword-only argument: 'port'`).

That is unless having it fallback to an empty string is useful, but that would probably just raise an exception when `pyserial` can't open the port

pymodbus/server/server.py
```diff
@@ -207,19 +207,27 @@ class ModbusSerialServer(ModbusBaseServer):
         Remember to call serve_forever to start server.
     """
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
```
janiversen: Now you landed in before! pylint do not like many arguments.

The correct way to solve this is not to ask pylint to ignore....but instead tell pylint to accept it (py project.toml)

Remark changes to py project.toml should be done in a new PR, that should be merged before this one.

janiversen (changes requested):

janiversen:
Looking at coverage (ubuntu 3.14 in CI), it seems you have introduced a line that is not being tested:

"Required test coverage of 99.95% reached. Total coverage: 99.97%"

Please correct that.


False alarm !!!! this is NOT caused by this PR.

janiversen:
Please do not misunderstand my review comments !!

I like what you are doing, and all in favor of getting it done....including the other parameter changes you suggested.

BUT when doing it, it should be done correctly, so it is inline with the rest of the code.

btw. I think we have used kwargs as escape goat in several APIs.

janiversen (approved):
LGTM, Thanks.

PR #2967, opened 2026-07-16
Allow ModbusBaseServer to directly accept a SimCore instance as context

km-64:
Related to discussion #2966


janiversen:
Please SimCore is an internal class, that we do not want as part of the API !

The reason being, when it is internal we can change it as needed without breaking the API.

janiversen (changes requested):
This is a change that we do not want.

SimCore is internal and not part of the API. Please use the action function in SimDevice if you want to the content.

janiversen:
Closing with no response to review comments.

PR #2971, opened 2026-07-19
Fix synchronous response timeout on continuous garbage

### LINKED ISSUE #2960: `sync_get_response` hangs forever when garbage data arrives on serial line
## Versions

- pymodbus version: 3.12.1 (same code also present in 3.14.0)
- Python version: 3.10.12
- OS: Windows 10
- Communication: RS-485 via USB-to-485 adapter, RTU mode

## Pymodbus Specific

Client: rtu sync
Timeout: 8 seconds

## Describe the bug

When using `ModbusSerialClient`, if the serial port receives **any non-Modbus garbage data** (e.g., ASCII text from a misconfigured device, line noise), `sync_get_response()` sometimes enters an infinite loop and **never returns** — no TimeoutError, no exception. The calling thread blocks permanently.

## To Reproduce

1. Connect a Modbus RTU device A and another device B that continuously outputs non-Modbus text (e.g., ASCII `" Busy "`) on the same RS-485 bus.
2. Use `ModbusSerialClient` to read holding registers (code 0x03) in a polling loop.
3. After several times of successful reads, the polling loop stops waiting response.
4. Remove the device B from bus, and the polling loop recovers.

## Root cause

In `transaction.py`, `sync_get_response()` (line 82–114, present in **v3.12.1 through v3.14.0**):

```python
def sync_get_response(self, dev_id, tid) -> ModbusPDU:
    databuffer = b""
    while True:                                          # ← no time limit
        if not (data := self.sync_client.recv(None)):
            raise asyncio.exceptions.TimeoutError()      # ← only triggers when recv() returns b""

        databuffer += data
        used_len, pdu = self.framer.handleFrame(...)
        databuffer = databuffer[used_len:]
        if pdu:
            return ...
```

The loop **only exits when `recv(None)` returns empty bytes** (triggering TimeoutError) **or** when a valid Modbus PDU is parsed. But if garbage data keeps arriving (even slowly), `recv()` never returns empty, and the framer never produces a valid PDU. The `while True` runs forever.

Note: the retry logic in `sync_execute()` is **also bypassed** because no exception is ever raised to trigger it.

## Expected behavior

When `sync_get_response()` cannot parse a valid Modbus PDU within the configured `timeout_connect` period, it should raise `TimeoutError`, allowing `sync_execute()` to retry and eventually raise `ModbusIOException`.

## Fix

Add a hard deadline to `sync_get_response()`:

```diff
+import time

 def sync_get_response(self, dev_id, tid) -> ModbusPDU:
     databuffer = b""
+    deadline = time.time() + self.comm_params.timeout_connect
     while True:
+        if time.time() > deadline:
+            raise asyncio.exceptions.TimeoutError()
         if not (data := self.sync_client.recv(None)):
             raise asyncio.exceptions.TimeoutError()
         ...
```

This ensures the method always returns (or raises) within `timeout_connect` seconds, regardless of whether garbage data is arriving.

sankalpsthakur:
## Summary

- bound synchronous response parsing with a monotonic deadline based on the configured timeout
- check the deadline after each unsuccessful parse so a frame already being received can complete
- add a deterministic regression test for continuous invalid input

`sync_get_response()` previously depended on `recv()` returning empty bytes. Continuous non-Modbus input therefore kept the parser loop alive indefinitely and bypassed the retry logic in `sync_execute()`.

The socket-backed serial examples now use a ten-second response timeout. Their previous 100 ms setting was suitable as an inactivity timeout, but too short once the timeout also bounds the complete response parse, especially on loaded macOS CI runners.

## Validation

- `uv run pytest test/transaction/test_transaction.py` — 43 passed
- `uv run pytest` — 1,905 passed, 5 skipped
- `uv run codespell`
- `uv run ruff format --check .`
- `uv run ruff check .`
- `uv run pylint --recursive=y examples pymodbus test`
- `uv run zuban check pymodbus examples test`

Fixes #2960

### DIFF
--- a/pymodbus/transaction/transaction.py
+++ b/pymodbus/transaction/transaction.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 from collections.abc import Callable
 from threading import RLock
+from time import monotonic
 
 from ..exceptions import ConnectionException, ModbusIOException
 from ..framer import FramerAscii, FramerBase, FramerRTU
@@ -83,6 +84,7 @@ class TransactionManager(ModbusProtocol):
     def sync_get_response(self, dev_id, tid) -> ModbusPDU:
         """Receive until PDU is correct or timeout."""
         databuffer = b""
+        deadline = monotonic() + self.comm_params.timeout_connect
         while True:
             if not (data := self.sync_client.recv(None)):
                 raise asyncio.exceptions.TimeoutError()
@@ -115,6 +117,8 @@ class TransactionManager(ModbusProtocol):
             databuffer = databuffer[used_len:]
             if pdu:
                 return self.trace_pdu(False, pdu)
+            if monotonic() >= deadline:
+                raise asyncio.exceptions.TimeoutError()
 
     def sync_execute(self, no_response_expected: bool, request: ModbusPDU) -> ModbusPDU:
         """Execute requests asynchronously.
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -119,7 +119,9 @@ def define_commandline_client(
 ):
     """Define commandline."""
     my_port = str(use_port)
-    cmdline = ["--comm", use_comm, "--framer", use_framer, "--timeout", "0.1"]
+    # Socket-backed serial can take seconds to complete long RTU frames under load.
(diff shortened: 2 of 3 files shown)


janiversen (approved):
LGTM, thanks

PR #2973, opened 2026-07-22
Fix sync disconnect counter reset after successful response

### LINKED ISSUE #2972: sync_execute does not reset count_until_disconnect after a successful response
### Versions

- Python: 3.12
- OS: macOS
- Pymodbus: `dev` (`2c9c96a`)
- Modbus Hardware (if used): n/a (library behavior)

### Pymodbus Specific

- Server: n/a
- Client: tcp/rtu - sync

### Description

`TransactionManager.execute()` (async) resets `count_until_disconnect` after a successful response:

```python
self.count_until_disconnect = self.max_until_disconnect
```

`TransactionManager.sync_execute()` does not. The methods are documented as mirrors:

> REMARK: this method is identical to execute, apart from the lock and sync_receive.
> any changes in either method MUST be mirrored !!!

On a sync client, intermittent timeouts that exhaust retries decrement `count_until_disconnect`. Later successful responses never restore it, so the client still moves toward the forced disconnect path (`retries+3` exhausted no-response cycles) even after the device recovers.

Async clients recover correctly because a successful response resets the counter.

Expected: sync and async paths both reset `count_until_disconnect` on a successful response.

### Code and Logs

```python
# Minimal illustration of the missing mirror (not a hardware repro).
# After a sync response success path, count_until_disconnect should equal
# max_until_disconnect, matching async execute().

import pymodbus
```

I can prepare a small patch with a regression test if welcome.

sankalpsthakur:
## Summary

- reset `count_until_disconnect` after a successful `sync_execute()` response
- add a regression test that depletes the counter, then confirms a successful sync response restores it

`TransactionManager.execute()` already resets the counter on success. `sync_execute()` did not, despite the documented requirement that the two paths stay mirrored. Intermittent sync timeouts that exhaust the disconnect budget therefore left a recovered device one successful transaction away from a forced disconnect.

Fixes #2972

## Test plan

- [x] New regression in `test/transaction/test_transaction.py`
- [x] Upstream CI matrix

### DIFF
--- a/pymodbus/transaction/transaction.py
+++ b/pymodbus/transaction/transaction.py
@@ -141,6 +141,7 @@ class TransactionManager(ModbusProtocol):
                     response = self.sync_get_response(
                         request.dev_id, request.transaction_id
                     )
+                    self.count_until_disconnect = self.max_until_disconnect
                     if response.dev_id != request.dev_id:
                         raise ModbusIOException(
                             f"ERROR: request uses device id={request.dev_id} but received {response.dev_id}."
--- a/test/transaction/test_transaction.py
+++ b/test/transaction/test_transaction.py
@@ -523,6 +523,35 @@ class TestSyncTransaction:
         resp = transact.sync_execute(False, request)
         assert response.bits == resp.bits
 
+    def test_sync_execute_resets_count_until_disconnect(self, use_clc):
+        """Successful sync response must reset the disconnect counter like async."""
+        client = self.dummy_client(use_clc)
+        transact = TransactionManager(
+            use_clc,
+            FramerRTU(DecodePDU(False)),
+            5,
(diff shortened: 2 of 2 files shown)


sankalpsthakur:
Windows 3.12 failed on an unrelated TLS flake (`test_connected[CommType.TLS-localhost]`), not this change. Empty commit pushed to retrigger.

sankalpsthakur:
Prior matrix failures were unrelated: uv install timeout on 3.14, and the same Windows TLS `test_connected` flake on 3.10. Retriggered with `a25c339`. Diff remains the sync counter reset plus regression test.

janiversen (approved):
LGTM, Thanks.

And special thanks for adding a test !

PR #2974, opened 2026-07-23
Fix write multiple coils byte count validation

MrAlaskan:
## Summary

Validate the byte count when processing a Write Multiple Coils (FC 0x0F) request.

## Problem

A Write Multiple Coils request specifies a `Quantity of Outputs`, a `Byte Count`, and the output-value bytes. The byte count must equal `ceil(Quantity of Outputs / 8)`, and the payload length must match that byte count.

Previously, `WriteMultipleCoilsRequest.decode()` read but ignored the `Byte Count`. If a request declared 16 coils but supplied only one data byte, the decoder produced eight coil values. `datastore_update()` could then write those eight values and return a successful response for eight coils instead of rejecting the inconsistent request. This silently changed the meaning of the request and could modify datastore state.

## Changes

The decoder now preserves the declared quantity, byte count, and actual payload length. Before accessing the datastore, the request verifies that the byte count matches `ceil(Quantity of Outputs / 8)` and that the payload has exactly the declared length. Invalid requests return `ILLEGAL_DATA_VALUE (0x03)`, while valid client-constructed and decoded requests retain their existing behavior.

Tests cover byte counts that are too small or too large, payloads that are shorter or longer than declared, and successful processing of a valid request. They also verify that malformed requests do not call `async_setValues()`.

### DIFF
--- a/pymodbus/pdu/bit_message.py
+++ b/pymodbus/pdu/bit_message.py
@@ -136,6 +136,7 @@ class WriteMultipleCoilsRequest(ModbusPDU):
 
     function_code = 15
     rtu_byte_count_pos = 6
+    count: int
     byte_count: int | None = None
     data_byte_count = 0
 
--- a/pymodbus/pdu/bit_message.py
+++ b/pymodbus/pdu/bit_message.py
@@ -136,6 +136,8 @@ class WriteMultipleCoilsRequest(ModbusPDU):
 
     function_code = 15
     rtu_byte_count_pos = 6
+    byte_count: int | None = None
+    data_byte_count = 0
 
     def encode(self) -> bytes:
         """Encode write coils request."""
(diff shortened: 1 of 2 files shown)


janiversen (changes requested):
Please fix variable name problem (ubuntu 3.14).

pymodbus/pdu/bit_message.py
```diff
@@ -136,6 +136,9 @@ class WriteMultipleCoilsRequest(ModbusPDU):
 
     function_code = 15
     rtu_byte_count_pos = 6
+    count: int
```
janiversen, line 139: All 3 have values that depends on the actual request.

Having those as class variables creates a potential problem, when used in parallel. This is one of the rare cases where __init__ is needed.

pymodbus/pdu/bit_message.py
```diff
@@ -149,14 +152,23 @@ def encode(self) -> bytes:
...
-        self.bits = unpack_bitstring(data[5:])[:count]
+        self.address, self.count, self.byte_count = struct.unpack(">HHB", data[0:5])
+        self.data_byte_count = len(data) - 5
+        self.bits = unpack_bitstring(data[5 : 5 + self.byte_count])[: self.count]
 
     async def datastore_update(
         self, context: ModbusServerContext, device_id: int
     ) -> ModbusPDU:
         """Run a request against a datastore."""
         count = len(self.bits)
```
janiversen: It is very confusing to have a class variable self.count and a local variable count.

janiversen (changes requested):

pymodbus/pdu/bit_message.py
MrAlaskan: Done. The three request-dependent values are now initialized per instance in __init__.

pymodbus/pdu/bit_message.py
MrAlaskan: Done. The local count variable has been removed, and self.count is now used consistently.

PR #2975, opened 2026-07-23
Fix routing of unsupported MEI types

MrAlaskan:
## Summary

Prevent unsupported Encapsulated Interface Transport (FC 0x2B) MEI types from being routed to the Read Device Identification implementation.

## Problem

Function code `0x2B` uses the `MEI Type` field to dispatch requests to different encapsulated interfaces. Read Device Identification is specifically assigned MEI type `0x0E`, while `0x0D` represents the separately defined CANopen General Reference interface and other ranges are reserved.

Previously, `ReadDeviceInformationRequest` was registered as both the base handler for function `0x2B` and the handler for MEI type `0x0E`. When the decoder received another MEI type and could not find a matching subtype, it retained the base `ReadDeviceInformationRequest` object. The server could consequently process an unsupported or reserved MEI request as Read Device Identification and return a normal `0x0E` response with semantics unrelated to the original request.

## Changes

This change introduces a generic Encapsulated Interface Transport dispatch PDU as the base handler for function `0x2B`. MEI type `0x0E` continues to route to `ReadDeviceInformationRequest`, while unsupported and reserved MEI types remain on the generic handler and return `ILLEGAL_FUNCTION (0x01)` instead of entering the device-identification logic.

Tests verify that `0x0E` still routes correctly and that MEI types `0x00`, `0x0D`, `0x0F`, and `0xFF` are not interpreted as Read Device Identification requests and return the expected exception.

### DIFF
--- a/pymodbus/pdu/mei_message.py
+++ b/pymodbus/pdu/mei_message.py
@@ -211,5 +211,6 @@ class ReadDeviceInformationResponse(ModbusPDU):
                     data[count - object_length : count],
                 ]
 
+
 DecodePDU.add_pdu(_EncapsulatedInterfaceTransport, _EncapsulatedInterfaceTransport)
 DecodePDU.add_sub_pdu(ReadDeviceInformationRequest, ReadDeviceInformationResponse)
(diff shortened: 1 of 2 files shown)


janiversen (approved):
LGTM, thanks.

PR #2976, opened 2026-07-23
Reject unknown specific device identification objects

MrAlaskan:
## Summary

Return `ILLEGAL_DATA_ADDRESS (0x02)` when an individual Read Device Identification request references an unknown object.

## Problem

Read Device Identification code `0x04` performs individual access to a single object. The Modbus specification requires the server to return exception code `0x02` when the requested object ID does not match a known object.

Previously, `ModbusDeviceIdentification.__getitem__()` used `setdefault()` when reading an object. Accessing an unknown ID therefore inserted it into the identity data with an empty value. `DeviceInformationFactory` then returned that value as a normal object, and `ReadDeviceInformationRequest.datastore_update()` constructed a successful response containing a zero-length object instead of returning an exception.

This also meant that simply reading an unknown object modified the device identity state.

## Changes

Unknown identity lookups are now non-mutating. For individual access, reserved object IDs and objects that are not configured or have no available value return `ILLEGAL_DATA_ADDRESS (0x02)`. Existing stream-access behavior and valid individual object requests remain unchanged.

Tests verify that a valid specific object is returned normally, unknown object `0x54` and unavailable private object `0xFE` return exception code `0x02`, and reading an unknown object does not insert it into the identity dictionary.

### DIFF
--- a/pymodbus/pdu/device.py
+++ b/pymodbus/pdu/device.py
@@ -205,7 +205,7 @@ class ModbusDeviceIdentification:
 
         :param key: The register to read
         """
-        return self.stat_data.setdefault(key, "")
+        return self.stat_data.get(key, "")
 
     def __str__(self):
         """Build a representation of the device.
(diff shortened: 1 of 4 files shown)


janiversen (changes requested):
Please resolve conflicts with merged PRs.

janiversen (approved):
LGTM, thanks

PR #2977, opened 2026-07-23
Fix: validate register write byte counts

MrAlaskan:
## Problem

Modbus function codes `0x10` (Write Multiple Registers) and `0x17` (Read/Write Multiple Registers) require the write byte count to equal twice the write quantity, because every register occupies two bytes.

The request decoders did not enforce this relationship. For `0x10`, decoding was based on the quantity while the declared byte count was discarded, which could cause extra data to be ignored or truncated data to raise a parsing error. For `0x17`, registers were decoded using the byte count, but a byte count inconsistent with the write quantity could still reach the datastore. As a result, malformed requests could be accepted, incorrectly parsed, or fail without returning the required `Illegal Data Value (03)` response.

## Changes

- Preserve the declared byte count and actual payload length for decoded `0x10` and `0x17` requests, validate both before accessing the datastore, and return `Illegal Data Value (03)` when they are inconsistent. Parsing now consumes only complete register values so malformed payloads can be rejected cleanly instead of raising `struct.error`.
- Add regression tests covering byte counts that are too small, too large, odd, shorter than the actual payload, or followed by extra data. The tests also verify that malformed requests do not write to the datastore and that valid requests continue to succeed.

### DIFF
--- a/pymodbus/pdu/register_message.py
+++ b/pymodbus/pdu/register_message.py
@@ -149,10 +149,9 @@ class ReadWriteMultipleRegistersRequest(ModbusPDU):
             self.write_byte_count,
         ) = struct.unpack(">HHHHB", data[:9])
         self._payload_byte_count = len(data) - 9
-        payload = data[9 : self.write_byte_count + 9]
         self.write_registers = [
-            struct.unpack(">H", payload[i : i + 2])[0]
-            for i in range(0, len(payload) - 1, 2)
+            struct.unpack(">H", data[i : i + 2])[0]
+            for i in range(9, min(len(data), self.write_byte_count + 9) - 1, 2)
         ]
 
     async def datastore_update(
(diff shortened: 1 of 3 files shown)


janiversen (changes requested):
Please solve formatting problem (ubuntu 3.14).

Please use ./check_ci.sh, that checks and reformats just like CI, and afterward commit changed files.

pymodbus/pdu/register_message.py
```diff
@@ -147,10 +148,12 @@ def decode(self, data: bytes) -> None:
             self.write_count,
             self.write_byte_count,
         ) = struct.unpack(">HHHHB", data[:9])
-        self.write_registers = []
-        for i in range(9, self.write_byte_count + 9, 2):
-            register = struct.unpack(">H", data[i : i + 2])[0]
-            self.write_registers.append(register)
+        self._payload_byte_count = len(data) - 9
+        payload = data[9 : self.write_byte_count + 9]
```
janiversen: Why make an extra split ? Line 154 could just use data directly.

janiversen (changes requested):

pymodbus/pdu/register_message.py
MrAlaskan: Done. The registers are now unpacked directly from data. I applied the same simplification to both 0x17 and 0x10.

janiversen (approved):
LGTM, thanks.

PR #2978, opened 2026-07-24
add: modbus example server and client, requirements and Chinese README

sjysjy8868:


client_poll.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
```
janiversen, line 1: This example belongs in examples/contrib, we do not allow files at top level.

README.md
```diff
@@ -0,0 +1,47 @@
+# pymodbus 示例：本地 Modbus TCP 从站（模拟器）与轮询客户端
```
janiversen, line 1: You cannot overwrite the standard README !!!

I suggest you make a directory: examples/contrib/chinese and add your files there.

Or change the names of the files so they start with chinese_

server.py
```diff
@@ -0,0 +1,64 @@
...
+        # simple update: increment first register, set second as timestamp low
+        new0 = (rr[0] + 1) & 0xFFFF
+        ts_low = int(time.time()) & 0xFFFF
+        context[slave_id].setValues(3, 0, [new0, ts_low] + rr[2:])
+        log.info("Updated holding registers: first=%d, ts_low=%d", new0, ts_low)
+        time.sleep(interval)
+
+def main():
+    # Create initial data blocks (100 items each)
+    store = ModbusSlaveContext(
```
janiversen, line 35: This is no longer being used, please use SimData instead.

server.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
```
janiversen, line 1: Please explain what is the difference between the standard example and this one ??

requirements.txt
```diff
@@ -0,0 +1 @@
+pymodbus>=2.5
```
janiversen, line 1: You cannot overwrite the standard requirements.txt !!!

Pymodbus do NOT support python 2.5 !

janiversen (changes requested):
Please solve the CI problems (ruff formatting)

Easiest way is to use ./check_ci.sh, that makes the changes automatically.

janiversen:
Any progress on the requested changes ???

janiversen:
Seems forgotten, if not so please update with the comments submitted.

PR #2984, opened 2026-07-25
Add --all-groups flag to uv sync in ci.yml

km-64:
Addresses #2982 with CI change. This can be merged before #2982, since it will also sync all groups if there any (currently there aren't any until #2982 is merged).


janiversen (approved):
LGTM, thanks

PR #2989, opened 2026-07-26
Update contents, and fix rendering of `SimDevice.action` 

km-64:
- Sphinx wasn't rendering the code block because the comment line was missing `:`. This was easily corrected replacing `#  ` with `#: `.

- Also updated the return type to reflect what the type hints state. It should be a member of the `ExcCodes` enum instead of `ExceptionResponse`.


janiversen (approved):
LGTM, thanks

PR #2992, opened 2026-07-28
Fix undecodable function-code exception identity

### LINKED ISSUE #2990: Server returns hardcoded function code (0x28) and transaction_id=0 for any undecodable function code
### Versions

- Python: 3.12
- OS: Linux
- Pymodbus: 3.14.0 (also present in 3.11.2)
- Modbus Hardware (if used): none — software server

### Pymodbus Specific

- Server: tcp - async
- Client: raw socket (issue is server-side; a raw frame is used only to trigger the decoder path)

### Description

When the server receives a request whose function code the decoder can't parse (any
unassigned/reserved function code), `ServerRequestHandler.callback_data()` catches
`ModbusIOException` and builds the exception response with two problems:

1. **The function code is a hardcoded literal `40` (`0x28`)** — not derived from the
   request. Every undecodable FC therefore produces the same exception FC byte
   (`0xA8` = `0x28 | 0x80`) regardless of what was actually sent.
2. **`response.transaction_id` is never set**, so it stays `0`. The sibling exception
   paths in the same file (`handle_request()`) explicitly do
   `response.transaction_id = self.last_pdu.transaction_id`; this branch omits it.

Net effect: every undecodable-FC request gets a byte-for-byte identical response
(`00 00 00 00 00 03 01 a8 01`), with a transaction_id of 0 that doesn't match the
request and an exception FC of 0x28 that was never sent. A TCP client correlating
responses by transaction_id (the standard practice with multiple in-flight requests)
cannot match this response to its request — in the repro below, pymodbus's *own* client
logs `Unable to decode frame Modbus Error: Unknown response`.

The offending branch on the current `dev` branch (`pymodbus/server/requesthandler.py`):

```python
def callback_data(self, data: bytes, addr: tuple | None = None) -> int:
    """Handle received data."""
    try:
        used_len = super().callback_data(data, addr)
    except ModbusIOException:
        response = ExceptionResponse(40, exception_code=ExcCodes.ILLEGAL_FUNCTION)
        self.server_send(response, 0)      # transaction_id never set -> stays 0
        return len(data)
```

Suggested fix: set `response.transaction_id = self.last_pdu.transaction_id` before
`server_send` (as the other exception branches already do), and derive the exception's
function code from the request rather than the hardcoded `40` (or leave it clearly
undefined instead of an unrelated fixed value).

### Code and Logs

Runnable, self-contained repro (pymodbus only, no downstream projects). Discovered while
fuzz-testing OpenPLC Runtime v4 (which pins pymodbus 3.11.2), but this reproduces against
a plain `ModbusTcpServer` on the latest 3.14.0, so it is a pymodbus issue, not a
downstream one.

```python
import asyncio, struct, socket, threading, time
import pymodbus
from pymodbus.datastore import (ModbusSequentialDataBlock, ModbusDeviceContext,
                                ModbusServerContext)
from pymodbus.server import StartAsyncTcpServer

PORT = 15021

def make_context():
    dev = ModbusDeviceContext(
        di=ModbusSequentialDataBlock(1, [0]*100), co=ModbusSequentialDataBlock(1, [0]*100),
        hr=ModbusSequentialDataBlock(1, [17]*100), ir=ModbusSequentialDataBlock(1, [0]*100))
    return ModbusServerContext(devices=dev, single=True)

def server_thread():
    asyncio.run(StartAsyncTcpServer(context=make_context(), address=("127.0.0.1", PORT)))

def mbap(tid, pdu, unit=1):
    return struct.pack(">HHHB", tid, 0, len(pdu)+1, unit) + pdu

def send(tid, fc):
    s = socket.socket(); s.settimeout(3); s.connect(("127.0.0.1", PORT))
    s.sendall(mbap(tid, bytes([fc, 0, 0, 0, 1])))
    r = s.recv(256); s.close(); return r

threading.Thread(target=server_thread, daemon=True).start(); time.sleep(2.5)
print("valid FC03 :", send(0x1234, 0x03).hex(), "(TID echoed)")
for tid, fc in [(0x0009, 0x09), (0x000A, 0x0A), (0x000D, 0x0D)]:
    print(f"FC 0x{fc:02X} TID 0x{tid:04X} ->", send(tid, fc).hex())
```

Output (pymodbus 3.14.0):

```
valid FC03 : 1234000000050103020011 (TID echoed)
FC 0x09 TID 0x0009 -> 00000000000301a801
FC 0x0A TID 0x000A -> 00000000000301a801
FC 0x0D TID 0x000D -> 00000000000301a801
```

All three undecodable-FC requests — sent with distinct transaction ids 0x0009/0x000A/0x000D
— return the identical `00000000000301a801`: transaction_id `0x0000` (matches none of them)
and exception FC `0xA8` (for FC 0x28, never sent). The valid FC03 read on the same server
echoes its transaction id (`0x1234`) correctly, confirming this is specific to the
undecodable-FC exception path.

---

*Reporter: Shad Malloy, shadmalloy3@gmail.com*

sankalpsthakur:
### Summary
When the server cannot decode a request function code (unassigned/reserved FC or serial noise), `ServerRequestHandler.callback_data` was building the exception response with two bugs:

1. **Hardcoded function code `40` (`0x28`)** → every undecodable request produced exception FC `0xA8`, unrelated to the request.
2. **`transaction_id` left at `0`** → TCP clients correlating in-flight requests by transaction id could not match the response.

`last_pdu` is cleared at the start of framing, so the framer now attaches the already-decoded MBAP/RTU `transaction_id` and `dev_id` to the `ModbusIOException`. The server uses those attributes on the exception path, and builds `ExceptionResponse(0x00, ILLEGAL_FUNCTION, ...)` instead of the unrelated `40`.

This matches maintainer guidance on #2990: echo transaction id, use `0x00` rather than inventing a function code from garbage.

### Validation
```text
pytest test/server/test_requesthandler.py  # 8 passed
pytest test/framer/                        # 279 passed
pytest test/transaction/test_transaction.py # 44 passed
```

New unit coverage:
- framing attrs echoed on undecodable path
- missing attrs fall back safely to zeros
- real socket frame with unknown FC `0x09` echoes MBAP tid `0x1234`

Fixes #2990

### AI/LLM disclosure
- AI coding tools (including Grok and/or Codex agent-assisted editing) were used to help draft or modify code and this PR description.
- I reviewed the complete change, understand the reasoning, and ran the reported local tests before submitting.
- This submission is original work of authorship under the project CLA / contributor terms; AI output was not pasted unreviewed.

### DIFF
--- a/pymodbus/exceptions.py
+++ b/pymodbus/exceptions.py
@@ -36,12 +36,24 @@ class ModbusException(Exception):
 class ModbusIOException(ModbusException):
     """Error resulting from data i/o."""
 
-    def __init__(self, string="", function_code=None):
+    def __init__(
+        self,
+        string="",
+        function_code=None,
+        *,
+        transaction_id: int = 0,
+        dev_id: int = 0,
+    ):
         """Initialize the exception.
 
         :param string: The message to append to the error
+        :param function_code: Optional function code associated with the error
+        :param transaction_id: Optional framing transaction id (e.g. MBAP TID)
+        :param dev_id: Optional framing device / unit id
         """
         self.fcode = function_code
+        self.transaction_id = transaction_id
+        self.dev_id = dev_id
         self.message = f"[Input/Output] {string}"
         ModbusException.__init__(self, self.message)
 
--- a/pymodbus/framer/base.py
+++ b/pymodbus/framer/base.py
@@ -88,7 +88,14 @@ class FramerBase:
                 )
                 continue
             if (pdu := self.decoder.decode(frame_data)) is None:
-                raise ModbusIOException("Unable to decode request")
+                # Preserve framing identity so the server can echo transaction/dev
+                # ids on the undecodable-function exception path (see #2990).
+                # Do not recover a function code from garbage payloads — noise on
+                # serial lines is the common cause of this path.
+                exc = ModbusIOException("Unable to decode request")
+                exc.transaction_id = tid
+                exc.dev_id = dev_id
+                raise exc
             pdu.dev_id = dev_id
             pdu.transaction_id = tid
             return used_len, pdu
--- a/pymodbus/framer/base.py
+++ b/pymodbus/framer/base.py
@@ -92,10 +92,11 @@ class FramerBase:
                 # ids on the undecodable-function exception path (see #2990).
                 # Do not recover a function code from garbage payloads — noise on
                 # serial lines is the common cause of this path.
-                exc = ModbusIOException("Unable to decode request")
-                exc.transaction_id = tid
-                exc.dev_id = dev_id
-                raise exc
+                raise ModbusIOException(
+                    "Unable to decode request",
+                    transaction_id=tid,
+                    dev_id=dev_id,
+                )
             pdu.dev_id = dev_id
             pdu.transaction_id = tid
             return used_len, pdu
--- a/pymodbus/server/requesthandler.py
+++ b/pymodbus/server/requesthandler.py
@@ -70,8 +70,8 @@ class ServerRequestHandler(TransactionManager):
             response = ExceptionResponse(
                 0x00,
                 exception_code=ExcCodes.ILLEGAL_FUNCTION,
-                device_id=getattr(exc, "dev_id", 0) or 0,
-                transaction=getattr(exc, "transaction_id", 0) or 0,
+                device_id=exc.dev_id,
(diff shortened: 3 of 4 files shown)


janiversen (approved):
VERY nice way to solve the problem!

Feel free to search for ModbusIOException, I am pretty sure there are a number of places where transaction id and/or function code is known.

LGTM, thanks

PR #2995, opened 2026-07-28
Update NullModem list with supported clients/servers

km-64:
Addresses: https://github.com/pymodbus-dev/pymodbus/discussions/2994#discussioncomment-17816099

Also removes `AsyncModbus<x>Servers` from the list as these don't exist.


janiversen (approved):
LGTM, thanks

PR #3010, opened 2026-08-27
Let CancelledError propagate from async execute

### LINKED ISSUE #3009: `Task.cancel()` can't cancel async clients, `TransactionManager.execute` swallows `CancelledError`
### Versions

- Python: 3.11
- OS: Win11
- Pymodbus: 3.12 / 3.15

### Pymodbus Specific

- Client: rtu - async

### Summary

`TransactionManager.execute` (async path) catches `asyncio.CancelledError` and re-raises it as a `ModbusIOException`. Since `ModbusIOException` is an `Exception` subclass (not `BaseException`), it gets caught by the caller's `except Exception` handler, making `Task.cancel()` ineffective for interrupting async client requests that are waiting for a response.

This affects any caller that wraps `client.read_*`/`client.write_*` in a `try/except Exception` (a common pattern for "skip on error" loops) and then tries to cancel the task from outside.

The same issue confirmed in **3.12.1** and **3.15.0**. The behavior is unchanged between these versions (3.15 just refactored `raise ModbusIOException(...)` → `raise self._io_exception_from_request(...)`).

### Location

`pymodbus/transaction/transaction.py`, in `TransactionManager.execute`:

```python
                try:
                    response = await asyncio.wait_for(
                        self.response_future, timeout=self.comm_params.timeout_connect
                    )
                    ...
                    return response
                    count_retries += 1
                except asyncio.exceptions.CancelledError as exc:   # <-- here
                    raise ModbusIOException(
                        "Request cancelled outside library."
                    ) from exc                       # 3.15: self._io_exception_from_request(...)
```

### The problem

`CancelledError` inherits from `BaseException` (since Python 3.8), **not** `Exception`. This lets `Task.cancel()` punch through `except Exception` blocks so cancellation propagates to the top of the task. By catching `CancelledError` and re-raising as `ModbusIOException` (an `Exception` subclass), pymodbus **downgrades the cancellation signal** into something ordinary error handlers will swallow.

Our Modbus testing tool runs routines that loop over multiple slave addresses, each wrapped in `try/except Exception` to skip unresponsive slaves (a standard "skip-on-error" pattern). When the user clicks "Stop", we call `Task.cancel()` on the running routine. Because of this issue, the cancellation is silently swallowed at every slave address and the scan continues until it naturally exhausts the address list or times out. The stop button appears to do nothing.

### Reproducer

```python
import asyncio
from pymodbus.client import AsyncModbusSerialClient  # or AsyncModbusTcpClient


async def routine():
    client = AsyncModbusSerialClient(port="COM3", baudrate=9600, timeout=10, retries=0)
    await client.connect()
    try:
        # This awaits a response that will never come (no slave connected)
        await client.read_holding_registers(address=0, count=1, device_id=1)
    except Exception:
        print("routine: caught Exception (cancel was swallowed!), continuing...")
        return "routine continued"
    finally:
        client.close()


async def main():
    task = asyncio.ensure_future(routine())
    await asyncio.sleep(0.5)   # let it start waiting for response
    task.cancel()              # try to cancel
    try:
        result = await task
        print(f"main: task returned normally: {result!r}")
        print("==> CANCEL DID NOT WORK")
    except asyncio.CancelledError:
        print("main: task cancelled ✓")

asyncio.run(main())
```

**Output (pymodbus 3.12 / 3.15):**
```
routine: caught Exception (cancel was swallowed!), continuing...
main: task returned normally: 'routine continued'
==> CANCEL DID NOT WORK
```

`Task.cancel()` should have terminated the task at the `await asyncio.wait_for(...)` suspension point. Instead, pymodbus caught the `CancelledError` and re-raised it as `ModbusIOException`, which `routine`'s `except Exception` swallowed, and the routine kept running.

### Suggested fix

Remove the `except CancelledError` clause from `execute`, letting `CancelledError` propagate unchanged:

```python
                    count_retries += 1
                # removed: except CancelledError → raise ModbusIOException
```

`CancelledError` will then propagate naturally through `wait_for` → `execute` → the caller, as asyncio intends. The `response_future` is an internal `asyncio.Future` that `wait_for` will cancel on its way out, so no cleanup is lost.

If the original intent was to log the cancellation, a `finally:` or a bare re-raise would preserve observability without downgrading the signal:

```python
                except asyncio.exceptions.CancelledError:
                    Log.debug("Request cancelled outside library.")
                    raise
```

### Our workaround

We monkeypatch `TransactionManager.execute` at startup with a version that omits the `except CancelledError` clause. This makes `Task.cancel()` work correctly (2ms stop latency in real-world testing on a USB-to-485 adapter).

ChrisJr404:
Fixes #3009.

The async execute path caught asyncio.CancelledError and re-raised it as a ModbusIOException. Because that is a plain Exception, callers using a try/except Exception around read/write calls swallow it, so Task.cancel() no longer actually cancels a request that is waiting for a response.

Dropping the handler lets the CancelledError propagate normally, which releases the lock on the way out and cancels the task as expected. Updated the cancel scenario in the transaction test to check the task is cancelled rather than expecting the wrapped exception.

### DIFF
--- a/pymodbus/transaction/transaction.py
+++ b/pymodbus/transaction/transaction.py
@@ -224,11 +224,6 @@ class TransactionManager(ModbusProtocol):
                     return response
                 except asyncio.exceptions.TimeoutError:
                     count_retries += 1
-                except asyncio.exceptions.CancelledError as exc:
-                    raise self._io_exception_from_request(
-                        "Request cancelled outside library.",
-                        request,
-                    ) from exc
             if self.count_until_disconnect < 0:
                 self.connection_lost(asyncio.TimeoutError("Server not responding"))
                 raise self._io_exception_from_request(
--- a/test/transaction/test_transaction.py
+++ b/test/transaction/test_transaction.py
@@ -247,10 +247,9 @@ class TestTransaction:
             await asyncio.sleep(0.1)
             resp.cancel()
             await asyncio.sleep(0.1)
-            with pytest.raises(ModbusIOException) as exc_info:
+            with pytest.raises(asyncio.CancelledError):
                 await resp
-            assert exc_info.value.fcode == request.function_code
-            assert exc_info.value.dev_id == request.dev_id
+            assert resp.cancelled()
         elif scenario == 7:  # response
             transact.comm_params.timeout_connect = 0.2
             resp = asyncio.create_task(transact.execute(False, request))


janiversen (approved):
LGTM, thanks

PR #3021, opened 2026-09-09
fix(async): synchronize connect() promise with connection_made callback

### LINKED ISSUE #3019: AsyncModbusTcpClient.connect() returns True while .connected property is False (3.13.1)
Minimal reproduction:
    import asyncio
    from pymodbus.client import AsyncModbusTcpClient

    async def test():
        client = AsyncModbusTcpClient('<device-ip>', port=502)
        connected = await client.connect()
        print(f'connect() return value: {connected}')      # True
        print(f'client.connected property: {client.connected}')  # False
        # client.transport raises AttributeError: no attribute 'transport'

    asyncio.run(test())

Environment: pymodbus 3.13.1, Python 3.14, Home Assistant OS 18.2 (both HA Core 2026.9.1 and 2026.8.3 affected identically).

Any subsequent request (e.g. read_input_registers) immediately raises:
    ConnectionException: Modbus Error: [Connection] Not connected[AsyncModbusTcpClient <ip>:502]

Even a 1-second delay between connect() and the first request does not help - this rules out a simple timing/race condition. The device itself is confirmed reachable and responsive (verified independently via a separate, unrelated Modbus TCP client against the same device at the same time).

Yuvrajup:
Closes #3019

### DIFF
--- a/pymodbus.log
+++ b/pymodbus.log
@@ -1,0 +1,81 @@
+2026-09-10 01:22:16,121 DEBUG test_logging:69 test 1
+2026-09-10 01:22:16,122 DEBUG test_logging:70 Repeating....
+2026-09-10 01:22:16,122 ERROR test_logging:72 get frames
+2026-09-10 01:22:16,123 CRITICAL test_logging:73 Repeating....
+2026-09-10 01:22:16,123 WARNING test_logging:77 test 2
+2026-09-10 01:22:16,124 WARNING test_logging:78 Repeating....
+2026-09-10 01:22:16,125 CRITICAL test_logging:81 test 3no
+>>>>> send: 0x61 0x62 0x63
+>>>>> Repeating....
+>>>>> None
+2026-09-10 01:22:16,125 CRITICAL test_logging:83 test 3
+2026-09-10 01:22:16,125 CRITICAL test_logging:84 Repeating....
+2026-09-10 01:22:16,126 ERROR test_logging:89 test 4
+2026-09-10 01:22:16,126 ERROR test_logging:90 Repeating....
+2026-09-10 01:22:16,127 INFO  test_logging:95 test 5
+2026-09-10 01:22:16,127 INFO  test_logging:96 Repeating....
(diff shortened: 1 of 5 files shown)


janiversen (changes requested):
This is overly complicated and not needed, current solution uses the standard asyncio methods.

client,connect returns the value comming from asyncio, which according to the documentation is true/false depending on whether a connection was established.

client.connected tests if the connection is stil open by referring to the transport object. Remark connection_made/ connection_lost changes the transport object.

The issue you refer to, is not actually an issue in pymodbus (see the discussion) but a device problem.

janiversen (changes requested):
If you want this merged, you need to make a test case that shows that the current solution is wrong and yours work.

Yuvrajup:
Challenge accepted, @janiversen! I completely understand wanting to keep the library lean and relying on standard `asyncio` methods. 

You are right that `create_connection` returns the underlying connection. However, the issue is that there is a microscopic event-loop gap between `create_connection` resolving and the protocol's `connection_made()` callback actually firing to set the `.transport` object inside the client. 

I just pushed a new test case (`test_async_connect_race_condition`) that artificially delays the `connection_made` callback by a few milliseconds to deterministically simulate a slow event loop or hardware latency. 
* Without this PR's synchronization, `client.connect()` returns `True`, but `client.connected` evaluates to `False` (crashing subsequent requests).
* With this PR, `client.connect()` safely awaits the callback, ensuring the transport is fully hydrated before returning control to the caller.

Let me know what you think of the test case!

janiversen:
Forget about that race, pymodbus do not have the transport object until connection_made is called.

Apart from that in my experience, when debugging, connection_made was always called before the connect returned.

Just checked your test case, the delay is inside connection_made() and thus in pymodbus code, so your test case does not show if connection_made is called before connect returns.

janiversen:
still working on this?

janiversen:
Closing as it seems abandoned, please reopen if this is wrong.

Yuvrajup:
Closing in favor of #3030 which correctly addresses the root cause.

